PR #14, opened 2018-01-18
NWNX_Events: Add OnUseFeat event

zunath:
Adds an event hook for feat usage.


Plugins/Events/Events.cpp
```diff
@@ -89,6 +90,11 @@ Events::Events(const Plugin::CreateParams& params)
         m_itemEvents = std::make_unique<ItemEvents>(GetServices()->m_hooks);
     }
 
+    if (GetServices()->m_config->Get<bool>("ENABLE_ITEM_EVENTS", true))
```
Liareth: I guess this should be ENABLE_FEAT_EVENTS?

zunath: Good catch. Fix'd

PR #19, opened 2018-01-19
Fix for PgSQL reconnect issue

ELadner:
with the existing code, the PGsql target is unusable.

I swear I tested this before.  I don't know why, but it freaked out with the parent level SQL changes.  Commented in the code, but basically, it doesn't like anything executed between the prepare and execute.  I guess it uses the same unnamed buffer.  And using named prepares is out because it forces the NWN scripters to call Destroy (not optional -- REQUIRED -- at that point or prepared queries will build up in the server forever.  it's not clear if they are ever reaped unless DEALLOCATED by the user).

Also, as a bonus.. don't willy-nilly echo the database password to the log files.

### DIFF
--- a/Plugins/SQL/Targets/PostgreSQL.cpp
+++ b/Plugins/SQL/Targets/PostgreSQL.cpp
@@ -36,13 +36,17 @@ void PostgreSQL::Connect(NWNXLib::ViewPtr<NWNXLib::Services::ConfigProxy> config
 
     const std::string port = "port=" + config->Get<std::string>("PORT", "5432");
 
-    // Build the m_connection string
-    std::string m_connectstring = host + " " + port + " " + db + " " + user + " " + pass;
+    // Build the m_connection string - this is used later in the PQping.
+    m_connectString = host + " " + port + " " + db + " " + user ;
 
-    m_log->Info("Connect String:  %s\n", m_connectstring.c_str());
+    // hide the password in the log file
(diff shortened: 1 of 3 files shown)


Liareth:
Looks good to me, thanks!

Plugins/SQL/Targets/PostgreSQL.cpp
```diff
@@ -36,13 +36,17 @@ void PostgreSQL::Connect(NWNXLib::ViewPtr<NWNXLib::Services::ConfigProxy> config
...
     const std::string port = "port=" + config->Get<std::string>("PORT", "5432");
 
-    // Build the m_connection string
-    std::string m_connectstring = host + " " + port + " " + db + " " + user + " " + pass;
+    // Build the m_connection string - this is used later in the PQping.
+    m_connectString = host + " " + port + " " + db + " " + user ;
 
-    m_log->Info("Connect String:  %s\n", m_connectstring.c_str());
+    // hide the password in the log file
+    m_log->Info("Connect String:  %s password=xxxxxxxx\n", m_connectString.c_str());
```
Liareth: You shouldn't need the \n here, since the logger writes \n each time.

ELadner: I will correct this.  I haven't tested the null parameter handling yet -- there may be some further fixes today.

PR #20, opened 2018-01-19
NWNX_Events: Add OnCastSpell event hooks.

zunath:
Adds a hook for OnCastSpell which fires whenever a spell lands on its target either from an item or from a normal creature spell casting sequence.

### DIFF
--- a/Plugins/Events/Events/SpellEvents.cpp
+++ b/Plugins/Events/Events/SpellEvents.cpp
@@ -1,0 +1,61 @@
+#include "Events/SpellEvents.hpp"
+#include "API/CNWSCreature.hpp"
+#include "API/Functions.hpp"
+#include "Events.hpp"
+#include "Helpers.hpp"
+
+namespace Events {
+
+using namespace NWNXLib;
+
+SpellEvents::SpellEvents(ViewPtr<Services::HooksProxy> hooker)
+{
+    hooker->RequestSharedHook<
+        NWNXLib::API::Functions::CNWSObject__SpellCastAndImpact, 
+        int32_t, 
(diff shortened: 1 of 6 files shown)


Plugins/Events/Events/SpellEvents.cpp
```diff
@@ -0,0 +1,61 @@
...
+    Events::PushEventData("SPELL_ID", std::to_string(spellID));
+
+    Events::PushEventData("TARGET_POSITION_X", std::to_string(targetPosition.x));
+    Events::PushEventData("TARGET_POSITION_Y", std::to_string(targetPosition.y));
+    Events::PushEventData("TARGET_POSITION_Z", std::to_string(targetPosition.z));
+
+    Events::PushEventData("TARGET_OBJECT_ID", Helpers::ObjectIDToString(oidTarget));
+    Events::PushEventData("MULTI_CLASS", std::to_string(multiClass));
+    Events::PushEventData("ITEM_OBJECT_ID", Helpers::ObjectIDToString(oidItem));
+    Events::PushEventData("SPELL_COUNTERED", std::to_string(spellCountered));
```
Liareth: By the standard, there isn't an overload of std::to_string for bool. It might be a non-standard extension (in which case, maybe double check that it prints 0/1 rather than "false" and "true"?) but more likely it is implicitly converting bool to an int which is why it works.

I just wanted to call it out here because I can totally see some platform making std::to_string(bool) which outputs the text representation of a bool instead. It might be worth explicitly casting the bool to an int, or doing something like:

`Events::PushEventData("SPELL_COUNTERED", spellCountered ? "1" : "0");`

Liareth:
Looks good to me!

PR #21, opened 2018-01-19
Normalize logging between the SQL targets

ELadner:
Tried to create a parity between both targets as to what and where logging takes place.  Added some Debug level output that could possibly assist script developers trying to debug SQL issues.  Also modified the query results output to either show number of rows returned (for selects) or the number of rows affected (inserts, updates, deletes) as appropriate.


Plugins/SQL/Targets/MySQL.cpp
```diff
@@ -39,11 +46,17 @@ bool MySQL::IsConnected()
...
+    if (!bConnected)
+    {
+        m_log->Warning("Disconnected state identified.");
+    }
     return bConnected;
 }
 
 bool MySQL::PrepareQuery(const Query& query)
 {
+    m_log->Debug("Preparing query %s\n", query.c_str());
```
Liareth: Shouldn't have a \n here!

Liareth:
Merging now ahead of the line ending change.

PR #27, opened 2018-01-21
NWNX_Redis

niv:
See README.md for what this is about.

Only sparingly tested.


Liareth:
There's a lot of code here so I'm not gonna review it all. The documentation looks good and the nwscript solution (_short) is great, so let's merge this and see what happens. Nice work!

PR #48, opened 2018-02-25
add -D_FILE_OFFSET_BITS=64 so it doesn't conk out on docker-on-mac|win

niv:
.. because, apparently, those APIs can't handle 32bit directory structure entries.

Impact should be minimal (famous last words) since we're not assuming any specific width on those types anywhere in the codebase. Right?


Liareth:
seems legit

PR #50, opened 2018-03-01
Add event data for getting the item property index used for the OnUsedItem Event

zunath:
Adds ITEM_PROPERTY_INDEX to the UseItem event. Useful for retrieving which individual item property is used on an item.


Liareth:
Look's good to me. Here's the actual function signature - so if you need the other uint8_t, this is what it is.

```
BOOL CNWSCreature::UseItem( OBJECT_ID oidItem, BYTE nActivePropertyIndex, BYTE nSubPropertyIndex, OBJECT_ID oidTarget, Vector vTargetPosition, OBJECT_ID oidArea )
```

PR #53, opened 2018-03-05
Change methods for accessing creature ability scores

zunath:
Renames SetAbilityScore to SetRawAbilityScore.
Adds GetRawAbilityScore (excludes racial bonuses/penalties)
Adds ModifyAbilityScore (excludes racial bonuses/penalties)
Deprecates NWNX_Creature_SetAbilityScore in the NWScript file.

This change is being made because of issues with racial bonuses. The NWScript GetAbilityScore() method returns racial bonuses which were throwing off calls like this:

int str = GetAbilityScore(oPC, ABILITY_STRENGTH) + 1;
NWNX_CREATURE_SetAbilityScore(oPC, ABILITY_STRENGTH, str);

Based on discussion with @mtijanic I made the changes above.

### DIFF
--- a/Plugins/Creature/Creature.cpp
+++ b/Plugins/Creature/Creature.cpp
@@ -70,7 +70,9 @@ Creature::Creature(const Plugin::CreateParams& params)
     REGISTER(GetClassByLevel);
     REGISTER(SetBaseAC);
     REGISTER(GetBaseAC);
-    REGISTER(SetAbilityScore);
+    REGISTER(SetRawAbilityScore);
+    REGISTER(GetRawAbilityScore);
+    REGISTER(ModifyRawAbilityScore);
     REGISTER(GetMemorisedSpell);
     REGISTER(GetMemorisedSpellCountByLevel);
     REGISTER(SetMemorisedSpell);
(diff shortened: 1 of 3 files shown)


zunath:
@Liareth I think this one is good. @mtijanic has signed off on it. 

I had to fix a merge conflict from the PR you accepted yesterday. Let me know if I missed something

Plugins/Creature/Creature.cpp
```diff
@@ -451,7 +453,83 @@ ArgumentStack Creature::SetAbilityScore(ArgumentStack&& args)
...
+    return stack;
+}
+
+ArgumentStack Creature::ModifyRawAbilityScore(ArgumentStack&& args)
+{
+    ArgumentStack stack;
+    if (auto *pCreature = creature(args))
+    {
+        const auto ability = Services::Events::ExtractArgument<int32_t>(args);
+        const auto offset  = Services::Events::ExtractArgument<int32_t>(args); assert(value <= 255);
```
Liareth: I guess technically speaking you want to assert that

offset >= -base (underflow)
base + offset <= 255 (overflow) or, practically speaking, base + offset <= 127

When they do break it's going to be obvious enough to whomever uses it why they broke. So not a big deal really. Just something you could improve if you felt like it, or not really.

PR #54, opened 2018-03-08
Updated for PostgeSQL build

kmsheehan:
Added libraries and settings so the builder will add PostgreSQL to the SQL library

### DIFF
--- a/Dockerfile.builder
+++ b/Dockerfile.builder
@@ -9,7 +9,7 @@ RUN apt-get install -y --no-install-recommends  \
   gcc-6 g++-6
 
 # NWNX_SQL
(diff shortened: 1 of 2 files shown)


Liareth:
I noticed we've hardcoded paths for PgSQL and MySQL now. I'm not very familiar with how these things are usually done on Linux - is this normal? Or is this a hack just to make it work?

kmsheehan:
The CMAKE find for PostgreSQL says those may have to be defined for it to work - we have a similar 'hack' for the JVM currently.  I can review the CMAKE structure over the weekend and suggest some changes that are more universal.

PR #60, opened 2018-03-13
In rundockerbuild.sh corrected the docker pull URL

nwngracken:
It was docker.nwnx.io:443/nwnxee/unified:builder, which
returned SSL x509 errors.  The new, correct URL is evidently
nwnxee/builder.


Liareth:
Thanks! I don't know much about the Docker setup so adding Glorwinger as a reviewer who set it all up.

PR #61, opened 2018-03-13
New Item Plugin

BhaalM:
Only two functions by now: SetWeight and SeGoldPieceValue. Both functions have been tested on my module.

There is already an Object plugin the functionality could be added to this plugin instead of a new one... But if in the future we add custom ItemProperties an do so, maybe it's better to have it separated. Anyway, as you wish.

Regards

### DIFF
--- a/Plugins/Item/CMakeLists.txt
+++ b/Plugins/Item/CMakeLists.txt
@@ -1,0 +1,2 @@
+add_plugin(Item
+    "Item.cpp")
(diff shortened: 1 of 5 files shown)


zunath:
This was on my list of things to do. Thanks for saving me time!

BhaalM:
Glad to help! :)

Liareth:
Looks good to me. Great work! For item properties, we can probably make a nwnx_itemproperty plugin. Similar deal for nwnx_effects.

PR #62, opened 2018-03-14
New version of the Item plugin

BhaalM:
Added the following functions (with NWNX_Item_...): 

- SetBaseItemType
- SetItemColor
- SetItemAppearance
- GetEntireItemAppearance
- RestoreItemAppearance

Tested in my module. Apparently all works as intended.

### DIFF
--- a/Plugins/Item/NWScript/nwnx_item.nss
+++ b/Plugins/Item/NWScript/nwnx_item.nss
@@ -15,7 +15,25 @@ void NWNX_Item_SetBaseItemType(object oItem, int nBaseItem);
 
 // Set a color value on an item. This will not be visible to PCs until
 // the item is refreshed for them (e.g. by logging out and back in).
-void NWNX_Item_SetItemColor (object oItem, int nIndex, int nColor);
+// nIndex range: [0, 6]
+// nColor range: [0,255]
+void NWNX_Item_SetItemColor(object oItem, int nIndex, int nColor);
+
+// Set an appearance value on an item. This will not be visible to PCs until
+// the item is refreshed for them (e.g. by logging out and back in).
+// nIndex range:
+//    - Color: [-9,-4]
+//    - General item: [-3,-1]
+//    - Armor: [0,18]
+// nValue range: [0,255]
+void NWNX_Item_SetItemAppearance(object oItem, int nIndex, int nValue);
+
+/* Return a string containing the entire appearance for oItem which can later be
+ * passed to RestoreItemAppearance(). */
+string NWNX_Item_GetEntireItemAppearance (object oItem);
+
+/* Restore an item's appearance with the value returned by GetEntireItemAppearance(). */
+void NWNX_Item_RestoreItemAppearance (object oItem, string sApp);
 
 
 void NWNX_Item_SetWeight(object oItem, int w)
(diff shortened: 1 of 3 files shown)


Plugins/Item/NWScript/nwnx_item.nss
```diff
@@ -2,32 +2,110 @@
...
+// Set a color value on an item. This will not be visible to PCs until
+// the item is refreshed for them (e.g. by logging out and back in).
+// nIndex range: [0, 6]
+// nColor range: [0,255]
+void NWNX_Item_SetItemColor(object oItem, int nIndex, int nColor);
+
+// Set an appearance value on an item. This will not be visible to PCs until
+// the item is refreshed for them (e.g. by logging out and back in).
+// nIndex range:
+//    - Color: [-9,-4]
```
Liareth: The -9 thing is super weird ...

BhaalM: I know :/

This is because I've used the values of Constants.hpp

I've checked that the nwnx2 plugin uses the same constants values:
Line 559 -> https://github.com/NWNX/nwnx2-linux/blob/master/include/nwn_const.h

But maybe you took this file from the nwnx2 project so...

If this is the case, maybe... maybe these constants represented "indexes" in an array whose address were misplaced in the item structure... don't know if you know what I mean... For example, suppose that a certain time of th development of the nwnx2 the internal struct of the item was:

...
uint8_t Color[6];
uint8_t ItemApp[3];
uint8_t ArmorApp[19];
...

If you use ArmorApp as a pointer for the three arrays, whith ArmorApp[-9] you are accessing the Color.

Anyway, this is not the case now, because internally we have three arrays: one for the color, one for the item appearance and one for the armor. BUT, if you want to preserve backward compatibility with the NWNX2 funcs plugin, -9 its a color.

I can, if you wish, modify the .nss and use three arguments with the values of the constants used by the nwnscript.nss. Of course this will not be compatible with the nwnx2 plugin (I personally prefer this option)

Liareth: I feel like we should update Constants.hpp, and throw an exception if anything tries to pass in a negative parameter. Probably the safest option. What do you think?

BhaalM: Actually, the constants are not needed. What I can do is:

1) Change the plugin function to accept only a positive index
2) Change the .nss with the new functionality, I can even add different functions: one for weapons, one for armors and one for the colors, and these ones can use the nwnscript constants.
3) The last point raises another question, I copied the functionality of the nwnx2 plugin so there is already a "SetItemColor" which can already be done with the "SetItemAppearance". For the sake of simplicity I think it's better to remove it.
4) Optional: add to the .nss a function backward compatible with the nwnx2 version

Plugins/Item/NWScript/nwnx_item.nss
```diff
@@ -2,32 +2,60 @@
 
 const string NWNX_Item = "NWNX_Item";
 
-// Sets the provided item's current weight to the provided value.
+// Set oItem's weight. Will not persist through saving.
 void NWNX_Item_SetWeight(object oItem, int w);
 
-// Sets the provided item's current golp piece value to the provided 
-// value.
+// Set oItem's value in golp pieces. Will not persist through saving.
```
zunath: Typo here - "golp" should be "gold".

Plugins/Item/NWScript/nwnx_item.nss
Liareth: That sounds good to me. Anything to avoid users having to pass -9 in as an index! Let me know once that's changed and we can get this merged. :)

BhaalM: ok!

Issue #67, opened 2018-03-16
NWN_Object crash


driller-nwn:
When you attempt to use the NWNX_Object_Deserialize or NWNX_Object_Serialize function, it causes nwserver to crash with a segmentation fault. This appeared after the 8164 update.

Liareth:
I've pushed a speculative fix, I can't test it right now but I'm hoping it helps. Can you try https://github.com/nwnxee/unified/commit/a3f2e51cc71b0dd74b0569775f8c4b29ecf4df0b please?

Liareth:
This was actually caused by the new logging system - NWNX_CORE_LOG_LEVEL=7 caused a massive object to be printed, writing over the max buffer length. @mtijanic kindly fixed it. You can work around it without any binary changes by lowering the log level.

PR #68, opened 2018-03-16
Fix log crashes when printing large messages

mtijanic:
See #67 - when printing large messages, such as serialized objects passed as arguments, we overflow the relatively tiny print buffer.

This PR makes it larger, and also stops crashing when insane messages are sent.


Liareth:
Thanks for fixing this. There will also be an issue with asserts which I'll fix tomorrow.

PR #73, opened 2018-03-18
Added SetPlaceableAppearance and GetPlaceableAppearance

BhaalM:
```
// Set plac's appearance. Will not update for PCs until they
// re-enter the area.
void NWNX_Object_SetPlaceableAppearance(object plac, int app);

// Get plac's appearance
int NWNX_Object_GetPlaceableAppearance(object plac);
```


mtijanic:
I'd suggest not limiting to placeables. Creature appearance is available in the base game, but e.g. door appearance would also be useful.

No need to implement that now, make sure the API is compatible in the future - no need for "Placeable" in the function names

BhaalM:
ok!

BhaalM:
I think it is correct now

PR #74, opened 2018-03-18
PerObjectStorage service

mtijanic:
A new service that allows us to store arbitrary data on any `CGameObject` to fetch afterwards - meant to replace various adhoc maps, or using the `CNWSScriptVarTable` for NWNX.

Still a lot to do, and not very thoroughly tested, but should not break anything by its mere presence. Several inefficiencies and workarounds for now:
- In the future, we'll use a dedicated pointer in `CGameObject`
- Cannot hook `CGameObject::~CGameObject` because multiple destructors exist and API is not generating the addresses. Or... base class destructor is simply not called?
- Using 4 different maps for now. Should probably be merged using something like `variant`

Testing: Writing, reading and erasing from the object storage in the creature plugin. Destroying the creature and checking whether the cleanup is done.

### DIFF
--- a/NWNXLib/Services/PerObjectStorage/PerObjectStorage.cpp
+++ b/NWNXLib/Services/PerObjectStorage/PerObjectStorage.cpp
@@ -1,0 +1,205 @@
+#include "Services/PerObjectStorage/PerObjectStorage.hpp"
+#include "API/Constants.hpp"
+
+namespace NWNXLib {
+
+namespace Services {
+
+
+void PerObjectStorage::Set(API::Types::ObjectID object, std::string key, int value)
+{
+    // TODO: check if the object is actually valid
+    if (object == API::Constants::OBJECT_INVALID)
+        return;
+
+    m_objectStorage[object]->GetIntMap().emplace(key, value);
+}
+void PerObjectStorage::Set(API::Types::ObjectID object, std::string key, float value)
+{
+    // TODO: check if the object is actually valid
+    if (object == API::Constants::OBJECT_INVALID)
+        return;
+
+    m_objectStorage[object]->GetFloatMap().emplace(key, value);
+}
+void PerObjectStorage::Set(API::Types::ObjectID object, std::string key, std::string value)
+{
+    // TODO: check if the object is actually valid
+    if (object == API::Constants::OBJECT_INVALID)
+        return;
+
+    m_objectStorage[object]->GetStringMap().emplace(key, value);
+}
+void PerObjectStorage::Set(API::Types::ObjectID object, std::string key, void *value, CleanupFunc cleanup)
+{
+    // TODO: check if the object is actually valid
+    if (object == API::Constants::OBJECT_INVALID)
+        return;
+
+    m_objectStorage[object]->GetPointerMap().emplace(key, std::make_pair<>(value, cleanup));
+}
+
+
+
+void PerObjectStorage::Remove(API::Types::ObjectID object, std::string key)
+{
+    auto it = m_objectStorage.find(object);
+    if (it != m_objectStorage.end())
+    {
+        it->second->GetStringMap().erase(key);
+        it->second->GetIntMap().erase(key);
+        it->second->GetFloatMap().erase(key);
+        it->second->GetPointerMap().erase(key);
+    }
+}
+
+PerObjectStorage::PerObjectStorage()
+{
+
+}
+PerObjectStorage::~PerObjectStorage()
+{
+
+}
+
+PerObjectStorage::ObjectStorage::IntMap& PerObjectStorage::ObjectStorage::GetIntMap()
+{
+    if (!m_IntMap)
+        m_IntMap = std::make_unique<IntMap>();
+    return *m_IntMap;
+}
+PerObjectStorage::ObjectStorage::FloatMap& PerObjectStorage::ObjectStorage::GetFloatMap()
+{
+    if (!m_FloatMap)
+        m_FloatMap = std::make_unique<FloatMap>();
+    return *m_FloatMap;
+}
+PerObjectStorage::ObjectStorage::StringMap& PerObjectStorage::ObjectStorage::GetStringMap()
+{
+    if (!m_StringMap)
+        m_StringMap = std::make_unique<StringMap>();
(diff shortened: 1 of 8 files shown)


NWNXLib/Services/PerObjectStorage/PerObjectStorage.cpp
```diff
@@ -0,0 +1,224 @@
...
+
+std::unordered_map<API::Types::ObjectID, std::unique_ptr<PerObjectStorage::ObjectStorage>> PerObjectStorage::g_objectStorage;
+
+PerObjectStorage::ObjectStorage& PerObjectStorage::GetObjectStorage(API::Types::ObjectID object)
+{
+    auto it = g_objectStorage.find(object);
+    if (it != g_objectStorage.end())
+        return *it->second;
+
+    return *g_objectStorage.emplace(object, std::make_unique<PerObjectStorage::ObjectStorage>(object)).first->second;
```
Liareth: I just want to say that I love this line.

NWNXLib/Services/PerObjectStorage/PerObjectStorage.cpp
```diff
@@ -0,0 +1,224 @@
...
+
+    GetObjectStorage(object).GetStringMap().emplace(key, value);
+}
+void PerObjectStorage::Set(API::Types::ObjectID object, std::string key, void *value, CleanupFunc cleanup)
+{
+    // TODO: check if the object is actually valid
+    if (object == API::Constants::OBJECT_INVALID)
+        return;
+
+    GetObjectStorage(object).GetPointerMap().emplace(key, std::make_pair<>(value, cleanup));
```
Liareth: You don't need the brackets on `make_pair<>` - it's fine to do `make_pair(first, second)` as the types are deduced automatically. Doesn't matter either way - just thought you might want to know!

NWNXLib/Services/PerObjectStorage/PerObjectStorage.cpp
```diff
@@ -0,0 +1,224 @@
...
+
+
+PerObjectStorageProxy::PerObjectStorageProxy(PerObjectStorage& perObjectStorage, std::string pluginName)
+    : ServiceProxy<PerObjectStorage>(perObjectStorage)
+{
+    m_pluginName = pluginName;
+}
+PerObjectStorageProxy::~PerObjectStorageProxy()
+{
+    // TODO cleanup all storage from this plugin
```
Liareth: I guess we want each plugin to have its own storage and can't query another plugin's storage. We'll want to give the PerObjectStorageProxy the plugin ID (0 to n) and have the NWNX pointer in CGameObject contain a vector of ObjectStorage or a map of pluginid->storage, one per plugin.

Liareth: Actually, scratch that, I just saw that you're adding the plugin name to each key. That works too.

NWNXLib/Services/PerObjectStorage/PerObjectStorage.cpp
```diff
@@ -0,0 +1,224 @@
...
+}
+template <> Maybe<float> PerObjectStorage::Get<float>(API::Types::ObjectID object, std::string key)
+{
+    auto it = g_objectStorage.find(object);
+    if (it != g_objectStorage.end())
+    {
+        auto map = it->second->GetFloatMap();
+        auto it2 = map.find(key);
+        if (it2 != map.end())
+            return Maybe<float>(std::move(it2->second));
```
Liareth: If you do this, when you call Get<>, it's going to steal the value from the map and move it into the Maybe. You don't want move here. Better to add an lvalue constructor to Maybe. I'll do it soon for you, then you can update this.

NWNXLib/Services/PerObjectStorage/PerObjectStorage.cpp
```diff
@@ -0,0 +1,224 @@
...
+void PerObjectStorage::CGameObjectArray__Delete__1_hook(Services::Hooks::CallType type, API::CGameObjectArray* thisPtr, uint32_t id, API::CGameObject **ptr)
+{
+    // unreferenced variables
+    (void)(sizeof(thisPtr) & sizeof(ptr));
+
+    if (type != Services::Hooks::CallType::AFTER_ORIGINAL)
+        return;
+
+    id &= 0x7FFFFFFF;
+    LOG_DEBUG("Destroying object storage for objectId:0x%08x", id);
```
Liareth: I guess practically speaking we don't want this as it will probably be very very slow ... ?

NWNXLib/Services/PerObjectStorage/PerObjectStorage.cpp
mtijanic: _Really_? `Maybe<T>` has a `T m_data`, so I assume it would create a copy in the constructor when it does ` m_data(data)`

NWNXLib/Services/PerObjectStorage/PerObjectStorage.cpp
mtijanic: We definitely don't want the printing. I was hoping the perf of an early GTFO is negligible.

NWNXLib/Services/PerObjectStorage/PerObjectStorage.cpp
Liareth: Actually, you're right, but only because of a bug.

```cpp
template <typename T>
Maybe<T>::Maybe(T&& data)
    : m_hasValue(true), m_data(data)
{ }
```

It takes an rvalue reference. The rvalue reference decays into an lvalue reference, which is then copied with `m_data(data)`. What SHOULD happen is this: `m_data(std::forward<T>(data))`, which moves the value instead.

See https://github.com/nwnxee/unified/pull/77

mtijanic: Updated with the new Maybe constructor..

mtijanic:
Note that I fully expect there to be issues here once we start using it. But it should not bring down the existing code, and those issues should hopefully be caught in plugin development, before it is shipped.

PR #76, opened 2018-03-18
events: re-enable DM action events

mtijanic:
For now, mostly only notify before and after, no arguments are passed. The events are triggered on the DM object.

Give gold/xp/level get their arguments since they are most requested. For others, check before/after etc for now.

~~The old behavior for XP/Gold can be emulated by checking the value on before and after.~~

When the API for skipping events lands, I'll change this to an exclusive hook and then I can properly extract the arguments from the message - ~~CNWSMessage does not have Peek() options, only Pop(). ~~

### DIFF
--- a/Plugins/Events/Events/DMActionEvents.cpp
+++ b/Plugins/Events/Events/DMActionEvents.cpp
@@ -3,6 +3,7 @@
 #include "API/CNWSPlayer.hpp"
 #include "API/CNWSCreature.hpp"
 #include "API/CNWSCreatureStats.hpp"
+#include "API/CNWSMessage.hpp"
 #include "API/Types.hpp"
 #include "API/Version.hpp"
 #include "API/Constants.hpp"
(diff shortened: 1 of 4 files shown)


Plugins/Events/Events/DMActionEvents.cpp
```diff
@@ -1,36 +1,198 @@
...
+template <typename T>
+static T& PeekMessage(CNWSMessage *pMessage, int32_t offset)
 {
-    const uint32_t oldExp = stats->m_nExperience;
-    stats->SetExperience(exp, doLevel);
-    Events::PushEventData("OLD_EXP", std::to_string(oldExp));
-    Events::PushEventData("NEW_EXP", std::to_string(exp));
-    Events::SignalEvent("NWNX_ON_DM_SET_EXP", stats->m_pBaseCreature->m_idSelf);
+    uint8_t *ptr = pMessage->m_pnReadBuffer + pMessage->m_nReadBufferPtr + offset;
+    return *(T*)ptr;
```
Liareth: This is going to break strict aliasing. Maybe return T instead of T& and do

```cpp
static_assert(std::is_pod<T>::value);
T value;
std::memcpy(&value, ptr, sizeof(T));
return T;
```

mtijanic: Strict aliasing allows casting to/from `char*` (and related types, such as `uint8_t*`). But I totally did forget about it (I always compile my code with `-fno-strict-aliasing`), so definitely a good idea to change as you suggested, otherwise I'll just forget again and do something more wrong.

I went with `T&` so the hooks could in theory modify the message, but that should be done by exclusive hooks only, and there's better ways to do it.

Liareth: I think you can freely cast anything *to* a char but you can't freely cast to anything *from* a char? Something to do with alignment ... ? I can never remember the exact rules and it's too early to read standards. I try to be on the safe side and never cast to anything from a char, but I could just be a little bit mixed up in my understanding!

mtijanic: Yeah, that sounds about right. Fixed.

PR #77, opened 2018-03-19
Fix Maybe's move constructor and add a copy constructor

Liareth:


Liareth:
Well, this doesn't immediately break things, so I'm happy enough with it.

PR #83, opened 2018-03-21
Modify default Log level to SEV_INFO 

kmsheehan:
Modify default Log level to SEV_INFO to show ServerLogRedirector messages - should sweep NWNX_SQL to review INFO vs DEBUG messages as this plugin seems verbose with lots of info messages

### DIFF
--- a/Plugins/SQL/SQL.cpp
+++ b/Plugins/SQL/SQL.cpp
@@ -228,12 +228,12 @@ Events::ArgumentStack SQL::OnExecutePreparedQuery(Events::ArgumentStack&&)
         if (m_target->GetAffectedRows() >= 0)
         {
             // this was not a result set type query
-            LOG_INFO("Successful SQL query. Query ID: '%i', Query: '%s', Rows affected: '%u'.",
+            LOG_DEBUG("Successful SQL query. Query ID: '%i', Query: '%s', Rows affected: '%u'.",
                 queryId, m_activeQuery.c_str(), m_target->GetAffectedRows());
         }
         else
         {
-            LOG_INFO("Successful SQL query. Query ID: '%i', Query: '%s', Results Count: '%u'.",
+            LOG_DEBUG("Successful SQL query. Query ID: '%i', Query: '%s', Results Count: '%u'.",
                 queryId, m_activeQuery.c_str(), m_activeResults.size());
         }
     }
--- a/NWNXLib/Log.cpp
+++ b/NWNXLib/Log.cpp
@@ -91,7 +91,7 @@ Channel::Enum GetLogLevel(const char* plugin)
 {
     // TODO: Is this thread safe? I think so if we're just looking up but I don't know.
     auto entry = s_LogLevelMap.find(plugin);
-    return entry == std::end(s_LogLevelMap) ? Channel::SEV_NOTICE : entry->second;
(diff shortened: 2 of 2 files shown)


Plugins/SQL/SQL.cpp
```diff
@@ -228,12 +228,12 @@ Events::ArgumentStack SQL::OnExecutePreparedQuery(Events::ArgumentStack&&)
         if (m_target->GetAffectedRows() >= 0)
         {
             // this was not a result set type query
-            LOG_INFO("Successful SQL query. Query ID: '%i', Query: '%s', Rows affected: '%u'.",
+            LOG_DEBUG("Successful SQL query. Query ID: '%i', Query: '%s', Rows affected: '%u'.",
```
mtijanic: This shouldn't be DEBUG. At debug level, we already get all the info and much more. If it is too verbose for INFO, then just remove entirely.

kmsheehan:
Should I remove the messages then?  I don't think every successful query needs to be logged as info.

mtijanic:
I'm fine with it, but it was @Liareth who added them in the first place.

ELadner:
The user could always reduce the logging to WARN or ERROR if INFO is too verbose.

kmsheehan:
No the level cannot be reduced below info if the ServerLogRedirector is used - unless the way that logs is changed.  The ServrLogRedirector logs everything as INFO level.  I would like to see that change instead as there are two flags for the ServerLogRedireactor- one that enables it then it can be again 'disabled' by setting a log level less verbose that INFO.

Liareth:
I think logging SQL queries is pretty important. and INFO seems to me to be the ideal channel for it. Setting your log level to INFO isn't meant to be human readable at a glance on a buys server - that's what NOTICE is for, I think. In my mind, NOTICE gives you the things that need attention, INFO gives what should be logged, and DEBUG gives extra debug information to help track down bugs. You can set log levels per plugin - if you don't want SQL queries logged you can just do NWNX_SQL_LOG_LEVEL=5.

I'm opening to changing it from INFO to DEBUG if everyone disagrees with me, mind you. I just feel that it's a very common use case to log all queries, and as a developer or user I wouldn't want to be spammed with NWNX debugging information just so I can log the queries.

kmsheehan:
I'll remove the pull request until we can come to a consensus.  I understand logging queries can be important but there isn't a LOG_INFO for a Redis cache hit or each time some other regular function is called... If INFO is the correct level then the underlying issue is that the ServerLogRedirector needs to be handled as a special case.  The nwserver does not no about log levels and NWNX provides a plugin to commingle those logs with the NWNX stream.  From the server I DO want see when players log in and out but I don't need to see every database query that fires.

I did individually set the SQL level differently (WARN) to prevent that but that makes logging configuration a rather complex process for new users... This is not a required item just thought it might clear some things up.

mtijanic:
How about a config for the ServerLogRedirector plugin that sets the log level at which it will print? Something like `NWNX_SERVERLOGREDIRECTOR_DEFAULT_LEVEL=INFO`?

kmsheehan:
That can be done as well - let me test that... will just have to document things.

kmsheehan:
Not a problem... Tested this...

      NWNX_CORE_LOG_LEVEL: 5
      NWNX_SERVERLOGREDIRECTOR_LOG_LEVEL: 6

that works.

I will remove this and add some documentation to the ServerLogRedirector

PR #88, opened 2018-03-22
Remove the weird regex PIC thing.

Liareth:
Replace it with -fno-pic, which is the correct command to turn off PIC (in combination with setting
CMAKE_POSITION_INDEPENDENT_CODE which we already do). This has no functional change but it's cleaner.


mtijanic:
Preexisting, but: `-march=pentium4`? Really?

Liareth:
From back when I used to compile on a modern system but built for Arelith, which, at the time, literally was hosted on a Pentium 4. Thankfully times have changed. We can probably do -march=native now.

PR #89, opened 2018-03-22
Added compatibility with EE per-part coloring (and brief documentation :O)

BhaalM:
For example:

```C
// Make a single change to the appearance of an item. This will not be visible to PCs until
// the item is refreshed for them (e.g. by logging out and back in).
// Helmet models and simple items ignore iIndex.
// nType                            nIndex                              nValue
// ITEM_APPR_TYPE_SIMPLE_MODEL      [Ignored]                           Model #
// ITEM_APPR_TYPE_WEAPON_COLOR      ITEM_APPR_WEAPON_COLOR_*            0-255
// ITEM_APPR_TYPE_WEAPON_MODEL      ITEM_APPR_WEAPON_MODEL_*            Model #
// ITEM_APPR_TYPE_ARMOR_MODEL       ITEM_APPR_ARMOR_MODEL_*             Model #
// ITEM_APPR_TYPE_ARMOR_COLOR       ITEM_APPR_ARMOR_COLOR_* [0]         0-255 [1]
//
// [0] Alternatively, where ITEM_APPR_TYPE_ARMOR_COLOR is specified, if per-part coloring is
// desired, the following equation can be used for nIndex to achieve that:
//
// ITEM_APPR_ARMOR_NUM_COLORS + (ITEM_APPR_ARMOR_MODEL_ * ITEM_APPR_ARMOR_NUM_COLORS) + ITEM_APPR_ARMOR_COLOR_
//
// For example, to change the CLOTH1 channel of the torso, nIndex would be:
//
//     6 + (7 * 6) + 2 = 50
//
// [1] When specifying per-part coloring, the value 255 corresponds with the logical
// function 'clear colour override', which clears the per-part override for that part.
void NWNX_Item_SetItemAppearance(object oItem, int nType, int nIndex, int nValue);
```

### DIFF
--- a/Plugins/Item/Item.cpp
+++ b/Plugins/Item/Item.cpp
@@ -25,7 +25,7 @@ NWNX_PLUGIN_ENTRY Plugin::Info* PluginInfo()
 	"Functions exposing additional item properties",
 	"Various / sherincall / Bhaal",
 	"marca.argentea at gmail.com",
-	2,
+	3,
 	true
     };
 }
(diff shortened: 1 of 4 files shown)


mtijanic:
You can force a PC refresh without a login! I can check tomorrow exactly which message needs to be sent. If you're impatient, this can go in and the refresh can come as a follow up. :)

BhaalM:
I have to complete the weapon plugin so I would prefer to add this first in case I modify something of the weapon.

mtijanic:
Sure, fine by me. I haven't looked at the code yet, will do tomorrow if Liareth doesn't beat me to it.

Plugins/Item/Item.cpp
```diff
@@ -217,32 +250,48 @@ ArgumentStack Item::GetEntireItemAppearance(ArgumentStack&& args)
...
+   ArgumentStack stack;
+
+   if (auto *pItem = item(args))
+     {
+	const auto sAppString = Services::Events::ExtractArgument<std::string>(args);
+	int idx;
+	int stringPos=0;
+	char texture, part;
+	
+	if(sAppString.length()==2*142)
```
Liareth: This is going to break any existing item appearance strings, right? Normally this wouldn't be an issue but GetEntireItemAppearance/RestoreItemAppearance is tailor made for serialisation, and if any modules have used this function already to store item appearances, they will now be lost. Should we add a check for the old string length and replicate the old behaviour here too, to support reading it back in?

BhaalM:
Seems good to me.

BhaalM:
Hmm.. the only changes are in function RestoreItemAppearance, sorry for the big diff report but I've changed my editor to Sublime.

Liareth:
No problem, the indentation is more readable now! (2 spaces too few! But I won't complain ...)

Liareth:
Code seems good to me. Thanks!

PR #109, opened 2018-03-31
Add NWNX_Tweaks

Liareth:
This just comes with one thing so far - HIDE_CLASSES_ON_CHAR_LIST, which hides classes on the character list (default to off).

### DIFF
--- a/Plugins/Tweaks/Tweaks/HideClassesOnCharList.cpp
+++ b/Plugins/Tweaks/Tweaks/HideClassesOnCharList.cpp
@@ -1,0 +1,64 @@
+#include "Tweaks/HideClassesOnCharList.hpp"
+
+#include "Services/Hooks/Hooks.hpp"
+
+#include "API/CAppManager.hpp"
+#include "API/CNWSCreature.hpp"
+#include "API/CNWSMessage.hpp"
+#include "API/CServerExoApp.hpp"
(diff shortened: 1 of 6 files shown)


Plugins/Tweaks/Tweaks/HideClassesOnCharList.cpp
```diff
@@ -0,0 +1,64 @@
...
+
+            thisPtr->WriteBYTE(0, 8);
+        }
+        else
+        {
+            return 0;
+        }
+    }
+
+    unsigned char* message;
```
mtijanic: Nittiest of the nits, but to me `char` suggests a human readable string. The message is just a byte buffer, so `uint8_t` is a better fit semantically.

mtijanic:
This looks great. And thanks for doing it properly with a full implementation, and not a hacky inline patch or a message buffer scrub.

Plugins/Tweaks/Tweaks/HideClassesOnCharList.cpp
Liareth: I agree, `uint8_t` does fit better. Or even `std::byte` (if we had C++17). I just copied the API here, but I haven't done the same for `Types::PlayerID playerId` and `Types::ObjectID charId` so I'll change it.

PR #117, opened 2018-04-04
Weapon plugin update

BhaalM:
Added debug messages and:
```
// Set nFeat as greater weapon specialization feat for nBaseItem
void NWNX_Weapon_SetGreaterWeaponSpecializationFeat(int nBaseItem, int nFeat);

// Set nFeat as greater weapon focus feat for nBaseItem
void NWNX_Weapon_SetGreaterWeaponFocusFeat(int nBaseItem, int nFeat);
```

### DIFF
--- a/Plugins/Weapon/Documentation/README.md
+++ b/Plugins/Weapon/Documentation/README.md
@@ -35,3 +35,5 @@ NWNX_Weapon_SetEpicWeaponSpecializationFeat | Associate an epic weapon specializ
 NWNX_Weapon_SetEpicWeaponOverwhelmingCriticalFeat | Associate an epic weapon overwhelming critical feat to a weapon 
 NWNX_Weapon_SetEpicWeaponDevastatingCriticalFeat | Associate an epic weapon devastating critical feat to a weapon 
 NWNX_Weapon_SetWeaponOfChoiceFeat | Associate a weapon of choice feat to a weapon 
+NWNX_Weapon_SetGreaterWeaponFocusFeat | Associate a greater weapon focus feat (+1 attack bonus) to a weapon
+NWNX_Weapon_SetGreaterWeaponSpecializationFeat | Associate a greater weapon specialization feat (+2 damage bonus) to a weapon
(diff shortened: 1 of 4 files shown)


Liareth:
I think there's a bit of a bug here but I'm not sure ... let's assume rapier.

GWF: Rapier drives rapier by default. So, let's say you have GWF: Rapier. You gain +1 ab.
Then you change it so GWF: Longsword drives rapier.

So you come in here:

```int nBonus = plugin.m_GetMeleeAttackBonusHook->CallOriginal<int32_t>(pStats, bOffHand, bIncludeBase, bTouchAttack);```

This value already includes the extra +1 ab AB from the rapier focus because the original function accounts for that. Then we get here:

```+   if(feat>-1 && pStats->HasFeat(feat))
+   {
+      return nBonus+=plugin.m_GreaterFocusAttackBonus;
+   }
```

And we double up on the extra AB.

Similarly, just setting rapier focus to GWF: Rapier, will also trigger the doubling behaviour.

Am I misunderstanding or is this normal / expected?

BhaalM:
I'm not sure to understand you.

Greater Weapon Focus is a feat that requires Weapon Focus, for example, extracted from: http://www.d20srd.org/srd/feats.htm#greaterWeaponFocus

> Greater Weapon Focus [General]
> Choose one type of weapon for which you have already selected Weapon Focus. You can also choose unarmed strike or grapple as your weapon for purposes of this feat.
> 
> Prerequisites
> Proficiency with selected weapon, Weapon Focus with selected weapon, fighter level 8th.
>
> Benefit
>You gain a +1 bonus on all attack rolls you make using the selected weapon. This bonus stacks with other bonuses on attack rolls, including the one from Weapon Focus (see below).

In your feats.2da you should add WF as a prerequisite for GWF, so you will have a +1 from weapon focus a +1 from.

And for the the GWF longsword, why are you creating two GWF custom feats and apply them to the same weapon?

BhaalM:
Correct, but the GWF is not added on the original call:

- From the original call tou have a +1 from weapon focus (but not GWF)
- And the if you have GWF you add a +1 as intended

BhaalM:
Sorry I clicked the "close and comment button"...

PR #123, opened 2018-04-08
Add C# scripting support with Mono

Liareth:
This PR isn't for merging yet. It's to track my progress and get feedback in early. This is now in a semi-usable state so I wanted to open it up.

- To build and use, just install mono-complete on Debian 9.4.
- To compile code with Mono, do this: `csc -platform:x86 -t:library -out:test.so Internal.cs Types.cs nwscript.cs m_hb.cs`
- Then load it by doing this: `export NWNX_MONO_ASSEMBLY="m_hb.so"`, if m_bh.so is in the working directory.
- Where m_hb is set to the module heartbeat script, and m_hb.cs is:

```csharp
using System;

namespace NWN.Scripts
{
    class m_hb
    {
        static void Main()
        {
            Console.WriteLine("I wrote this line from Mono!");
            NWScript.SendMessageToPC(NWScript.GetFirstPC(), "Rawr! It is me! Mario");
        }
    }
}
```

You get this:

https://i.imgur.com/PRCUGgl.png

The principle is pretty simple. When the game tries to run a script, if a class with that name containing a `static void Main()` exists in the assembly, it is called instead. Some things are untested. Actions aren't supported. Using game defined structures might memory leak or crash the server. Proceed with caution.

WIP tooling exist to convert .nss to .cs. See here: https://github.com/Liareth/nss2csharp


Liareth:
This is getting very close to merge-ready now.

Liareth:
This is missing only StartingConditional support now. I think it's pretty usable. Gonna hit the merge button and put out any fires later.

PR #131, opened 2018-04-15
Fix infinite recursion when signals are overwritten

Liareth:
- NWNX overwrites signals and chains
- Mono overwites signals and chains
- Crash happens
- Mono does its thing
- NWNX does its thing - ASSERT_FAIL_MSG causes `std::abort()`
- Mono does its thing ...
- NWNX does its thing ...

This fixes the above scenario by writing to stderr rather than asserting when we encounter a signal.

### DIFF
--- a/Core/NWNXCore.cpp
+++ b/Core/NWNXCore.cpp
@@ -34,12 +34,15 @@ extern "C" void nwnx_signal_handler(int sig)
         default:       err = "Unknown error";            break;
     }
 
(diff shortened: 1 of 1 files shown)


mtijanic:
The assert was primarily so we get a callstack. This works, but be sure to also add:
```C
std::fputs(NWNXLib::Platform::Debug::GetStackTrace(20).c_str(), stderr);
```

PR #132, opened 2018-04-15
Revised ClearMemorisedSpell()

nwngracken:
Last implemention called SetMemorisedSpell() with a
struct NWNX_Creature_MemorisedSpell spell parameter
whose id field was set to -1.  That implementation was
based off of legacy nwnx implentation. It seemed to
work, but was hacky.  This implemenation should use
internals correctly.


nwngracken:
This commit was tested.  Use of the ClearMemorisedSpell() function removed memorized spells as expected.

PR #134, opened 2018-04-15
Added ClearMemorisedKnownSpells() function

nwngracken:
That removes a specific spell from a creature's spellbook
by class.


nwngracken:
Tested; Was able to remove "endure elements" and "shield of faith" from a cleric.

PR #136, opened 2018-04-17
Added Desvastating Critical Event Script

BhaalM:
```
// Set Devastating Critical Event Script
void NWNX_Weapon_SetDevastatingCritalEventScript(string sScript);

// Get Devastating Critical Event Weapon (to use only on Devastating Crital Event Script)
object NWNX_Weapon_GetDevastatingCritalEventWeapon();

// Get Devastating Critical Event Target (to use only on Devastating Crital Event Script)
object NWNX_Weapon_GetDevastatingCritalEventTarget();

// Get Devastating Critical Event Damage (to use only on Devastating Crital Event Script)
int NWNX_Weapon_GetDevastatingCritalEventDamage();

// Bypass Devastating Crtical (to use only on Devastating Crital Event Script)
void NWNX_Weapon_BypassDevastatingCritical();
```


Plugins/Weapon/Weapon.cpp
```diff
@@ -490,9 +550,34 @@ int32_t Weapon::GetEpicWeaponDevastatingCritical(NWNXLib::API::CNWSCreatureStats
...
+       
+      plugin.m_DCData.oidWeapon = pWeapon->m_idSelf;
+      plugin.m_DCData.oidTarget = pCreature->m_oidAttackTarget;
+      plugin.m_DCData.nDamage   = pAttackData->GetTotalDamage(1); 
+      plugin.m_DCData.bBypass   = false;
+      
+      NWNXLib::API::CExoString script = plugin.m_DCScript.c_str();
+
+      LOG_DEBUG("Devastating Critical Event Prev Call Script %s", plugin.m_DCScript.c_str());
+      Globals::VirtualMachine()->RunScript(&script, pCreature->m_idSelf, 1);
```
mtijanic: You can just do `Utils::ExecuteScript(plugin.m_DCScript, pCreature->m_idSelf)`. Include "Utils.hpp", no need to include the VM stuff

Plugins/Weapon/Weapon.cpp
```diff
@@ -342,6 +348,59 @@ ArgumentStack Weapon::SetOption(ArgumentStack&& args)
    return stack;   
 }
 
+ArgumentStack Weapon::SetDevastatingCritalEventScript(ArgumentStack&& args)
+{
+   ArgumentStack stack;
+   
+   m_DCScript = Services::Events::ExtractArgument<std::string>(args);
+   LOG_DEBUG("Set DevastatingCritalEventScript to %s", m_DCScript.c_str());
```
mtijanic: I'd say this warrants an INFO print.

Plugins/Weapon/Weapon.hpp
```diff
@@ -13,6 +13,20 @@
 #define NWNX_WEAPON_OPT_GRTFOCUS_AB_BONUS 0
 #define NWNX_WEAPON_OPT_GRTSPEC_DAM_BONUS 1
 
+#define NWNX_WEAPON_GETDATA_DC_WEAPON  0
+#define NWNX_WEAPON_GETDATA_DC_TARGET  1
```
mtijanic: Preexisting, but I'd say favor `const int` or `constexpr int` over defines. This can be confusing, but consts in C++ are different than consts in C, they are much more like defines with type and syntax safety.

Plugins/Weapon/NWScript/nwnx_weapon.nss
```diff
@@ -48,6 +56,21 @@ void NWNX_Weapon_SetWeaponIsMonkWeapon(int nBaseItem);
 // Set plugin options
 void NWNX_Weapon_SetOption(int nOption, int nVal);
 
+// Set Devastating Critical Event Script
+void NWNX_Weapon_SetDevastatingCritalEventScript(string sScript);
+
+// Get Devastating Critical Event Weapon (to use only on Devastating Crital Event Script)
+object NWNX_Weapon_GetDevastatingCritalEventWeapon();
```
mtijanic: Would it maybe make more sense to just return the struct, instead of having separate function calls?
Example: https://github.com/nwnxee/unified/blob/master/Plugins/Player/NWScript/nwnx_player.nss#L109

mtijanic:
Overall looks great, minor comments in the code.

Liareth:
Great work! Let me know if you're going to address any of @mtijanic's comments, or if you want me to merge this as it is. I'm happy either way.

BhaalM:
I'll change the code and add the proposed modifications. Thank you for your comments!

BhaalM:
Ready to merge (if you agree of course)

PR #141, opened 2018-04-19
Added support for iso 8859-1 encoding, slack username and mrkdwn.

BhaalM:


BhaalM:
It seems Slack standard mrkdwn feature is not compatible with discord, so I'm using "backslash". The code should be compatible with previous version of the plugin

Plugins/WebHook/WebHook.cpp
```diff
@@ -92,4 +115,21 @@ Events::ArgumentStack WebHook::OnSendWebhookHTTPS(Events::ArgumentStack&& args)
     return Events::ArgumentStack();
 }
 
+std::string WebHook::iso_8859_1_to_utf8(std::string &str)
+{
+    std::string strOut;
+    for (std::string::iterator it = str.begin(); it != str.end(); ++it)
+    {
+        uint8_t ch = *it;
+        if (ch < 0x80) {
```
Liareth: Needs new line!

Plugins/WebHook/WebHook.cpp
```diff
@@ -92,4 +115,21 @@ Events::ArgumentStack WebHook::OnSendWebhookHTTPS(Events::ArgumentStack&& args)
...
+            strOut.push_back(ch);
+        }
+        else {
```
Liareth: Also here!

Liareth:
Let me give this a shot tomorrow to make sure it still works in my use case. Btw - what about Discord wasn't working for you? I was able to send WebHooks to Discord just fine using this plugin.

BhaalM:
Sorry for the missing new lines, I copied the code from another file....

The WebHook works well, what is not working is the "mrkdwn", please look at the slack API here:

https://api.slack.com/docs/messages/builder?msg=%7B%0A%20%20%22text%22%3A%20%22*bold*%20%60code%60%20_italic_%20~strike~%22%2C%0A%20%20%22username%22%3A%20%22markdownbot%22%2C%0A%20%20%22mrkdwn%22%3A%20true%0A%7D

If you set markdwn to 0, the formatting should disappear but Discord seems to ignore this Slack parameter.

Liareth:
Ahh I see, gotcha! Thanks for the clarification. The code looks good to me actually so once you fix up the new lines I'll merge it. Great work :)

PR #145, opened 2018-04-26
New plugin: NWNX_Spell

Morderon:
Uses Hunspell 1.4.1 or above

See readme.md for anything more


Liareth:
This definitely should not be called NWNX_Spell ... maybe NWNX_SpellChecker, or NWNX_Hunspell?

Plugins/Spell/Spell.cpp
```diff
@@ -0,0 +1,182 @@
...
+            sc = Spell::spell_e(Spell::created, word.c_str());
+            if(sc == 0)
+              output += word + ",";         
+        }
+
+        
+    }
+  
+    Spell::dest_e(Spell::created);
+    dlclose(Spell::handle);
```
Liareth: I don't think you want to be closing this after each event - after all, you only open it once, in the Init function. You probably want to close it in the destructor.

Liareth: Oh wait, scratch that, you call Init every time ... is that a good idea? Seems like you should only initialise it once in the constructor, and free it in the destructor.

Plugins/Spell/Spell.cpp
```diff
@@ -0,0 +1,182 @@
...
+    if(!word.empty())
+    {
+        cword = word.c_str();
+        sc = Spell::spell_e(Spell::created, cword);
+        if(sc == 0)
+        {
+            char** wlst;
+           int ns = Spell::suggest_e(Spell::created, &wlst, cword);
+            for (int i = 0; i < ns; i++) {
+                output += (std::string)wlst[i]  + ",";
```
Liareth: Does casting a char* to an std::string even work? If it does, how ... ?

Liareth:
Code looks good to me. I haven't tested it, but I'm sure you'll fix any issues that arise! Thanks!

PR #149, opened 2018-04-30
Emergency fix: Overwrite old BioWare hardcoded address if DNS resolut…

Liareth:
…ion fails.


Liareth:
Merging now, will handle review later.

PR #153, opened 2018-05-05
New damage plugin

BhaalM:
Ported from Baaleos nwnx2 damage plugin.

### DIFF
--- a/Plugins/Damage/Damage.cpp
+++ b/Plugins/Damage/Damage.cpp
@@ -1,8 +1,3 @@
-
-// Log currently generates warnings when no arguments are given to format string
-// TODO: Should really clean up the log so it doesn't warn in these cases
-#pragma GCC diagnostic ignored "-Wformat-security"
-
 #include "Damage.hpp"
 #include "API/Functions.hpp"
 #include "API/Constants.hpp"
--- a/Plugins/Damage/Damage.cpp
+++ b/Plugins/Damage/Damage.cpp
@@ -1,0 +1,142 @@
+
+// Log currently generates warnings when no arguments are given to format string
+// TODO: Should really clean up the log so it doesn't warn in these cases
+#pragma GCC diagnostic ignored "-Wformat-security"
+
+#include "Damage.hpp"
+#include "API/Functions.hpp"
+#include "API/Constants.hpp"
+#include "API/CGameEffect.hpp"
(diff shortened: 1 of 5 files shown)


Plugins/Damage/Damage.cpp
```diff
@@ -0,0 +1,142 @@
+
+// Log currently generates warnings when no arguments are given to format string
+// TODO: Should really clean up the log so it doesn't warn in these cases
+#pragma GCC diagnostic ignored "-Wformat-security"
```
mtijanic: This can be removed now :)

BhaalM:
I've seen that the event plugin is usually for shared hooks this one require exclusive one to change the damage values... I thought this was not the idea for the Event plugin. But, as you wish.

The problem with the damage event is that you cannot change the damage inflicted to a creature. You can heal a creature after been damaged, or apply a new bonus damage... but you cannnot directly set the amount of damage or what is displayed on the combat log.

mtijanic:
Ack on both accounts. I still think it belongs in events, but it would actually require extending that plugin to support passing the data back (I'm not sure if NWNX_Events_PushEventData() would cover it).

Christian-Vassallo:
It belong to nwnx_damages.

It is different because it allows to modify the damages and we can think it as a layer between the system and the creature (when the creature get damaged).

Nwnx events instead is a lauer between the creature and the system (when you fire an action)

Moreover, in my opinion, for a back-compatibility of concepts, nwnx_damages should stay as a plugin separated :)

Plugins/Damage/Damage.hpp
```diff
@@ -0,0 +1,36 @@
...
+namespace Damage {
+
+class Damage : public NWNXLib::Plugin
+{
+   	public:
+    	Damage(const Plugin::CreateParams& params);
+      	virtual ~Damage();
+
+   	private:      
+   		ArgumentStack SetDamageEventScript(ArgumentStack&& args);
```
Liareth: These should be spaces rather than tabs.

Liareth:
I think this would be better as an event with the ability to set the damage value that goes through, but that would require changing the events plugin, and to be honest ... where it is is much less important than if it is. If we ever decide to make it an even we can just deprecate this.

Awesome work @BhaalM!

PR #155, opened 2018-05-06
Two fixes to Mono stuff; Handle empty string in StackPopString() and handle character encoding. Mono assumes UTF-8, while NWN strings are in CP-1252 aka ANSI. (Also sneakily added an extra path to FindMySQL)

tzaeru:
The function added, ISO88959ToUTF8, is defined in the bottom of Mono_Handlers.cpp.

I could use a sanity check for it.

To convert to UTF-8 from ISO-88959, we take a character array in 88959 and create an array twice its size, as that's the maximum size the converted array could have in theory.

When a character code is at or above 0x80, we need to split the character to two bytes. The first byte is c2 is for characters from 0x80 to 0xc0. For characters between 0xc0 and 0xff, the byte is c3. The second byte goes from 0x80 to 0xbf for both quarters. This byte layout is easily seen here: http://www.utf8-chartable.de/

This seems to have fixed my issues and now I can use "¦" as a data separator.

..Also, I added a path to FindMySQL.

### DIFF
--- a/Plugins/Mono/Mono_Handlers.cpp
+++ b/Plugins/Mono/Mono_Handlers.cpp
@@ -244,7 +244,11 @@ MonoString* StackPopString()
     }
 
     LOG_DEBUG("Popped string '%s'.", value.m_sString);
-    return mono_string_new(g_Domain, value.m_sString);
+
+    if (value.m_sString != nullptr)
+        return mono_string_new(g_Domain, ISO88959ToUTF8(value.m_sString).get());
+    else
+        return mono_string_new(g_Domain, "");
 }
 
 uint32_t StackPopObject()
@@ -394,6 +398,33 @@ int32_t ClosureActionDoCommand(uint32_t oid, uint64_t eventId)
     }
 
     return 0;
+}
+
+std::shared_ptr<char> ISO88959ToUTF8(const char *str)
+{
+    char* utf8 = new char[1 + (2 * strlen(str))];
+
+    int len = 0;
+
+    if (utf8) {
+        char *c = utf8;
+        for (; *str; ++str) {
+            if (!(*str & 0x80)) {
+                *c++ = *str;
+                len++;
+            } else {
+                *c++ = (char) (0xc2 | ((unsigned char)(*str) >> 6));
+
+                *c++ = (char) (0x3f & *str);
+                len += 2;
+            }
+        }
+        *c++ = '\0';
+    }
+
+    realloc(utf8, len);
+
+    return std::shared_ptr<char> (utf8);
 }
 
 }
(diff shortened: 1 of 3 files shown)


Plugins/Mono/Mono_Handlers.cpp
```diff
@@ -396,4 +400,31 @@ int32_t ClosureActionDoCommand(uint32_t oid, uint64_t eventId)
...
+                *c++ = (char) (0xc2 | ((unsigned char)(*str) >> 6));
+
+                *c++ = (char) (0x3f & *str);
+                len += 2;
+            }
+        }
+        *c++ = '\0';
+    }
+
+    realloc(utf8, len);
```
mtijanic: Mixing `realloc` with `new` is not a good idea. glibc implements them both with the same allocator, so it will work in most cases, but that can change between versions too. Just use `new`.. If the memory usage is problematic, you'll need a secondary `new`'d buffer, then manually memmove the data over.
But I think a wee bit of extra usage is no big deal.

Plugins/Mono/Mono_Handlers.cpp
```diff
@@ -244,7 +244,11 @@ MonoString* StackPopString()
     }
 
     LOG_DEBUG("Popped string '%s'.", value.m_sString);
-    return mono_string_new(g_Domain, value.m_sString);
+
+    if (value.m_sString != nullptr)
+        return mono_string_new(g_Domain, ISO88959ToUTF8(value.m_sString).get());
```
mtijanic: Won't this leak the memory? When is it freed?

Plugins/Mono/Mono_Handlers.cpp
```diff
@@ -244,7 +244,11 @@ MonoString* StackPopString()
     }
 
     LOG_DEBUG("Popped string '%s'.", value.m_sString);
-    return mono_string_new(g_Domain, value.m_sString);
+
+    if (value.m_sString != nullptr)
+        return mono_string_new(g_Domain, ISO88959ToUTF8(value.m_sString).c_str());
```
mtijanic: In the interest of brevity:
`return mono_string_new(g_Domain, ISO88959ToUtf(value.CStr()).c_str());`

CExoString::CStr() does the check and returns `""` in case of `nullptr`. This is purely aesthetics, so don't worry about it, but I expect @Liareth will require a few more aesthetic changes so good time to sneak this one in :)

Plugins/Mono/Mono_Handlers.cpp
```diff
@@ -396,4 +400,20 @@ int32_t ClosureActionDoCommand(uint32_t oid, uint64_t eventId)
     return 0;
 }
 
+std::string ISO88959ToUTF8(const char *str)
+{
+    std::string utf8("");
+    utf8.reserve(2*strlen(str) + 1);
+
+    for (; *str; ++str) {
```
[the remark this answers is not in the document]
tzaeru: Actually that's my preferred style too. Slipped in since the original function was copypaste, though I then discovered it was written totally wrong in respects to the UTF-8 spec and needed to redo it.

Liareth:
Thanks!

PR #157, opened 2018-05-07
Update tweaks plugin README

weavejester:
The environment variables in the tweaks plugin README seem to be incomplete. At least in build 8166-1 they require a `NWNX_TWEAKS_` prefix to work.

### DIFF
--- a/Plugins/Tweaks/Documentation/README.md
+++ b/Plugins/Tweaks/Documentation/README.md
@@ -6,6 +6,6 @@ Tweaks stuff. See below.
 
 ## Environment Variables
 
-HIDE_CLASSES_ON_CHAR_LIST: true or false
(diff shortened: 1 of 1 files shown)


zunath:
The plugin prefix is a requirement across the board, as far as I know. If we're going to change this documentation to reflect that maybe we should update all of the other documentation too?

weavejester:
The [SQL plugin docs](https://github.com/nwnxee/unified/tree/master/Plugins/SQL/Documentation) include the prefix, and as a newcomer I found that to be the more useful format. More redundant certainly, but if there's nothing in the plugin README that mentions the prefix, it's easily omitted.

zunath:
Agreed. I don't think we run the risk of "over documentation" :)

PR #161, opened 2018-05-07
./Scripts/buildnwnx.sh -l

Morderon:
The -l Flag will force cmake to locate 32-bit libraries/packages.
Allows the 64-bit versions of the libraries to be installed without confusion during the building process . For example the 64-bit library counting as the package even if the 32-bit library is required for the plugin.

Decided on a flag as obviously people have been being able to build fine without it.


Liareth:
Seems like this should be the default, and not behind a flag ... I'll add a bug for it.

PR #169, opened 2018-05-11
NWNX_Lua

dparoli:
Updated version of old nwnx_lua plugin 
needs luajit or a regular libluadev, better with luajit


Liareth:
Looks great! Awesome work!

PR #196, opened 2018-06-09
new utility plugin with misc functions

mtijanic:
Need some dumping ground for these things, so a new plugin. Maybe NWNX_Misc would have been a better name?

For now just the things I actively need, and I don't want to maintain a separate fork..


Liareth:
We could add an assert function, a debugbreak function, and a 'write to console' function here too. Good idea!

PR #197, opened 2018-06-10
plugins/jvm: don't memleak popped CGameEffects and ScriptLocations

niv:
Something I noticed while browsing the new nwnxee codebase.

Caveat: I haven't actually tried running this patch.

CScriptLocation and CGameEffect (and any other engine structures) are allocated on the nwn core side. NWNX_JVM would not free them again. This means that every script call touching these variable types would leak a small amount of memory.

Pretty sure all other language plugins suffer the same issue. I had a quick browse on the Mono plugin and I didn't find where it would release the allocated CGameEffects, for example. JVM would solve this by tracking all requested CGameEffect instances and then, with this patch, free them after we drop out of the script context. An alternative would be to store all CGameEffect internal data in a JVM/Mono class and let those VMs handle it, but that seems needlessly interlinked.

Maybe someone using this plugin (or Mono, I guess) can have a look.

### DIFF
--- a/Plugins/JVM/Internal.cpp
+++ b/Plugins/JVM/Internal.cpp
@@ -8,6 +8,7 @@
 
 #include "API/Functions.hpp"
 #include "API/CNWVirtualMachineCommands.hpp"
+#include "API/CGameEffect.hpp"
 
 using namespace NWNXLib;
 using namespace NWNXLib::API;
@@ -165,6 +166,11 @@ ArgumentStack Internal::EntryPoint(jmethodID method, ArgumentStack& args)
 
     // reset event mappings
     if (m_contextDepth == 0) {
+        for (const CGameEffect* eff : m_touchedEffects)
+        {
+            delete eff;
+        }
+
         m_touchedEffects.clear();
     }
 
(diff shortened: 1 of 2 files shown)


dparoli:
In the Lua plugin there is already defined a __gc() methametod for each engine structure. With this metamethod in place the Lua (or Luajit) garbage collector takes care of them and when an engine structure goes out of scope or is left without any reference is automatically freed.

Liareth:
Structures are freed here: https://github.com/nwnxee/unified/blob/master/Plugins/Mono/CSharpAPI/INTERNAL_Types.cs

CVirtualMachineScript is freed here: https://github.com/nwnxee/unified/blob/master/Plugins/Mono/Mono_Handlers.cpp

Liareth:
(somebody should try running this PR before I merge it)

dparoli:
I also used `CNWVirtualMachineCommands::DestroyGameDefinedStructure`, in fact I copied a lot from the Mono plugin. I made some tests when I was developing and the mem increase is noticeable after you popped some thousands CGameEffects without freeeing them. But after I added the __gc() metamethod (garbage collector) to free engine structures, i.e. something like the descructors of Mono, the memory usage does not increase.

dparoli:
I am not a java expert but I think it could be used the Object#finalize() method to call CNWVirtualMachineCommands::DestroyGameDefinedStructure. For what I read the finalize() method is called by the JVM garbage collector when an oject is at the end of life; i.e without references left.

PR #202, opened 2018-06-13
Skip new plugins Dialog and Util

jakkn:


mtijanic:
I wonder if it would not be better to just have a `NWNX_CORE_SKIP_PLUGINS_BY_DEFAULT` variable that inverts the behavior..

jakkn:
Agreed.

mtijanic:
I'm okay with merging this for now, and if no one does it before I'll add the change to core on the weekend.

PR #205, opened 2018-06-27
Exposed a function to get the name of the bic resource associated with the player

tzaeru:
The BIC filename is useful for e.g. moving character files.

### DIFF
--- a/Plugins/Player/Player.cpp
+++ b/Plugins/Player/Player.cpp
@@ -307,7 +307,7 @@ ArgumentStack Player::GetBicFileName(ArgumentStack&& args)
     ArgumentStack stack;
     if (auto *pPlayer = player(args))
     {
-        Services::Events::InsertArgument(stack, std::string(pPlayer->m_resFileName.GetResRef()));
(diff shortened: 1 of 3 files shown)


Plugins/Player/Player.cpp
```diff
@@ -301,4 +302,14 @@ ArgumentStack Player::SetQuickBarSlot(ArgumentStack&& args)
     return stack;
 }
 
+ArgumentStack Player::GetBicFileName(ArgumentStack&& args)
+{
+    ArgumentStack stack;
+    if (auto *pPlayer = player(args))
+    {
+        Services::Events::InsertArgument(stack, std::string(pPlayer->m_resFileName.GetResRef()));
```
Liareth: This will fail in some circumstances- resrefs are 16 characters long and are not guaranteed to be null terminated. You could try this instead:

```c++
Services::Events::InsertArgument(stack, std::string(pPlayer->m_resFileName.GetResRef(), pPlayer->m_resFileName.GetLength()));
```

PR #208, opened 2018-07-13
Redis decrement float

Urothis:
Probably a small piece needed.


Urothis:
I'm dumb, you just use negatives.

PR #209, opened 2018-08-08
creature: Add LevelUp and LevelDown functions

mtijanic:
LevelUp() is similar to LevelUpHenchman except it does not do any validation at all, so you can do things like change the class at level 1, then add more levels of that class, etc. Also does not care about the creature experience, server level settings and the like.

LevelDown() is a simple way to permanently strip levels from a creature. For PCs, you're better off taking the XP instead.


mtijanic:
(Hold off on merging, eyeso reported some issues that need further fixing)

mtijanic:
Okay, good to go now. I added `SetLevelByPosition()` which can be used for creatures that didn't go through the levelup process.
Also added a SetCHallengeRating rider cause I didn't want to deal with merge conflicts from different branches :)

PR #211, opened 2018-08-09
time: Add GetHighResTimestamp function

mtijanic:
Similar format to `gettimeofday`. Microsecond precision is good enough to measure most of the nwscript stuff. Creating and destroying a creature takes about a 100us on my machine.


Liareth:
Not that it matters much, but worth noting - `system_clock` has 1 millisecond precision on Windows.

mtijanic:
Ah. Well, `high_resolution_clock` works just as well I guess. Not that it matters currently.

PR #217, opened 2018-08-29
Damage: Allow specifying the damage script per-object

mtijanic:
Preserving the nwscript interface, so old code will still work (though you will get a warning if you didn't recompile), but you can specify a per-object script that takes priority over the global one.

Plus a bugfix in the NSS.

@BhaalM please review :)

### DIFF
--- a/Plugins/Damage/Damage.cpp
+++ b/Plugins/Damage/Damage.cpp
@@ -7,7 +7,10 @@
 #include "API/CAppManager.hpp"
 #include "API/CServerExoApp.hpp"
 #include "API/CNWSCreature.hpp"
+#include "Services/PerObjectStorage/PerObjectStorage.hpp"
 #include "Utils.hpp"
+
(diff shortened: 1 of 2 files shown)


Plugins/Damage/Damage.cpp
```diff
@@ -61,11 +64,39 @@ Damage::~Damage()
...
-   return stack;   
+   Types::ObjectID oidOwner = Constants::OBJECT_INVALID;
+   std::string script = Services::Events::ExtractArgument<std::string>(args);
+
+   // If compiled with old NSS, they won't be pushing the object after the script, so need to handle that.
+   try
+   {
+      oidOwner = Services::Events::ExtractArgument<Types::ObjectID>(args);
+   }
+   catch(std::runtime_error e)
```
Liareth: You probably want to capture the exception by (const) reference in the future.

PR #220, opened 2018-09-04
Healer's kit event

BhaalM:
- Added the Healer's kit event. The event can be bypassed.
- Added the skip event function (a "copy" of @Liareth skip event code, hope you don't mind)

### DIFF
--- a/Plugins/Events/NWScript/nwnx_events.nss
+++ b/Plugins/Events/NWScript/nwnx_events.nss
@@ -53,6 +53,12 @@ string NWNX_Events_GetEventData(string tag);
 // THIS SHOULD ONLY BE CALLED FROM WITHIN AN EVENT HANDLER.
 void NWNX_Events_SkipEvent();
 
+// Set the return value of the event.
+//
+// THIS SHOULD ONLY BE CALLED FROM WITHIN AN EVENT HANDLER.
+void NWNX_Events_SetEventResult(string data);
+
+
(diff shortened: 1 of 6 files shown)


Plugins/Events/NWScript/nwnx_events.nss
```diff
@@ -41,6 +42,23 @@ int NWNX_Events_SignalEvent(string evt, object target);
...
+// Skips execution of the currently executing event.
+// If this is a NWNX event, that means that the base function call won't be called.
+// This won't impact any other subscribers, nor dispatch for before / after functions.
+// For example, if you are subscribing to NWNX_ON_EXAMINE_OBJECT_BEFORE, and you skip ...
+// - The other subscribers will still be called.
+// - The original function in the base game will be skipped.
+// - The matching after event (NWNX_ON_EXAMINE_OBJECT_AFTER) will also be executed.
+//
+// THIS SHOULD ONLY BE CALLED FROM WITHIN AN EVENT HANDLER.
+void NWNX_Events_SkipEvent();
```
Liareth: We should document that this only works for the healing kit event here.

BhaalM: Yes, it should even be better if we add the SkipEvent() to the other events :D

Liareth:
Sorry for the slow merge. Looks good to me.

PR #221, opened 2018-09-05
mono: Add utf8->iso8859 conversion too, not just the other way.

mtijanic:
This fixes issues with >127 character points needed for color tokens.

Relying on C# to do the encoding doesn't _quite_ work:
Encoding.ASCII

With this change, passing direct C# utf8 strings:
direct string push

Also fix typo in standard name in previous function


Liareth:
I have no idea what this magic is but seems legit.

PR #223, opened 2018-09-10
pos: Use c++ casts like a normal person

mtijanic:
CNWSArea uses multiple inheritance so a C style cast is bad. Not sure why I used those in the first place, maybe trying to cut down on the includes?

Anyway, this fixes a crash when destroying areas. Testing:

```c
DestroyArea(CopyArea(GetFirstArea()));
```

### DIFF
--- a/NWNXLib/Services/PerObjectStorage/PerObjectStorage.cpp
+++ b/NWNXLib/Services/PerObjectStorage/PerObjectStorage.cpp
@@ -1,5 +1,7 @@
 #include "Services/PerObjectStorage/PerObjectStorage.hpp"
 #include "API/CGameObject.hpp"
+#include "API/CNWSArea.hpp"
+#include "API/CNWSObject.hpp"
 #include "API/Constants.hpp"
 
 namespace NWNXLib {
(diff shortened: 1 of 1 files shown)


Liareth:
I don't understand how this works ...

The only difference between a c-style cast and dynamic_cast is that dynamic_cast will return null if the cast isn't valid. From what I see here, `DestroyObjectStorage` doesn't guard against the possible null pointer, so in the best case, the functionality should be identical before and after this change.

Even if the code did guard against null, this shouldn't work. The type info is usually stored alongside the vtable. Because we (in our code, with our version of GCC and its corresponding ABI) didn't allocate whatever pThis is, we can't guarantee that the location of the type information is the same. For the sake of argument, let's say that it is -  the internal representation of the type would surely mismatch between objects allocated by NWN and objects allocated by us, since `NWNXLib::API::CGameObject` (our code) is not the same type as `CGameObject` (NWN code), nor will the types be compiled in the same order as NWN code, etc.

NWN compiles with RTTI turned off which means that dynamic_cast could never work on an object allocated in their code even if the stars aligned on all of the above.

Could it be that the crash you're seeing is actually because `API::CNWSObject* pThis` is null due to a GCC destructor implementation detail? Maybe what's happening is that dynamic_cast is throwing an exception because of this which is being caught elsewhere, so the per-object storage is never actually freed but the crash doesn't occur with your change?

mtijanic:
So, we have this marvelous class:
```C
struct CNWSArea
    : CNWArea, CResHelperTemplatedCResARE2012u, CGameObject
```

That's proper multiple inheritance, not "interface" style. For single inheritance (and interface style implementations), the upcast should be just a reinterpret cast, which is what the C-style cast was doing. But since we inherit the kitchen sink here, the cast was bad. Dynamic cast, even without RTTI knows to actually check the structure definition and deal with the other base classes that get in the way.

It is quite likely that a static_cast would do the same thing, actually. I can try that tonight and amend if it would make it more obvious. I just did the first thing that came to mind, tested and sent :).

Actually, a **horrifying** possibility is that even the C-cast would work once you include the full definition. I really hope that's not the case.

mtijanic:
Yep :(

http://eel.is/c++draft/expr.cast

TL;DR: C-style cast will perform a static_cast if the definition is available, and a reinterpet_cast if not. So changing the includes will silently change the behavior.
Lesson learned: Never ever _ever_ ___ever___ use C-style casts in C++ code. Ever.

mtijanic:
Switched to `static_cast` since it makes it more obvious. The issue was, as noted previously, that C-style casts will do reinterpret_cast if static_cast can't be done because full definition is not available.

PR #236, opened 2018-11-12
Flurry of blows

BhaalM:
Added Flurry of blows for monk weapons defined with NWNX_Weapon_SetWeaponIsMonkWeapon function.

### DIFF
--- a/Plugins/Weapon/Weapon.cpp
+++ b/Plugins/Weapon/Weapon.cpp
@@ -834,8 +834,7 @@ void Weapon::SetCombatMode(CNWSCreature *pCreature, unsigned char nMode, int32_t
     {
         plugin.m_SetCombatModeHook->CallOriginal<int32_t>(pCreature, nMode, bForceMode);
     }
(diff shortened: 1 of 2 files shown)


Plugins/Weapon/Weapon.cpp
```diff
@@ -809,6 +815,52 @@ int32_t Weapon::GetRangedAttackBonus(NWNXLib::API::CNWSCreatureStats* pStats, bo
...
+    
+    return plugin.m_ToggleModeHook->CallOriginal<int32_t>(pCreature, nMode);
+}
+
+void Weapon::SetCombatMode(CNWSCreature *pCreature, unsigned char nMode, int32_t bForceMode)
+{
+    Weapon& plugin = *g_plugin;
+    if(pCreature==nullptr)
+    {
+        plugin.m_SetCombatModeHook->CallOriginal<int32_t>(pCreature, nMode, bForceMode);
```
Liareth: shouldn't this be `CallOriginal<void>` and then a `return;` after it?

Issue #201, opened 2018-06-12
Feature Request: Depleting Charges Doesn't Destroy Item


1d3s:
Using up all the charges on an item destroys it. Would anyone be interested in adding a tweak that prevents this behavior? There are workarounds, such as adding unlimited casts of Light, but they're ugly.

http://nwn.wikia.com/wiki/Charge

mtijanic:
I think we can close this now.

PR #237, opened 2018-11-18
tweaks: Add PreserveDepletedItems tweak

mtijanic:
As requested in #201 , running out of charges no longer destroys the item if tweak is activated.

Had to do this trick as AIActionItemCastSpell() is an absolute monster of a function and there's no way it would get done properly.

Only did a basic test that it works in the expected case and that it didn't break anything. Counting on @1d3s and whoever else wants to use this to give it a more thorough testing and report bugs.

### DIFF
--- a/Plugins/Tweaks/Tweaks/PreserveDepletedItems.cpp
+++ b/Plugins/Tweaks/Tweaks/PreserveDepletedItems.cpp
@@ -1,0 +1,44 @@
+#include "Tweaks/PreserveDepletedItems.hpp"
+#include "API/CNWSCreature.hpp"
+#include "API/CNWSItem.hpp"
+#include "API/CNWSObjectActionNode.hpp"
+#include "API/Functions.hpp"
+#include "API/Globals.hpp"
+#include "API/Version.hpp"
+
+#include "Services/Hooks/Hooks.hpp"
+#include "Utils.hpp"
+
+
+namespace Tweaks {
+
+using namespace NWNXLib;
+using namespace NWNXLib::API;
+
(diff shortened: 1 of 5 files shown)


Plugins/Tweaks/Tweaks/PreserveDepletedItems.cpp
```diff
@@ -0,0 +1,44 @@
...
+    pAIActionItemCastSpell_hook = hooker->FindHookByAddress(Functions::CNWSCreature__AIActionItemCastSpell);
+}
+
+
+uint32_t PreserveDepletedItems::CNWSCreature__AIActionItemCastSpell_hook(CNWSCreature *pThis, CNWSObjectActionNode *pNode)
+{
+    // If at risk of destroying the item, inflate charge count temporarily to bypass the destroy
+    // event, then set it back to mark the spells as unusable.
+    auto *pItem = Utils::AsNWSItem(Utils::GetGameObject((Types::ObjectID)pNode->m_pParameter[0]));
+    if (pItem && pItem->m_nNumCharges <= 5)
```
Liareth: Why 5? What if the item consumes more than 5 charges on use? (for example, an item with multiple spell levels on it?) Is 5 the most charges you can consume at once?

mtijanic: Yes, there's no in-game property to make it cost more than 5 charges per use. This is hardcoded, so simple CC can't bypass.

1d3s:
Awesome, thanks @mtijanic!

I'm away from my nwn-capable machine for about a week, but I'll definitely be adding this to my PW when I can and I'll update you on how it goes.

Issue #234, opened 2018-10-22
Reset Server after ThreadWatchdog detected a LongStall


milliorn:
https://github.com/nwnxee/unified/blob/master/Plugins/ThreadWatchdog/ThreadWatchdog.cpp

Can we do a NWNX_Administration_ShutdownServer(); once this is detected?  Is there a reason why this is not done?  Is there something else in Metrics that does the restart or something that keeps the server going?  Currently it seems what happens is that the server hangs and you must restart it manually.

Liareth:
The purpose of the ThreadWatchdog is to submit a metric when it indicates a potential stall on the main thread. Be design it isn't responsible for restarting the server, and it can't even be assumed that a 15 second stall is an unrecoverable lock either - it could just be a long IO operation. A configurable option could be added (defaulting to off) which just kills the server when it detects one of these stalls.

mtijanic:
The metrics system isn't the easiest to set up, unfortunately, so I don't fault people for not using it. It'd be nice if either the metrics could be sent to a simple file, or if the Watchdog had a config option that gives it a file to write to when it detects a hang.

Then you can have a separate process wait on that file and kill nwserver if needed.

Urothis:
Killing the server would be the preferred behavior docker wise. 

If we could expose that into an env setting
NWNX_WATCHDOG_SHUTDOWN=

milliorn:
> A configurable option could be added (defaulting to off) which just kills the server when it detects one of these stalls. 

> 

Can we do this so we can address the hang if we don't have metrics setup or not in use?  Be greatly appreciated and keep things automated.

mtijanic:
I think we can close this now.

PR #238, opened 2018-11-18
Watchdog autokill

mtijanic:
As discussed in #234 , config option to kill the server if it's stalled for a certain time.

* `NWNX_THREADWATCHDOG_PERIOD`: Set the period at which the watchdog fires, in seconds
* `NWNX_THREADWATCHDOG_KILL_THRESHOLD`: Number of successive long stall detections needed to kill the server

The kill isn't quite friendly, but it gets the job done.

### DIFF
--- a/Plugins/ThreadWatchdog/ThreadWatchdog.cpp
+++ b/Plugins/ThreadWatchdog/ThreadWatchdog.cpp
@@ -34,19 +34,17 @@ namespace ThreadWatchdog {
 static bool g_exit;
 static uint64_t s_mainThreadCounter = 0;
 static uint64_t s_watchdogLastObservedCounter = 0;
-static uint32_t s_watchdogPeriod = 15;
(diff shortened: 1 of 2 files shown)


Plugins/ThreadWatchdog/ThreadWatchdog.cpp
```diff
@@ -85,10 +94,19 @@ void ThreadWatchdog::MainLoopUpdate(Services::Hooks::CallType, API::CServerExoAp
                     // tasks, and so we remain productive until somebody can come along and figure out why we've
                     // hit a long stall.
                     tasks->ProcessWorkOnMainThread();
```
Liareth: Not related to your change, but this is actually pretty dangerous, and could turn a 15 second stall due to SQL into a crash pretty easily. There are major issues with not doing this (for one, the metric never goes through), so maybe we should consider raising the default interval to 60 seconds. If you've stalled for 60 seconds you may as well have crashed anyway.

Issue #192, opened 2018-06-03
Tracking Plugin: DM Connection Scenario


zeroark:
On file: unified/Plugins/Tracking/Targets/Activity.cpp

`const bool isDm = false; //TODO: reinterpret_cast<uintptr_t>(player->m_vtable[3]) == Functions::CNWSDungeonMaster__AsNWSDungeonMaster; NWNX_EXPECT_VERSION(8109);`

It looks like the DM flag for the tracking plugin is still pending to be done. This is no priority, but I'm just leaving this here so if anybody comes up with a solution or knows how to solve it, they can pick it up!

Liareth:
This needs API generation changes to make it work - the vtable order needs to be embedded in the class. Or, it would require manual inspection of the vtable every API update,.

mtijanic:
I think we can close this now.

PR #239, opened 2018-11-18
tracking: fix player iteration and player/dm split

mtijanic:
See #192 .  No need for vtable faffery, exoapp already exposes a way to cycle through all clients and get both their client OIDs and their nwsobject OIDs. Can detect DM-ness from CNWSCreatureStats.

GetFirstPC/GetNextPC is safe to do as this is in the main loop update and won't trigger in the middle of someone else's iteration (e.g. while a script is running).

Verified it shows both PC and DMs properly. Can't really imagine a case where the nwsobject is null, but it'll show as `(unknown)` then if it ever happens.


Liareth:
Will this show a DM being online even when they are possessing an NPC? Merging it anyway because it's 100% better than before even if that case breaks.

Issue #183, opened 2018-05-23
Feature request: open object inventory GUI


weavejester:
Is it possible for a function to open an creature or placeable inventory, bypassing the usual `ActionInteractObject`, preferably without regard to distance?

I'm thinking of a function like:

```
void NWNX_Player_ForceOpenInventoryGUI(object player, object container);
```

I notice there's a function `CNWSPlayerInventoryGUI` that suggests this might be possible, and a function `SendServerToPlayerGuiContainerObject_Status` that maybe could be used? I don't yet have a good enough understanding of the code to figure out if this is possible. If someone could point me in the right direction I could have a bash at this, or else confirm that this isn't currently possible.

My reason for wanting this is to create treasure chests that generate different loot for different players. Each player that opens the chest will get a different inventory with loot that only they can see. This lessens the need to divvy up loot at the end of the dungeon.

I've played around with creating invisible objects by the chest that contain the instanced loot, and then using `ActionInteractObject` to force the player to open them, but I frequently ran into pathing problems, as the PCs needed a clear path to the invisible inventory objects. Being able to force the inventory GUI up on screen would solve the issue.

mtijanic:
Done, can be closed.

PR #245, opened 2018-11-22
player: add ForcePlaceableInventoryWindow()

mtijanic:
As requested in #183 .

- Only works on placeables. 
    - For creatures, I couldn't get PCs to properly display random NPC inventories. And if you want to look at your henchmen, or if you're a DM, there's already the builtin `OpenInventory` that does it. Couldn't really add much value to that.
    - For items (boxes) I was getting weird behavior at times, so just ignoring it for now.
- Placeable's open/close animation will be played
- If the placeable is in a different area than the player, the portrait may not be shown
- Clicking the 'close' button will cause the player to walk to the placeable; or wander off to the edge of the area if placeable is in the different area
- Walking anywhere automatically closes the inventory

But generally works well, especially for usecase listed in #183.


weavejester:
Oh excellent! Thank you very much.

Issue #46, opened 2018-02-10
Need a way to selectively disable building some plugins


mtijanic:
Right now, we build all or nothing, with cmake being able to skip the few plugins that have external dependencies (e.g. JVM, SQL), if those dependencies are not found.

It'd be nice if there was a way to blacklist (and/or whitelist) plugins to be built using the environment or `make` arguments, regardless of external libraries present.
Most noticeable is probably the i386 version of ruby on amd64 systems, which if present causes the ruby plugin to build, but cannot actually compile it, resulting in a build failure.

PR #246, opened 2018-11-22
Add buildtime plugin filtering and out-of-tree plugin support

mtijanic:
Allow some more control for the build by setting a few environment variables:

```
export NWNX_SKIP_PLUGINS="WebHook;Chat;Weapon;SQL;Mono"
```

Will cause those plugins not to be built. Additionally, you could do:
```
export NWNX_ADDITIONAL_PLUGINS="/path/to/private/Plugin;/path/to/other/Plugin2"
```

which will pick up plugins that live outside of the `unified/Plugins` directory. This is useful for plugins that might not be wanted by most people, or that don't play well with others. Or just for organization, so you keep server-specific plugins in the server code, and keep the nwnx folder clean.

Note that this is the first time I've ever worked with cmake, so if this is horrible for some reason, please let me know.

This implements #46

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -97,8 +97,19 @@ add_subdirectory(Core)
 # Detect every plugin and store it in plugins . . .
 file(GLOB plugins Plugins/*/CMakeLists.txt)
 
+# Allow skipping certain plugins by putting their names in env. variable
+foreach(skipped $ENV{NWNX_SKIP_PLUGINS})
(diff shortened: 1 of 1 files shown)


mtijanic:
Note: these variables must be set before running `cmake`, not `make`. They are used to generate the makefiles, but the makefiles themselves don't understand them. If you change the env variables, you must rerun cmake.

PR #252, opened 2018-12-02
New service: Commands

mtijanic:
As mentioned in #6 , this allows plugins (and core) to register console commands on the server. It works by overriding `fgets` when the server tries to read stdin, and checks NWNX commands first. Some notes:

- The input is handled on a side thread, so commands must be scheduled into a queue and then run on the main thread.
- No (useful) commands currently. We can try doing config changes and plugin load/unload in a follow up change
- nwnx commands can shadow nwserver ones - If a command is registered for nwnx, a command of the same name can't be run on nwserver. This might be an issue if we want to use load/unload/reload for plugins, since `load` is already a nwserver command to load a savegame. Might need to prefix them explicitly
- Plugin commands are in the global namespace, so plugins can conflict over the command names. It will print a warning in this case, and only use the first one.
- Original implementation done on a phone via ssh. Not recommended for replication.

### DIFF
--- a/NWNXLib/Services/Commands/Commands.cpp
+++ b/NWNXLib/Services/Commands/Commands.cpp
@@ -1,14 +1,88 @@
 #include "Services/Commands/Commands.hpp"
+#include <stdio.h>
+#include <dlfcn.h>
 
 namespace NWNXLib {
 namespace Services {
 
+static Commands *g_commands;
+
 Commands::Commands()
 {
+    g_commands = this;
+    RegisterCommand("nwnx_commands_test", [](std::string& s){ LOG_NOTICE("Selftest command. Args: '%s'", s.c_str()); });
 }
 
 Commands::~Commands()
 {
+    g_commands = nullptr;
+
+}
+
+bool Commands::RegisterCommand(const std::string& cmd, CommandFunc func)
+{
+    auto it = m_commandMap.find(cmd);
+    if (it == m_commandMap.end())
+    {
+        m_commandMap[cmd] = func;
+        LOG_INFO("Registering command '%s'", cmd.c_str());
+        return true;
+    }
+    else
+    {
+        LOG_WARNING("Command '%s' already registered", cmd.c_str());
+        return false;
+    }
+}
+
+void Commands::UnregisterCommand(const std::string& cmd)
+{
+    LOG_INFO("Unregistering command '%s'", cmd.c_str());
+    m_commandMap.erase(cmd);
+}
+
+bool Commands::ScheduleCommand(const std::string& cmdline)
+{
+    std::string cmd, args;
+    size_t space = cmdline.find(' ');
+
+    if (space == std::string::npos) // No args
+    {
+        cmd = cmdline.substr(0, cmdline.length() - 1);
+        args = "";
+    }
+    else
+    {
+        cmd = cmdline.substr(0, space);
+        args = cmdline.substr(space+1, cmdline.length()-space-2);
+    }
+
+    auto it = m_commandMap.find(cmd);
+    if (it != m_commandMap.end())
+    {
+        std::lock_guard<std::mutex> guard(m_queueLock);
+        m_commandQueue.push_back(std::make_pair(cmd, args));
+        LOG_DEBUG("Scheduled command '%s' with args '%s'", cmd.c_str(), args.c_str());
+        return true;
+    }
+
+    return false;
+}
+
+void Commands::RunScheduledCommands()
+{
+    std::lock_guard<std::mutex> guard(m_queueLock);
+    for (auto item: m_commandQueue)
+    {
+        LOG_DEBUG("Running command '%s', args: '%s'", item.first.c_str(), item.second.c_str());
+        auto it = m_commandMap.find(item.first);
+        if (it != m_commandMap.end())
+        {
+            auto func = it->second;
+            func(item.second);
+        }
+    }
+    m_commandQueue.clear();
 }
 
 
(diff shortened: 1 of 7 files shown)


NWNXLib/Services/Commands/Commands.cpp
```diff
@@ -0,0 +1,128 @@
...
+    else
+    {
+        LOG_WARNING("Command '%s' already registered", cmd.c_str());
+        return false;
+    }
+}
+
+void Commands::UnregisterCommand(const std::string& cmd)
+{
+    LOG_INFO("Unregistering command '%s'", cmd.c_str());
```
Liareth: I guess there is a minor issue here - plugins can unregister each other's commands. It's minor and I don't care much about it. The reason the proxy exists is to avoid this - each plugin has its own instance of the proxy which stores instance-specific info (e.g. which commands it has already registered) to ensure it can only touch what it owns. Not saying you need to change it for me to merge this PR, just stating it here for consideration.

NWNXLib/Services/Commands/Commands.cpp
```diff
@@ -0,0 +1,128 @@
...
+    auto it = m_commandMap.find(cmd);
+    if (it == m_commandMap.end())
+    {
+        m_commandMap[cmd] = func;
+        LOG_INFO("Registering command '%s'", cmd.c_str());
+        return true;
+    }
+    else
+    {
+        LOG_WARNING("Command '%s' already registered", cmd.c_str());
```
Liareth: Perhaps this should be an exception, not a warning. That will stop NWNX from starting. This is how it works for the other services - so if there is a conflict between two plugins, the world stops until the users fixes their configuration. With that said, chat commands aren't exactly critical so a conflicting registration isn't the biggest deal. Hmm.

NWNXLib/Services/Commands/Commands.cpp
```diff
@@ -0,0 +1,128 @@
...
+
+    if (space == std::string::npos) // No args
+    {
+        cmd = cmdline.substr(0, cmdline.length() - 1);
+        args = "";
+    }
+    else
+    {
+        cmd = cmdline.substr(0, space);
+        args = cmdline.substr(space+1, cmdline.length()-space-2);
```
Liareth: If the user enters `test `, will this crash?

NWNXLib/Services/Commands/Commands.cpp
```diff
@@ -0,0 +1,128 @@
...
+    }
+
+    auto it = m_commandMap.find(cmd);
+    if (it != m_commandMap.end())
+    {
+        std::lock_guard<std::mutex> guard(m_queueLock);
+        m_commandQueue.push_back(std::make_pair(cmd, args));
```
Liareth: Minor: You could `std::move` the strings in (and do the log above the command queue push) to avoid two string cpies. Maybe the optimiser will do this for you. You make be able to use `emplace_back` and remove the `std::make_pair` to avoid a third allocation but I don't know if that will work for sure.

NWNXLib/Services/Commands/Commands.cpp
```diff
@@ -0,0 +1,128 @@
...
+}
+
+bool Commands::ScheduleCommand(const std::string& cmdline)
+{
+    std::string cmd, args;
+    size_t space = cmdline.find(' ');
+
+    if (space == std::string::npos) // No args
+    {
+        cmd = cmdline.substr(0, cmdline.length() - 1);
```
Liareth: Probably a dumb question, but why do you drop the last character from the command if there's no space?

NWNXLib/Services/Commands/Commands.cpp
```diff
@@ -0,0 +1,128 @@
...
+}
+}
+
+
+extern "C" char *fgets(char * str, int num, FILE *stream)
+{
+    using Type = char*(*)(char*,int,FILE*);
+    static Type real;
+    if (!real)
+        real = (Type)dlsym(RTLD_NEXT, "fgets");
```
Liareth: This won't work on Windows. But at this point I don't think we care. If you did care, you could call the thing in `Platform::`. Probably needs work to make it function with `RTLD_NEXT` tho.

NWNXLib/Services/Commands/Commands.cpp
mtijanic: Nope, it was one of the things I tested.

NWNXLib/Services/Commands/Commands.cpp
mtijanic: Yeah, I didn't see it as end-of-the-world kind of thing. Maybe ERROR would be better, possibly an assert (breaks if you have a debugger), but most users won't be using these commands anyway..

NWNXLib/Services/Commands/Commands.cpp
mtijanic: You're right. I'll fix that before we should merge this. There's also the issue of plugin hotplug/unload - we want to unregister all of its commands, so that next time it loads it'll be available. Hold the PR for now.

NWNXLib/Services/Commands/Commands.cpp
mtijanic: I'll do this in the cleanup pass. Thanks.

NWNXLib/Services/Commands/Commands.cpp
mtijanic: Because it returns the darned `\n` as well..

NWNXLib/Services/Commands/Commands.cpp
mtijanic: Yeah, the whole thing won't work on Windows, for many reasons. One being that nwserver.exe does not even have a console input on Windows. If we ever get Windows support, either the entire service needs to be stubbed out, or we need to spawn a new console window.

mtijanic:
Updated:
- Commands are now owned by a plugin; plugins can only unregister own commands and will unregister all commands when they are unloaded
- Bit more aggressive trimming of the command line. Don't care about perf there as this runs once in a never, on a side thread anyway.

The trimming works nicely in all the tests I could think to do.

NWNXLib/Services/Commands/Commands.cpp
```diff
@@ -0,0 +1,144 @@
...
+
+    cmd = cmdline.substr(0, space);
+    args = (space == std::string::npos) ? "" : cmdline.substr(space, std::string::npos);
+
+    trim(cmd); trim(args);
+
+    auto it = m_commandMap.find(cmd);
+    if (it != m_commandMap.end())
+    {
+        std::lock_guard<std::mutex> guard(m_queueLock);
```
Liareth: This mutex needs to cover the `m_commandMap.find(cmd)`, else there's a race condition where you may try to search the vector while the other thread is in the process of clearing it.

NWNXLib/Services/Commands/Commands.cpp
```diff
@@ -0,0 +1,144 @@
...
+        return true;
+    }
+
+    return false;
+}
+
+void Commands::RunScheduledCommands()
+{
+    std::lock_guard<std::mutex> guard(m_queueLock);
+    for (auto item: m_commandQueue)
```
Liareth: You could `std::swap` `m_commandQueue` with a vector on the stack and then release the lock early to prevent the scheduling thread from blocking if the command is expensive. It probably doesn't matter much, I just thought I'd mention it because I mentioned the other thing too.

NWNXLib/Services/Commands/Commands.cpp
mtijanic: Good spot, resolved.

NWNXLib/Services/Commands/Commands.cpp
mtijanic: Eh, I don't really want to. I don't care if we block the input thread for however long, it isn't doing anything useful anyway - can't collect new input while the current command is being processed. And adding the swap() will cause extra work on the main thread, which is on the critical path.

Liareth: Oh, I had it the wrong way around. For some reason I thought we were blocking the main thread.

Issue #251, opened 2018-11-30
Redis list cursor


Urothis:
"all the code to actually handle list cursors appears to be absent." @niv 

This affects multiple commands:
SCAN
SSCAN
HSCAN
ZSCAN

Probably more that I have not tested yet.

Urothis:
Closing as fixed by #253

PR #253, opened 2018-12-02
Redis: Fixes and array support

niv:
This is ready for testing by interested parties, especially those blocking on #251.

It's not done yet as far as redis feature completion goes. Needs more polish and there's some open outstanding bugs.

This PR also fixes some argument reordering nonsense, and regenerates redis includes for 5.x.

What doesn't work yet is any script command that has `multiple` arguments, and any argument that has a prefix/label to allow it being optional. For example, SCAN.


niv:
I've pushed another update that reworks the nwscript api to be typesafe and support nested results (lists-of-lists, etc).

This will still need some testing and a cleanup before merge.

Example usage:

```c
#include "nwnx_redis"

void Log(string msg)
{
	NWNX_Redis_PUBLISH("log", msg);
}

void PrintResult(int result, string prefix)
{
	int resultType = NWNX_Redis_GetResultType(result);
	string pfx = prefix + "/[" + IntToString(result) + ", type: " + IntToString(resultType) + "]";

	switch (resultType)
	{
		case NWNX_REDIS_RESULT_ARRAY:
			{
				int i; for (i = 0; i < NWNX_Redis_GetArrayLength(result); i++)
				{
					PrintResult(NWNX_Redis_GetArrayElement(result, i), pfx + "{" + IntToString(i) + "}");
				}
			}
			break;
		default:
			Log(pfx + " = " + NWNX_Redis_GetResultAsString(result));
			break;
	}
}

void main()
{
	int i;

	NWNX_Redis_PUBLISH("log", "============");

	NWNX_Redis_DEL("test");
	for (i = 0; i < 10; i++)
	{
		NWNX_Redis_RPUSH("test", IntToString(i));
	}

	Log("Testing LRANGE");
	int lrange = NWNX_Redis_LRANGE("test", 0, 1);
	PrintResult(lrange, "lrange");

	Log("Testing SCAN");
	int scan = NWNX_Redis_SCAN(0);
	PrintResult(scan, "scan");
}```

**Comment by Urothis** [comment]
Current issue with testing

```nwn-logic_1  | I [20:46:29] [NWNX_Core] [NWNXCore.cpp:255] Skipping plugin NWNX_Player due to configuration.
nwn-logic_1  | I [20:46:29] [NWNX_Core] [NWNXCore.cpp:264] Loaded plugin 4 (Profiler) v1 by Liareth.
nwn-logic_1  | E [20:46:29] [NWNX_Core] [NWNXCore.cpp:269] Failed to load plugin (NWNX_Redis.so) because 'An exclusive hook has already been applied in this memory location.'.
nwn-logic_1  | F [20:46:29] [NWNX_Core] [NWNXCore.cpp:403] The server encountered a fatal error 'An exclusive hook has already been applied in this memory location.' during setup and must now terminate.
nwn-logic_1  | We just crashed hard. :( Sorry about that. But you can help us fix this! Please report this crash at http://redmine.beamdog.com, with the following information attached:
nwn-logic_1  | 
```

Liareth:
This conflicts with NWNX_Profiler (unless you define `NWNX_PROFILER_ENABLE_SCRIPTS=false`) and NWNX_Mono. That conflict is fine I think, you just can't run them together. Mono also conflicts with Profiler without the above define.

You may be able to avoid the conflict by hooking somewhere else.

niv:
Boo. Using `ClearStack` then. Semantically the better one to grab on anyways.

Liareth:
@niv @Urothis Is this good to go yet?

Urothis:
I've yet to run into any issues. 

The only thing we might want to do is alert everyone in the discord to let them know that all code involving redis will need to be updated.

niv:
I'm ok with this being merged in. There's still stuff missing but I won't get to that this year probably.

Liareth:
Gonna merge this, since people who use Redis have tried it and told me it works.

Issue #247, opened 2018-11-22
Feature Request: Flag to Override/Disable Pick Pocket Action


talaefu:
Hi,

Would like to request the following if possible - the possibility to add some flag to disable or override the Pick Pocket action.

I can't recall how it was disabled for our server in 1.69 (it was also done way before my time there) but I assume it was with NWNX_Events back then. Our server is using a customized Pick Pocket system and tools, as the vanilla one doesn't allow much of control over to adjust/balance and keep track of.

Thanks!

Issue #34, opened 2018-01-30
Events Plugin - Trap Trigger Event


torradmin:
The events plugin is missing hooks for:
- ~~Skill Use~~
- ~~Polymorph~~
- ~~Unpolymorph~~
- ~~Action Mode Toggle (only has stealth)~~
- ~~Crit/Dev Crit (dev crit should also be cancellable - see other issue!)~~
- Trap Trigger
- ~~Trap Set~~

Mizzajl:
Hi, just wanted to say that I also would like to be able to detect skill use and replace the default behaviour with a custom script :)

Daztek:
One left!

Issue #16, opened 2018-01-18
Events: Skip executing event


Liareth:
We need to allow the Events plugin to skip events.

1. Switch all shared hooks to exclusive hooks.
2. Come up with the API. `void NWNX_Events_SkipEvent()` maybe. Possibly add a comment explaining event skipping is only valid on NWNX events and not user defined events, or do extra work to check if an event has been skipped between invoking each subscribing script.
3. If the API is called, set a flag.
4. Clear the flag on event completion. Figure out a clean way to do this, maybe Events::StartEvent() and Events::FinishEvent() between Events::SignalEvent().
5. If the flag is set, skip calling the original function.

Liareth:
@zunath @mtijanic This is the sort of interface I was thinking of:

https://github.com/Liareth/unified/commit/ff715cc2acd889f21aaf86f6426393d6f98ef66e

zunath:
@Liareth I like it - very similar to 1.69's BypassEvent so I'm on board with it.

What happens if you call SkipEvent in an "After" subscriber? Anything?

mtijanic:
Maybe also a `GetEventSkipped()` to call from the AFTER handler?

mtijanic:
The API is there and most of the vents were migrated by daz. Closing this.

PR #261, opened 2018-12-17
Events: Skippable Item/Feat/Party/Skill Events

Daztek:
- The OnUseItem and OnUseFeat events are now skippable.
- Exposed the following event data for OnUseItem:
ITEM_SUB_PROPERTY_INDEX int 
TARGET_POSITION_X float       
TARGET_POSITION_Y float       
TARGET_POSITION_Z float      
- Added skippable Party events
- Added skippable Skill events


Plugins/Events/Events/FeatEvents.cpp
```diff
@@ -49,7 +69,9 @@ void FeatEvents::UseFeatHook
     Events::PushEventData("TARGET_POSITION_Y", pvTarget ? std::to_string(pvTarget->y) : "0.0");
     Events::PushEventData("TARGET_POSITION_Z", pvTarget ? std::to_string(pvTarget->z) : "0.0");
 
-    Events::SignalEvent(before ? "NWNX_ON_USE_FEAT_BEFORE" : "NWNX_ON_USE_FEAT_AFTER", thisPtr->m_idSelf);
+    Events::SignalEvent("NWNX_ON_USE_FEAT_AFTER", thisPtr->m_idSelf);
```
mtijanic: Do we want to signal the AFTER event if it was bypassed?

Daztek: It's signaled in BhaalM's Healer's Kit event and Liareth's [example](https://github.com/Liareth/unified/blob/ff715cc2acd889f21aaf86f6426393d6f98ef66e/Plugins/Events/Events/AssociateEvents.cpp) code from #16, so probably? :D

Daztek:
Now with 100% more CombatMode events

Daztek:
- Removed CombatModeEvents since it's handled by NWNX_CombatModes now
- Party events are now skippable

Daztek:
- Added skippable Skill events
This implements #247 and partially implements #34

Plugins/Events/Events/SkillEvents.cpp
```diff
@@ -0,0 +1,63 @@
...
+    uint8_t subSkill,
+    API::Types::ObjectID target,
+    NWNXLib::API::Vector targetPosition,
+    API::Types::ObjectID area,
+    API::Types::ObjectID usedItem,
+    int32_t activePropertyIndex)
+{
+    int32_t retVal;
+
+    Events::PushEventData("SKILL_ID", std::to_string(skill));
```
mtijanic: I actually like the lambda approach as opposed to duplicated code. When you push all this stuff, you can wrap it as:
```C
auto Push = [&]() { /* all the push events here */ };
```
Then you just call it as `Push()` both before and after the event. Removes code duplication.

Can also make it look like:
```C
auto PushAndSignal = [&](std::string &ev) -> bool {
    Events::PushEventData("...", ...);
    ...
    return EventsSignal(ev, thisPtr->m_idSelf);
```

And then your core function is:

```C
if (PushAndSignal("NWNX_ON_USE_SKILL_BEFORE"))
{
    retVal = ...;
}

PushAndSignal("NWNX_ON_USE_SKILL_AFTER");
return retVal;
```

And same logic can be applied to anything that needs to push more than 1-2 pieces of event data.

Daztek: Added, that does make it look a lot more readable

mtijanic:
I think this is good to merge now, but let's wait for @Liareth to take a look as well.

Issue #15, opened 2018-01-18
Events: Toggle mode


Liareth:
The StickyCombatModes plugin used to use the messaging bus to add two events - `NWNX_ON_MODE_ON` and `NWNX_ON_MODE_OFF`.

The code looked like this:

```cpp
GetServices()->m_hooks->RequestExclusiveHook<API::Functions::CNWSCreature__SetCombatMode>(&SetCombatModeHook);
```

... later ...

```cpp
void StickyCombatModes::SetCombatModeHook(API::CNWSCreature* thisPtr, uint8_t newMode, bool forceNewMode)
{
    if (thisPtr->m_bPlayerCharacter)
    {
        const uint8_t currentMode = thisPtr->m_nCombatMode;

        if (g_plugin->m_dispatchMessages && currentMode != newMode)
        {
            // ToggleMode gets called with modes which are +1 to what the script constants expect.
            // It then calls SetCombatMode with the 'proper' modes.
            // To simplify things, we just correct the mode we dispatch here, so the game events
            // receive what they expect to.
            auto messaging = g_plugin->GetServices()->m_messaging.get();

            if (currentMode != 0)
            {
                messaging->BroadcastMessage("NWNX_EVENT_PUSH_EVENT_DATA", { "MODE", std::to_string(currentMode + 1) } );
                messaging->BroadcastMessage("NWNX_EVENT_SIGNAL_EVENT", { "NWNX_ON_MODE_OFF", ObjectIDToString(thisPtr->m_idSelf) });
            }

            if (newMode != 0)
            {
                messaging->BroadcastMessage("NWNX_EVENT_PUSH_EVENT_DATA", { "MODE", std::to_string(newMode + 1) } );
                messaging->BroadcastMessage("NWNX_EVENT_SIGNAL_EVENT", { "NWNX_ON_MODE_ON", ObjectIDToString(thisPtr->m_idSelf) });
            }
        }
    }

    return g_setCombatModeHook->CallOriginal<void>(thisPtr, newMode, forceNewMode);
}
```

Now that we no longer have the StickyCombatModes plugin, we need to move this event to the Events plugin with the same effective outputs as above.

mtijanic:
Done.

PR #263, opened 2018-12-17
Plugin: CombatModes

Daztek:
Basically Liareth's stuff from #15 in its own plugin except the following since I'm not sure it would work as intended.

>// ToggleMode gets called with modes which are +1 to what the script constants expect.
  // It then calls SetCombatMode with the 'proper' modes.
  // To simplify things, we just correct the mode we dispatch here, so the game events
  // receive what they expect to.

These are the constants from nwscript:

>// Combat Mode Defines
int COMBAT_MODE_INVALID                 = 0;// 0 
int COMBAT_MODE_PARRY                   = 1;// mode correctly returns 1
int COMBAT_MODE_POWER_ATTACK            = 2;// modet correctly returns 2
int COMBAT_MODE_IMPROVED_POWER_ATTACK   = 3;//  mode correctly returns 3
 // counterspell aka 4 doesn't have a constant for some reason, mode correctly returns 4
int COMBAT_MODE_FLURRY_OF_BLOWS         = 4;// mode returns 5
int COMBAT_MODE_RAPID_SHOT              = 5;// mode returns 6
int COMBAT_MODE_EXPERTISE               = 6;// mode returns 7
int COMBAT_MODE_IMPROVED_EXPERTISE      = 7;// mode returns 8
int COMBAT_MODE_DEFENSIVE_CASTING       = 8;//mode returns 9
int COMBAT_MODE_DIRTY_FIGHTING          = 9; // mode returns 10
int COMBAT_MODE_DEFENSIVE_STANCE        = 10;// mode returns 11

Sherincall gave me the following list which is correct

>NONE                    0
PARRY                   1
POWER_ATTACK            2
IMPROVED_POWER_ATTACK   3
COUNTERSPELL            4
FLURRY_OF_BLOWS         5
RAPID_SHOT              6
EXPERTISE               7
IMPROVED_EXPERTISE      8
DEFENSIVE_CASTING       9
DIRTY_FIGHTING          10
DEFENSIVE_STANCE        11


mtijanic:
Step 1: Need to have the Events plugin send the data back. Add these two to `Events::SignalEvent`, after the loop:
```C
    GetServices()->m_messaging->BroadcastMessage("NWNX_EVENT_SIGNAL_EVENT_RESULT",  { eventName, *result});
    GetServices()->m_messaging->BroadcastMessage("NWNX_EVENT_SIGNAL_EVENT_SKIPPED", { eventName, std::to_string(skipped?1:0)});
```

Step 2: CombatModes must subscribe to these messages.
2a) Add `bool m_skipped;` to CombatModes.h
2b) Add this to the plugin constructor:
```C
    GetServices()->m_messaging->SubscribeMessage("NWNX_EVENT_SIGNAL_EVENT_SKIPPED",
        [this](const std::vector<std::string> message)
        {
            if (message[0] == "NWNX_ON_MODE_ON" || message[0] == "NWNX_ON_MODE_OFF")
                this->m_skipped = std::strtoul(message[1].c_str()) == 1;
        });
```

Step 3: Check `m_skipped` in `SetCombatMode` to decide whether to call the original or not.

I haven't actually compiled this code, so there might be some quirks to iron out, but it should be right in a nutshell. Ping on discord if there are issues.

Daztek:
Added skipping functionality with thanks to Sherincall

Liareth:
So I think I misunderstood something ... when we were discussing moving the events to a new plugin, I thought the plugin would *do* something. In other words, I thought you wanted to do something with SetCombatMode and since it was already used elsewhere, you were going to extract it, plus your functionality, to a second plugin. I'm hesitant to have a plugin here that does nothing but send messages to Events - in that case, why not just put it in Events? What are your thoughts?

mtijanic:
Yeah, the stuff from the Weapon plugin should go here.

Liareth:
Is this waiting for more work to add the stuff from the Weapon plugin before merging? Or is the idea to merge it now and move the stuff later?

mtijanic:
I think it's okay to merge now. Weapon and CombatModes will be incompatible, but since this is a new plugin, that's not a big deal. And once it's in, we'll have an easier time moving the Weapon stuff in here.

PR #265, opened 2018-12-22
WIP: Commands to get and set environment variables

mtijanic:
Incremental update to the commands per #6 , allows querying and setting local variables. To be used with hot reload commands that are WIP.

### DIFF
--- a/Core/NWNXCore.cpp
+++ b/Core/NWNXCore.cpp
@@ -285,14 +285,22 @@ void NWNXCore::InitialSetupCommands()
             std::string varname = s.substr(0, equals);
             std::string value = s.substr(equals+1, std::string::npos);
             if (setenv(varname.c_str(), value.c_str(), 1) == 0)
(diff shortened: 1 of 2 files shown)


Core/NWNXCore.cpp
```diff
@@ -275,6 +275,35 @@ void NWNXCore::InitialSetupPlugins()
...
+void NWNXCore::InitialSetupCommands()
+{
+    m_services->m_commands->RegisterCommand("set", [](std::string& s)
+    {
+        size_t equals = s.find('=');
+        if (equals != std::string::npos)
+        {
+            std::string varname = s.substr(0, equals);
+            std::string value = s.substr(equals+1, std::string::npos);
+            if (setenv(varname.c_str(), value.c_str(), 1) == 0)
```
Liareth: I think this is problematic. setenv is not thread safe. I think a safer design would be to extend the Config service so that you can change config settings that way.

mtijanic:
(Closing this for not as all of the code is gonna be thrown away anyway)

PR #270, opened 2018-12-29
Events: Fix the TARGET event data in DMActionEvents

Daztek:


Liareth:
Could you explain briefly why chopping off the last bit is necessary?

mtijanic:
> why chopping off the last bit is necessary?

Because NWN. It has a concept of 'Local' and 'Remote' objects, and the last bit means that this is a "remote" object ID. The message protocol says that you're supposed to send the object IDs as you see them, and are supposed to fix the bit on receipt. There are no "remote" objects on the server.

Now, that is absolutely not presented that way. NWN attempts to implements "server" and "client" objects, but does so poorly and inconsistently. The above is the only way to understand the system and preserve your sanity.

tl;dr: `ReadOBJECTIDServer()` does it as well.

PR #273, opened 2019-01-03
Flurry of Blows compatibility with CombatModes plugin

BhaalM:


Liareth:
tbh, this seems a bit weird to me. Why have it in the Weapons plugin at all? Shouldn't this exist purely in the CombatModes plugin?

BhaalM:
@Liareth if you look at my last commit I removed the hook from the weapon plugin. Now it is only in the CombatModes plugin.

Liareth:
But it uses the messaging system, and requires NWNX_Weapons to be enabled for it to turn on? What is the purpose of that? Wouldn't it be better to just delete it completely from NWNX_Weapons and add it to NWNX_CombatModes?

BhaalM:
Because the new monk weapons are defined in the Weapon plugin (not only for the flurry of blows but for the monk attack table). If you don't define new monk weapons you don't need to activate the flurry of blows thing.

PR #274, opened 2019-01-05
Tweaks: Hide Item Lost And Acquired Messages

Daztek:
Small tweak to hide the "Lost Item: "/ "Acquired Item: " messages.


Christian-Vassallo:
Would it be possible to hide also login/logout?

ELadner:
Also, would it be possible to hide acquired items only on login. Every time
a character logs in it looks like they acquire their entire inventory.   it
makes anti-cheating scripts difficult.

PR #275, opened 2019-01-06
Tweaks: Improved Hide Item Feedback Messages

Daztek:
Pretty big change, it now hides more item feedback messages with an optional environment var to only hide the lost/acquired messages.

Added a second mode where the player needs to have a local int set on them to hide the messages, pretty useful to only set when crafting or for other spammy things.

Let me now if I've committed any horrible coding crimes :D

Documentation:

## Hide Item Feedback Messages Notes

This tweak will hide the following messages from the combat log:

* Acquired Item: 
* Lost Item:
* Armor/Shield Penalty
* Weapon equipped as one handed weapon
* Weapon equipped as two handed weapon
* Equipped item swapped out

`NWNX_TWEAKS_HIDE_ITEM_FEEDBACK_MESSAGES` has two modes:

* `1` will always suppress the item feedback messages.
* `2` will only suppress the item feedback messages if the player has `SetLocalInt(oPlayer, "NWNX_HIDE_ITEM_FEEDBACK_MESSAGES", TRUE)` set on them.

If you only want to hide the lost/acquired messages you can set the following optional environment variable to true:

* `NWNX_TWEAKS_ONLY_HIDE_LOST_ACQUIRED_MESSAGES`: true or false


mtijanic:
NWNX stance so far was to avoid local variables as much as possible. Additionally, a "tweak" is supposed to be a tiny change in behavior that is configurable via a single simple variable.

I think this is shaping up to be a separate plugin. One where you can hide/show any message globally, and then also overwrite it per-PC. Something like:

```C
// If oPC is OBJECT_INVALID, applies to all PCs that don't have an explicit value set
void NWNX_Feedback_SetMessageHidden(int message, int bHidden, object oPC = OBJECT_INVALID);
int NWNX_Feedback_GetMessageHidden(int message, object oPC = OBJECT_INVALID);
```

For per-PC variables, take a look at e.g. AlwaysWalk in NWNX_Player.

Daztek:
Yeah, a separate plugin is probably for the best, I'll look into that sometime, closing this for now

PR #279, opened 2019-01-10
Plugin to facilitate renaming, overriding and customization of player names.

Silvard:
This plugin allows you to set a global override name and makes connected clients update it so the change is reflected consistently in, as far as I can tell, all UI elements except the player list, and the player's own portrait on hover. It hooks the player list message function to temporarily rename the player when it's sent. It allows you to set a prefix and a suffix to the name which will be appended to the name on UI elements, but the player list will only keep the name without affixes.


Liareth:
Could you please convert all the tabs to spaces in this?

Plugins/Rename/Rename.cpp
```diff
@@ -0,0 +1,219 @@
...
+            LOG_ERROR("No creature object found for Player ID %x, oidNWSObject %x",
+                pPlayer->m_oidPCObject, playerObjectID);
+			return stack;
+        }
+		
+		const auto sPrefix = Services::Events::ExtractArgument<std::string>(args);
+        const auto newName = Services::Events::ExtractArgument<std::string>(args);
+		const auto sSuffix = Services::Events::ExtractArgument<std::string>(args);
+		
+		const std::string fullDisplayName = sPrefix + newName + sSuffix + "</c>"; //put together the floaty name, escape all colors
```
Liareth: Does `</c>` work without a matching `<c>`?

Silvard: In my testing it does, and not escaping colors makes whatever color is applied to the name apply to the semi colons, but on second thought it might just be better to let the user handle it, since they can add it to the suffix should they want to.

Silvard:
> Could you please convert all the tabs to spaces in this?

No more tabs!

Christian-Vassallo:
This plugin is something like nwnx names?

mtijanic:
@Silvard please add the plugin to skip list at https://github.com/nwnxee/unified/blob/master/Dockerfile

Otherwise this looks reasonable to me. My personal preference would be to use the POS service, rather than keeping your own maps, and to iterate using ExoApp API rather than poking at the list manually (see https://github.com/nwnxee/unified/blob/master/Plugins/Tracking/Targets/Activity.cpp#L47-L49 ), but I don't feel too strongly about it.

Silvard:
@mtijanic Added it to the skip list. I'll look into switching over to POS and traversing through ExoApp as I add a bit more functionality into it, thanks for the heads up. 

@Christian-Vassallo It depends on your use case. NWNX_Names allowed you to map a name override to individual Player <-> Player pairs. This plugin only allows a global override which applies to every observer.

Plugins/Rename/Rename.cpp
```diff
@@ -0,0 +1,248 @@
...
+        
+        if (pCreature && !lsFirstName.empty() && ( bOriginal || !pCreature->m_sDisplayName.IsEmpty()))
+        {
+            pCreature->m_pStats->m_lsFirstName = ContainString(lsFirstName);
+            pCreature->m_pStats->m_lsLastName = ContainString(lsLastName);
+        }
+    }
+}
+
+std::string Rename::ExtractString(CExoLocString locStr)
```
mtijanic: This will create a copy of the parameter, with the additional memory allocations and everything. A better option would be to have the argument as `const CExoLocString& locStr` to pass it by (constant) reference. Same below.

Plugins/Rename/Rename.cpp
```diff
@@ -0,0 +1,248 @@
...
+        }
+        
+        const auto sPrefix = Services::Events::ExtractArgument<std::string>(args);
+        const auto newName = Services::Events::ExtractArgument<std::string>(args);
+        const auto sSuffix = Services::Events::ExtractArgument<std::string>(args);
+        
+        const std::string fullDisplayName = sPrefix + newName + sSuffix; //put together the floaty/chat/hover name
+        
+        pCreature->m_sDisplayName = fullDisplayName.c_str(); //sets the override floaty name, this goes away on logout/reset
+        pCreature->m_bUpdateDisplayName = true; //unsure if this is necessary
```
mtijanic: This will break with next patch, whenever that comes. Can just remove it then..

mtijanic:
I'm gonna go ahead and merge this now. If you wanna address the remaining feedback (reinterpret_cast, pass-by-reference arguments, m_bUpdateDisplayName) you can do in a follow up PR.

PR #280, opened 2019-01-12
SQL: Add SQLite Target

Daztek:
This seems to run `nwx_sql_t` without issues, so I guess it works, not sure if I'm doing everything right implementation wise.


Build/CMakeModules/FindSQLite3.cmake
```diff
@@ -0,0 +1,46 @@
+# - Find sqlite3
```
mtijanic: I wonder if it's not better to just add the sqlite source directly in the repo?

Plugins/SQL/Targets/SQLite.cpp
```diff
@@ -0,0 +1,218 @@
...
+SQLite::~SQLite()
+{
+    sqlite3_close(m_dbConn);
+}
+
+void SQLite::Connect(NWNXLib::ViewPtr<NWNXLib::Services::ConfigProxy> config)
+{ 
+    if (auto database = config->Get<std::string>("DATABASE"))
+    {
+        m_dbName = (*database).c_str();
```
mtijanic: FYI, `(*var).something` is equivalent to `var->something`. Often people consider the second easier to read.

Plugins/SQL/Targets/SQLite.cpp
```diff
@@ -0,0 +1,218 @@
...
+
+        LOG_INFO("SQLite Database name set to %s", m_dbName.c_str());
+    }
+    
+    // Save the database file to UserDirectory/database
+    static std::string dbPath = CombinePaths(
+        CombinePaths(std::string(Globals::ExoBase()->m_sUserDirectory.CStr()),  std::string("database")), m_dbName + ".sqlite3");
```
mtijanic: I like this!

mtijanic:
Shouldn't there be a change to `nwnx_sql_t.nss` to include the new DB type?

Daztek:
Updated nwnx_sql_t.nss :D

Plugins/SQL/Targets/SQLite.cpp
```diff
@@ -32,7 +32,7 @@ void SQLite::Connect(NWNXLib::ViewPtr<NWNXLib::Services::ConfigProxy> config)
 { 
     if (auto database = config->Get<std::string>("DATABASE"))
     {
-        m_dbName = (*database).c_str();
+        m_dbName = database->c_str();;
```
mtijanic: double `;` :P. Doesn't hurt any, though.

Daztek: Ugh, it hurts me tho

Build/CMakeModules/FindSQLite3.cmake
ELadner: That's common, but the sqlite parts won't get updated from upstream frequently (if at all).

PR #281, opened 2019-01-12
Core: Add NWNX_CORE_SKIP_ALL env var to skip all plugins on load

Daztek:
No more need to edit the dockerfiles whenever someone adds a new plugin, woo.


mtijanic:
Did you test this with docker?

Core/NWNXCore.cpp
```diff
@@ -253,7 +254,7 @@ void NWNXCore::InitialSetupPlugins()
 
         Plugin::CreateParams params = { services };
 
-        if (services->m_config->Get<bool>("SKIP", false))
+        if (services->m_config->Get<bool>("SKIP", !!skipAllPlugins))
```
mtijanic: No need for `!!` normalization, this is a `bool` which is guaranteed to fall within the possible values of a `bool`.

Daztek:
@Urothis told me it works with docker, so yay

Issue #269, opened 2018-12-27
Add GetCustomToken() function


MhlChesnokov:
possible to add to NWNX_Dialog the function "Get Selected Node Test ()" - which returns the value of the selected string in the dialog (i.e. the contents of <CUSTOM00000> in the string) ???
GetSelectedNodeText()
this would be very convenient for use in dialog systems when using " SetCustomToken()" . in 1.69 it was

mtijanic:
Is this not `GetCurrentNodeText()`? I'm not sure if it would return the actual token or `<CUSTOM123>`, but if the latter, then the request should be for `string GetCustomToken(int nTokenNum);` instead.

Daztek:
`GetCurrentNodeText()` returns `<CUSTOM****>`

MhlChesnokov:
possible to make the NWNX_WebHook plugin work not only with Latin encoding, to make it possible to use different encodings in the plugin.??

mtijanic:
(renaming to better reflect what is needed here)

mtijanic:
This should be fully implemented now, closing.

PR #283, opened 2019-01-13
Util: added GetCustomToken()

Daztek:
Implementation of #269

Adds GetCustomToken() to NWNX_Util.

### DIFF
--- a/Plugins/Util/Util.cpp
+++ b/Plugins/Util/Util.cpp
@@ -4,6 +4,8 @@
 #include "API/Globals.hpp"
 #include "API/CExoString.hpp"
 #include "API/CVirtualMachine.hpp"
+#include "API/CTlkTable.hpp"
+#include "API/CTlkTableTokenCustom.hpp"
 #include "Utils.hpp"
 #include "ViewPtr.hpp"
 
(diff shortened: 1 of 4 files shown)


Plugins/Util/Util.cpp
```diff
@@ -100,5 +104,30 @@ ArgumentStack Util::Hash(ArgumentStack&& args)
...
+    ArgumentStack stack;
+    std::string retVal = "";
+    auto *pTlk = API::Globals::TlkTable();
+    auto *pTokens = pTlk->m_pTokensCustom;
+    int numTokens = pTlk->m_nTokensCustom;
+
+    const auto tokenNumber = Services::Events::ExtractArgument<int32_t>(args);
+
+    CTlkTableTokenCustom token;
+    token.m_nNumber = tokenNumber < 0 ? 0 : tokenNumber;
```
mtijanic: There's no reason to set this. `token` here is just used to store the result.

mtijanic: If `tokenNumber` is negative, just return "" immediately.

Daztek: Done

Issue #121, opened 2018-04-05
Missing Get/Set ItemPropertySpellId


ELadner:
(from old nwnx_structs plugin)

```
/* Returns the spell id that created the item property. */
int GetItemPropertySpellId (itemproperty ipProp);

/* Sets the item property's spell id as specified, which will later be
 * returned with GetItemPropertySpellId(). */
void SetItemPropertySpellId (itemproperty ipProp, int nSpellId);
```

mtijanic:
See #31 - need an ability to push/pop itemproperty and effect types first.

PR #284, opened 2019-01-13
New core interface, allows passing engine structures to nwnx (nwnx_structs)

mtijanic:
Pretty massive core change:

- NWNX interface with nwscript moved from `NWSScriptVarTable` methods to `CNWVirtualMachineCommands`. 
  - This makes the nwscript side of the things much simpler
  - The core is reserves the right to take exclusive hooks on any other VM commands as well.
  - It is no longer possible to trigger NWNX functions from DM console.
  - `TagEffect` and `TagItemProperty` are used for passing structs. `PlaySound` is used to the CALL operation
  - **This breaks existing ABI** (yes, ABI _is_ the correct term here). This means you'll need to recompile your scripts with the new `nwnx.nss`. It gives a nice warning in the console.
- The `Events` service has been significantly chopped. The mandate of this service is (from now) to pass arguments/return values and trigger events. How these values are pushed across the VM boundary is not any of its business (anymore).
  - This will allow us to have alternative ABIs in the future, e.g. where NWNX calls are first class citizens with regular NWN function calls
- `effect` and `itemproperty` NWScript types are both `CGameEffect` and _must_ live on the heap. So they are always handled as `CGameEffect*`.
  - Due to some API weirdness, `delete (CGameEffect*)pEffect` has unpredictable results. Use `Utils::DestroyGameEffect(pEffect)` instead.
  - Since they are the same underlying type, the Util plugin is providing a function to cast between them in nwscript. Probably not much use here, but it helps cut down some code repetition.
- New NWNX_Effect plugin to operate on effects.
  - Currently only supports unpacking the effect into a pretty massive struct, and packing it back.
  - That (plus effect<->IP casting) should allow most of old nwnx_structs functions to be implemented in pure nwscript.

**TESTING**
I've tested this using all our unit tests (plus additional ones included here) and a few hand crafted tests. I will now migrate my dead PW to the new ABI and let it run for 24 hours at least to see if there's any stability issues.
If someone else has anything to suggest or offer regarding testing, it'd be very nice.


Core/NWNXCoreVM.cpp
```diff
@@ -0,0 +1,337 @@
...
+                oid = vartable->GetObject(varname);
+            }
+            success = vm->StackPushObject(oid);
+            break;
+        }
+        case VMCommand::GetLocalLocation:
+        {
+            CScriptLocation loc;
+            if (vartable)
+                vartable->GetLocation(varname);
```
Daztek: Missing `loc = ` here

mtijanic: Thanky!

mtijanic:
I think this is converging towards the final version. Just needs a bit more polish.

ELadner:
Does this cover #121 also?

mtijanic:
@Christian-Vassallo If I understand correctly, then not really. You only have the variables exposed in the struct. Some of them (e.g. strings 2 and 3) are unused and you can use those if needed.
There's ways to make it work, but they're a bit involved, or downright hacky. Can you give me a use case? Maybe there's a simple way to do what you need.

@ELadner Not _yet_. I'll add an itemproperty plugin as well. The core plumbing changes support it, but it needs a a few more functions exposed to nwscript.

Core/NWNXCoreVM.cpp
```diff
@@ -0,0 +1,473 @@
...
+    StackUnderflow               = -639,
+    InvalidOpCode                = -640,
+    InvalidExtraDataOnOpCode     = -641,
+    InvalidCommand               = -642,
+    FakeShortcutLogicalOperation = -643,
+    DivideByZero                 = -644,
+    FakeAbortScript              = -645,
+    IPOutOfCodeSegment           = -646,
+    CommandImplementerNotSet     = -647,
+    UnknownTypeOnRunTimeSTack    = -648
```
Daztek: Nitpicking, sorry :D

UnknownTypeOnRunTimeS**T**ack -> UnknownTypeOnRunTimeStack

mtijanic:
Removing the WIP tag and updating description. This is now enough to be the first proper version.

Christian-Vassallo:
Let’s consider that you want to apply a particular effect. For example during polymorphing.

Then you unpolymorph and you want that all the effect you applied are removed.
This would be useful for the special ability during poly, example fear aura, that after polymorphing is still there.

Another example: if you are in poly and you have a magic like: improved invisibility.
This effect works like the basic magic but you want to remove it if you unpoly.
So actually you determine if that effect is from some particular spell. In this case improved invisibility. But actually this could conflict with the normal spell improved invisibility, since the only you casted is from a special ability. This mean that sometime you have to create a new spell (improved invisibility polymorphing) but it is just a copy of another spell. 

If you can set some variable, so you can determine if that spell has been applied on a particular situation.
And if you want to remove it, you can check either the spell id of course but also some other variable.

However, I can live without this feature. I was just asking if it could be possible that’s all :)

mtijanic:
@Christian-Vassallo I mean, the `TagEffect()` function in NWN is perfect for this. You can give any effect an arbitrary string tag (e.g. "REMOVE_ON_UNPOLYMORPH"), and you can iterate through all effects looking at this with `GetEffectTag()`. Effect tags are a single string, but you can chain them as `"REMOVE_ON_POLYMORPH; DO_MORE_STUFF"` and then just do substring search with FindSubString().
Basically, you can do this already, no NWNX required.

Christian-Vassallo:
Perfect.

I didn’t see the possibility to have TagEffect. My bad

Core/NWNXCoreVM.cpp
```diff
@@ -0,0 +1,478 @@
...
+                pItemProperty = *res;
+            }
+        }
+        else ASSERT_FAIL_MSG("Only PUSH and POP operations allowed on TagItemProperty");
+    }
+    else
+    {
+        if (pItemProperty->m_nType == 91) // ItemProperty effect
+        {
+            // Why, Liareth, WHY?
```
Liareth: It's not my fault!

NWNXLib/Utils.cpp
```diff
@@ -271,5 +272,31 @@ bool CompareVariables(API::CNWSScriptVarTable *pVars1, API::CNWSScriptVarTable *
...
+            return &static_cast<API::CNWSObject*>(pObject)->m_ScriptVars;
+    }
+}
+
+void DestroyGameEffect(API::CGameEffect* pEffect)
+{
+    if (pEffect)
+    {
+        // Don't actually need the this pointer here.
+        static_cast<API::CNWVirtualMachineCommands*>(nullptr)->DestroyGameDefinedStructure(0, pEffect);
```
Liareth: Maybe it would be better to get the VM rather than casting nullptr? I think this is undefined behaviour, and even if it wasn't, it might cause issues later down the line when niv changes that function to spam cats or something.

Plugins/Effect/Effect.cpp
```diff
@@ -0,0 +1,171 @@
...
+    eff->m_sParamString[2] = Services::Events::ExtractArgument<std::string>(args).c_str();
+    eff->m_sParamString[1] = Services::Events::ExtractArgument<std::string>(args).c_str();
+    eff->m_sParamString[0] = Services::Events::ExtractArgument<std::string>(args).c_str();
+
+    eff->m_nParamFloat[3] = Services::Events::ExtractArgument<float>(args);
+    eff->m_nParamFloat[2] = Services::Events::ExtractArgument<float>(args);
+    eff->m_nParamFloat[1] = Services::Events::ExtractArgument<float>(args);
+    eff->m_nParamFloat[0] = Services::Events::ExtractArgument<float>(args);
+
+    eff->SetNumIntegers(8); // allocate array
```
Liareth: I haven't looked at the NWN code for this, but is there any chance that `SetNumIntegers` allocates 8 integers, and then overwriting `m_nNumIntegers` causes the incorrect number of integers to be freed later?

Core/NWScript/nwnx.nss
```diff
@@ -31,77 +31,83 @@ void NWNX_PushArgumentInt(string pluginName, string functionName, int value);
...
-object NWNX_INTERNAL_GetReturnValueObject(string pluginName, string functionName, string type);
+string NWNX_INTERNAL_BuildString(string pluginName, string functionName, string operation)
+{
+    return "NWNXEE!ABIv2!" + pluginName + "!" + functionName + "!" + operation;
+}
 
 void NWNX_CallFunction(string pluginName, string functionName)
 {
-    NWNX_INTERNAL_CallFunction(pluginName, functionName);
+    PlaySound(NWNX_INTERNAL_BuildString(pluginName, functionName, "CALL"));
```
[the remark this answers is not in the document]
mtijanic: I wanted to use a separate function to keep it isolated. PlaySound() was a meaningful void function that only took a single string as an argument. May be others like that, didn't continue searching afterwards.

NWNXLib/Services/Events/Events.hpp
```diff
@@ -18,7 +19,25 @@ namespace Services {
 class Events : public ServiceBase
 {
 public: // Structures
-    using Argument = std::string;
+    struct Argument
+    {
+        Maybe<int32_t>              Int;
```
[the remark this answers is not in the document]
mtijanic: Noted, will fix.

NWNXLib/Services/Events/Events.inl
```diff
@@ -40,34 +72,15 @@ T Events::ExtractArgument(ArgumentStack& arguments)
...
-
-    if (!data)
-    {
-        throw std::runtime_error("Failed to match pushed arguments to the provided type.");
-    }
+    T real = std::move(data.Extract());
+    arguments.pop();
 
-    return data.Extract();
+    return std::move(real);
```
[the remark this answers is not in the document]
mtijanic: Good point. I'd expect the compiler to do return optimizations here anyway. I'll just `return real;`

NWNXLib/Utils.cpp
mtijanic: This isn't just `CVirtualMachine`, it's some other singleton that is surprisingly hard to get at. But I'll try.

Plugins/Effect/Effect.cpp
mtijanic: They are a continuous array. It'll just do `delete [] m_nParamInteger;`.

Somewhat related, but EffectDamage() actually uses more than 8. But since this is a limitation of a `util` module function and not the core, I'd rather leave at 8 which should cover all other uses, and if someone needs to do EffectDamage() they can elsewhere.

PR #285, opened 2019-01-14
New plugin: Visibility

Daztek:
Moved {Get/Set}VisibilityOverride out of NWNX_Player and added the ability to globally override the visibility state of an object.

Also made it possible to make objects visible to DMs only!

### DIFF
--- a/Plugins/Player/NWScript/nwnx_player.nss
+++ b/Plugins/Player/NWScript/nwnx_player.nss
@@ -1,6 +1,5 @@
 #include "nwnx"
 
-
 struct NWNX_Player_QuickBarSlot
 {
     object oItem;
@@ -17,10 +16,6 @@ struct NWNX_Player_QuickBarSlot
     int    nAssociateType;
     object oAssociate;
 };
-
-const int NWNX_PLAYER_VISIBILITY_DEFAULT = 0;
-const int NWNX_PLAYER_VISIBILITY_HIDDEN  = 1;
-const int NWNX_PLAYER_VISIBILITY_VISIBLE = 2;
 
 // Force display placeable examine window for player
 // If used on a placeable in a different area than the player, the portait will not be shown.
@@ -56,16 +51,6 @@ void NWNX_Player_SetQuickBarSlot(object player, int slot, struct NWNX_Player_Qui
 
 // Get the name of the .bic file associated with the player's character.
 string NWNX_Player_GetBicFileName(object player);
-
-// Overrides the default visibility rules about how player perceives the target object.
-// NWNX_PLAYER_VISIBILITY_DEFAULT - Restore normal behavior
-// NWNX_PLAYER_VISIBILITY_HIDDEN - Object is always hidden from the player
-// NWNX_PLAYER_VISIBILITY_VISIBLE - Object is always shown to the player
-void NWNX_Player_SetVisibilityOverride(object player, object target, int override);
-
-// Queries the existing visibility override for given (player, object) pair
-// Returns NWNX_PLAYER_VISIBILITY_DEFAULT if no override exists
-int NWNX_Player_GetVisibilityOverride(object player, object target);
 
 // Plays the VFX at the target position in current area for the given player only
 void NWNX_Player_ShowVisualEffect(object player, int effectId, vector position);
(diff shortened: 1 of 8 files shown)


Christian-Vassallo:
I had a problem with this plugin in the old version of nwnx (1.69).

Actually there were 3 options: default, visibile and invisibile.
The problem was that once visibile or invisibile were set, default case was not working anymore.

I had been able to solve the problem, cleaning the vismap that was created for the given target. 

Question: how this plugin works? Set visibile with the three options and all of them are working?

So I can set B visibile to A and invisibile to C and then put back That B is visible in standard way to A but still invisibile to C?

Silvard:
@Christian-Vassallo I think this is a global override. For an individual object to player map that's on the Player plugin, I believe.

Plugins/Player/NWScript/nwnx_player.nss
```diff
@@ -57,16 +52,6 @@ void NWNX_Player_SetQuickBarSlot(object player, int slot, struct NWNX_Player_Qui
 // Get the name of the .bic file associated with the player's character.
 string NWNX_Player_GetBicFileName(object player);
 
-// Overrides the default visibility rules about how player perceives the target object.
-// NWNX_PLAYER_VISIBILITY_DEFAULT - Restore normal behavior
-// NWNX_PLAYER_VISIBILITY_HIDDEN - Object is always hidden from the player
-// NWNX_PLAYER_VISIBILITY_VISIBLE - Object is always shown to the player
-void NWNX_Player_SetVisibilityOverride(object player, object target, int override);
```
mtijanic: Can't do this. Well, can, but shouldn't. Leave the old functions (and constants) in place, and make them call the new plugin instead, along with a deprecation warning to the log.

Plugins/Visibility/NWScript/nwnx_visibility.nss
```diff
@@ -0,0 +1,53 @@
+#include "nwnx"
+
+const int NWNX_VISIBILITY_DEFAULT   = -1;
+const int NWNX_VISIBILITY_VISIBLE   = 0;
+const int NWNX_VISIBILITY_HIDDEN    = 1;
```
mtijanic: Would be nice to have NWNX_VISIBILITY_DM_ONLY.

Plugins/Visibility/Visibility.cpp
```diff
@@ -0,0 +1,139 @@
...
+{
+    auto personalOverride = GetPersonalOverride(pPlayerGameObject->m_idSelf, pAreaObject->m_idSelf);
+    bool bInvisible = (personalOverride == -1) ? GetGlobalOverride(pAreaObject->m_idSelf) : personalOverride;
+    
+    return bInvisible ? false : g_plugin->m_TestObjectVisibilityHook->CallOriginal<int32_t>(pThis, pAreaObject, pPlayerGameObject);
+}
+
+bool Visibility::GetGlobalOverride(Types::ObjectID targetId)
+{
+    return g_plugin->m_GlobalVisibilityOverrideSet.find(targetId) != g_plugin->m_GlobalVisibilityOverrideSet.end();
```
mtijanic: Just use POS (on the target object) here as well.

Daztek:
@Silvard This plugin replaces the NWNX_Player ones

@Christian-Vassallo 
Here's the blurb for NWNX_Visibility_SetVisibilityOverride()
```
// Overrides the default visibility rules about how player perceives the target object
// If player is OBJECT_INVALID, the global visibility override will be set
//
// Player == VALID -> override:
//   NWNX_VISIBILITY_DEFAULT = Remove the player override
//   NWNX_VISIBILITY_VISIBLE = Target is visible
//   NWNX_VISIBILITY_HIDDEN  = Target is hidden
//
// Player == OBJECT_INVALID -> override:
//   NWNX_VISIBILITY_DEFAULT = Remove the global override
//   NWNX_VISIBILITY_VISIBLE = Target is visible
//   NWNX_VISIBILITY_HIDDEN  = Target is hidden
//   NWNX_VISIBLITY_DM_ONLY  = Target is visible to DMs only
//
// Note:
// Player state overrides the global state which means if a global state is set
// to NWNX_VISIBILITY_HIDDEN or NWNX_VISIBLITY_DM_ONLY but the player's state is
// set to NWNX_VISIBILITY_VISIBLE for the target, the object will be visible to the player
void NWNX_Visibility_SetVisibilityOverride(object player, object target, int override);
```

As you can see you can set a per player override or a global override, the player state overrides the global state.

The per player override is lost when the player leaves the server but the global override of a non player object should exist until the object is destroyed or the server is restarted.


>So I can set B visibile to A and invisibile to C and then put back That B is visible in standard way to A but still invisibile to C?

You should be able to.

Here's some examples:
```
// Set A is invisible to B
NWNX_Visibility_SetVisibilityOverride(b, a, NWNX_VISIBILITY_HIDDEN);

// Set C globally invisible
NWNX_Visibility_SetVisibilityOverride(OBJECT_INVALID, c, NWNX_VISIBILITY_HIDDEN);

// Set D visible to DMs Only
NWNX_Visibility_SetVisibilityOverride(OBJECT_INVALID, d, NWNX_VISIBILITY_DM_ONLY);

// Set D visible to E, even though its globally invisible to non DMs
NWNX_Visibility_SetVisibilityOverride(e, d, NWNX_VISIBILITY_VISIBLE);

// Remove D's global override
NWNX_Visibility_SetVisibilityOverride(OBJECT_INVALID, d, NWNX_VISIBILITY_DEFAULT);

// Remove E's personally override for D
NWNX_Visibility_SetVisibilityOverride(e, d, NWNX_VISIBILITY_DEFAULT);
```

mtijanic:
>The per player override is lost when the player leaves the server but the global override of a non player object should exist until the object is destroyed or the server is restarted.

I will fix this later, so that players relogging preserve their POS.

PR #286, opened 2019-01-16
Added the ability to obfuscate player names.

Silvard:
NWNX_SetPCNameOverride can be set to randomize playernames for characters with an override mapped, replace them with their affixless overrides or be left alone.

### DIFF
--- a/Plugins/Rename/NWScript/nwnx_rename.nss
+++ b/Plugins/Rename/NWScript/nwnx_rename.nss
@@ -2,14 +2,24 @@
 
 const string NWNX_Rename = "NWNX_Rename";
 
-//Set a player's floaty name(sPrefix+sNewName+sSuffix) and name on the player list (sNewName). Will not persist through saving, resets or logout.
-void NWNX_Rename_SetPlayerNameOverride(object oPlayer, string sPrefix, string sNewName, string sSuffix);
+//Constant
+const int NWNX_RENAME_PLAYERNAME_DEFAULT = 0;
+const int NWNX_RENAME_PLAYERNAME_OBFUSCATE = 1;
+const int NWNX_RENAME_PLAYERNAME_OVERRIDE = 2;
+
+//Set a PC's floaty/chat name(sPrefix+sNewName+sSuffix) and name (sNewName) on the player list.
+//If (iPlayerNameState) is set to NWNX_RENAME_PLAYERNAME_OVERRIDE the player name will change to (sNewName) on tells.
+//If (iPlayerNameState) is set to NWNX_RENAME_PLAYERNAME_OBFUSCATE the player name will be set to a random string.
+//If (iPlayerNameState) is set to NWNX_RENAME_PLAYERNAME_DEFAULT the player name will be untouched.
+//Will not persist through saving, resets or logout.
+void NWNX_Rename_SetPCNameOverride(object oPlayer, string sPrefix, string sNewName, string sSuffix, int iPlayerNameState);
 
 
-void NWNX_Rename_SetPlayerNameOverride(object oPlayer, string sPrefix, string sNewName, string sSuffix)
+void NWNX_Rename_SetPCNameOverride(object oPlayer, string sPrefix, string sNewName, string sSuffix, int iPlayerNameState)
 {
-    string sFunc = "SetPlayerNameOverride";
+    string sFunc = "SetPCNameOverride";
     
+    NWNX_PushArgumentInt(NWNX_Rename, sFunc, bOverridePlayerName);
     NWNX_PushArgumentString(NWNX_Rename, sFunc, sSuffix);
     NWNX_PushArgumentString(NWNX_Rename, sFunc, sNewName);
     NWNX_PushArgumentString(NWNX_Rename, sFunc, sPrefix);
--- a/Plugins/Rename/NWScript/nwnx_rename.nss
+++ b/Plugins/Rename/NWScript/nwnx_rename.nss
@@ -2,7 +2,7 @@
 
 const string NWNX_Rename = "NWNX_Rename";
 
-//Constant
+//Constants for Player Name states.
 const int NWNX_RENAME_PLAYERNAME_DEFAULT = 0;
 const int NWNX_RENAME_PLAYERNAME_OBFUSCATE = 1;
 const int NWNX_RENAME_PLAYERNAME_OVERRIDE = 2;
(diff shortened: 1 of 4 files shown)


Plugins/Rename/NWScript/nwnx_rename.nss
```diff
@@ -2,18 +2,29 @@
...
+const int NWNX_RENAME_PLAYERNAME_OBFUSCATE = 1;
+const int NWNX_RENAME_PLAYERNAME_OVERRIDE = 2;
 
+//Set a PC's floaty/chat name(sPrefix+sNewName+sSuffix) and name (sNewName) on the player list.
+//If (iPlayerNameState) is set to NWNX_RENAME_PLAYERNAME_OVERRIDE the player name will change to (sNewName) on tells.
+//If (iPlayerNameState) is set to NWNX_RENAME_PLAYERNAME_OBFUSCATE the player name will be set to a random string.
+//If (iPlayerNameState) is set to NWNX_RENAME_PLAYERNAME_DEFAULT the player name will be untouched.
+//DMs and the player themselves will still see their original player name regardless of option set
+//Will not persist through saving, resets or logout.
+void NWNX_Rename_SetPCNameOverride(object oPC, string sPrefix, string sNewName, string sSuffix, int iPlayerNameState);
```
mtijanic: ~~We generally want to keep the source backwards compatible, so maybe set default values on new args?~~

~~Belay that, we don't care much about compatibility within a single release (same NWN version)~~

_Actually_... Default arguments make some APIs much easier to do. How about this instead:
```C
void NWNX_Rename_SetPCNameOverride(object oPC, string sNewName, string sPrefix = "", string sSuffix = "", int iPlayerNameState = NWNX_RENAME_PLAYERNAME_DEFAULT);
```

Which means that for the average case, it's just `NWNX_Rename_SetPCNameOverride(oPC, sNewName);`

Plugins/Rename/Rename.cpp
```diff
@@ -95,74 +110,107 @@ CNWSPlayer *Rename::player(Types::ObjectID playerId)
...
-        std::string lsLastName = (bOriginal) ?  *g_plugin->GetServices()->m_perObjectStorage->Get<std::string>(pcObjectID,lastNameKey) : std::string("");
+        lsFirstName = (bOriginal) ?  *g_plugin->GetServices()->m_perObjectStorage->Get<std::string>(pcObjectID,firstNameKey) :  *g_plugin->GetServices()->m_perObjectStorage->Get<std::string>(pcObjectID,overrideNameKey);
+        lsLastName = (bOriginal) ?  *g_plugin->GetServices()->m_perObjectStorage->Get<std::string>(pcObjectID,lastNameKey) : std::string("");
         CNWSCreature* pCreature = server->GetCreatureByGameObjectID(pcObjectID);
         
         if (pCreature && !lsFirstName.empty() && ( bOriginal || !pCreature->m_sDisplayName.IsEmpty()))
         {
+            if (pPlayer != nullptr || static_cast<bool>(*g_plugin->GetServices()->m_perObjectStorage->Get<int>(pcObjectID, playerNameOverrideStateKey)) //if the pPlayer pointer is null (from a party invite) or the player name override is default...
```
mtijanic: Style nit, but can you please keep line width to say <120 characters. It's hard to read in diff mode, and also pretty complex to reason about. In this case I'd do:
```C
    if (pCreature && !lsFirstName.empty() && ( bOriginal || !pCreature->m_sDisplayName.IsEmpty()))
    {
        bool isDM = server->GetCreatureByGameObjectID(pPlayer->m_oidNWSObject)->m_pStats->m_bIsDM;
        bool noOverride = *g_plugin->GetServices()->m_perObjectStorage->Get<int>(pcObjectID, playerNameOverrideStateKey) != 0;
        if (pPlayer != nullptr || noOverride || isDM || pPlayer->m_oidNWSObject != pcObjectID)
        {
            if (bOriginal)
```

And now that I've written it out like that, I see:
```C
if (pPlayer != nullptr || ... || pPlayer->m_oidNWSObject != pcObjectID)
```
That's by definition invalid. The only time the OIDs can be compared is if `pPlayer == nullptr`. Because, if pPlayer is not null, the checks will end early and you'll enter the if. But if `pPlayer == nullptr`, your check for OIDs is invalid, because `pPlayer->m_oidNWSObject` will crash due to dereferencing a null pointer.

Plugins/Rename/Rename.cpp
```diff
@@ -95,74 +110,107 @@ CNWSPlayer *Rename::player(Types::ObjectID playerId)
...
 
+std::string Rename::GenerateRandomPlayerName(size_t length)
+{
+    static std::mt19937 rngEngine(std::random_device{}());
+    static const std::string charSet = "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789"; 
+    std::uniform_int_distribution<int> distribution(0, charSet.length() - 1);
+    std::string randomPlayername;
+    for (size_t i = 0; i < length; ++i)
+    {
+        randomPlayername += charSet[distribution(rngEngine)];
```
mtijanic: Each time you do this `+=`, new storage for the string is created, one byte larger than the previous, and previous is destroyed. So you do `length` allocations/deletes. Add `randomPlayername.reserve(length+1)` before the loop to reserve all the storage immediately.

Silvard:
Addressed the observations above.

Plugins/Rename/NWScript/nwnx_rename.nss
```diff
@@ -2,18 +2,29 @@
...
+void NWNX_Rename_SetPCNameOverride(object oPC, string sNewName, string sPrefix = "" , string sSuffix = "" , int iPlayerNameState = NWNX_RENAME_PLAYERNAME_DEFAULT);
 
-void NWNX_Rename_SetPlayerNameOverride(object oPlayer, string sPrefix, string sNewName, string sSuffix)
+
+void NWNX_Rename_SetPCNameOverride(object oPC, string sNewName, string sPrefix = "" , string sSuffix = "" , int iPlayerNameState = NWNX_RENAME_PLAYERNAME_DEFAULT)
 {
-    string sFunc = "SetPlayerNameOverride";
+    string sFunc = "SetPCNameOverride";
     
+    NWNX_PushArgumentInt(NWNX_Rename, sFunc, bOverridePlayerName);
```
mtijanic: Did you mean `iPlayerNameState` instead of `bOverridePlayerName` here?

Silvard: I indeed mean that.

PR #289, opened 2019-01-18
Couple of plugin unit tests + Administration Fix

Daztek:
Adds unit tests for the following plugins: Administration, Feedback, Visibility and Area

Also fixes a nullptr crash in Administration.


mtijanic:
You're awesome. Thanks!

Issue #139, opened 2018-04-18
Feature Request: Hide DMs from the Player List


Balanor:
It would be nice to see added functionality to nwnx_tweaks for an option to hide DMs from the player list.

PR #291, opened 2019-01-19
tweaks: Add ability to hide DMs on module char list

mtijanic:
Implements #139. Unfortunately, I couldn't easily make it so that if you're logging in as DM you can see them, since your DMness is tied to your characters. I can work around it if it is an important feature, but for now I think this will do.


Daztek:
Don't forget the documentation!

mtijanic:
Done. Those are pretty useless as they are, though.

PR #302, opened 2019-01-26
Update README.md for combatmodes plugin.

Urothis:


mtijanic:
We don't document this for any other plugin individually. Since it's the same for all plugins, it'd be better suited as some global documentation.

PR #308, opened 2019-01-28
The start of a changelog!

Daztek:
I think I got most stuff since build8186-2.

Feedback is welcome!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,98 @@
+# Changelog
+All notable changes to this project will be documented in this file.
+
+The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
+
+## [Unreleased]
+
+NOTICE: The ABI has changed, please make sure to update your nwnx.nss and recompile all your scripts. 
+
+### Added
+- New plugins: Area, CombatModes, Effect, Encounter, Feedback, Rename, Visibility
+- Added support to skip building certain plugins by setting an environment variable, for example: `NWNX_SKIP_PLUGINS="JVM;Lua;Mono;Ruby;SpellChecker"`
+- Added support to build out of tree plugins with the following environment variable: `NWNX_ADDITIONAL_PLUGINS={pathtoplugins}`
+- Core: Allow skipping all plugins on load with the following environment variable: `NWNX_CORE_SKIP_ALL=y`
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -0,0 +1,98 @@
...
+The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
+
+## [Unreleased]
+
+NOTICE: The ABI has changed, please make sure to update your nwnx.nss and recompile all your scripts. 
+
+### Added
+- New plugins: Area, CombatModes, Effect, Encounter, Feedback, Rename, Visibility
+- Added support to skip building certain plugins by setting an environment variable, for example: `NWNX_SKIP_PLUGINS="JVM;Lua;Mono;Ruby;SpellChecker"`
+- Added support to build out of tree plugins with the following environment variable: `NWNX_ADDITIONAL_PLUGINS={pathtoplugins}`
```
mtijanic: Perhaps better: `NWNX_ADDITIONAL_PLUGINS=/path/to/Plugin1;/path/to/Plugin2` so the `{}` don't confuse people, and also so it shows how to add more than one.

CHANGELOG.md
```diff
@@ -0,0 +1,98 @@
...
+- Core: Allow skipping all plugins on load with the following environment variable: `NWNX_CORE_SKIP_ALL=y`
```
mtijanic: Maybe different wording: "Allow changing default plugin state from 'loaded' to 'skipped' [...]. Use `NWNX_PLUGIN_SKIP=n` to enable specific plugins in this case."
or something to that effect. From this line alone, it's not clear why you'd ever do that.

CHANGELOG.md
```diff
@@ -0,0 +1,98 @@
...
+All notable changes to this project will be documented in this file.
+
+The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
+
+## [Unreleased]
+
+NOTICE: The ABI has changed, please make sure to update your nwnx.nss and recompile all your scripts. 
+
+### Added
+- New plugins: Area, CombatModes, Effect, Encounter, Feedback, Rename, Visibility
```
mtijanic: Probably better to break the list into one line per each, with a short description.

mtijanic:
Awesome stuff! Thanks!

Daztek:
Added clarifications and a link in README.md

PR #310, opened 2019-01-29
Map pin events

plenarius:
Map pin event handling

Can be used for example:
- persistence (updating db per pin instead of on_client_leave for example)
- sharing pins with party members (party defines one mapper and all party members get the pins)
- DMs can add pins for players by setting their own and propagating
- deny pin usage for low INT chars

Note: Vector.Z is not used for pins

### DIFF
--- a/Plugins/Events/Events/MapEvents.cpp
+++ b/Plugins/Events/Events/MapEvents.cpp
@@ -30,6 +30,7 @@ MapEvents::MapEvents(ViewPtr<Services::HooksProxy> hooker)
     hooker->RequestExclusiveHook<Functions::CNWSMessage__HandlePlayerToServerMapPinDestroyMapPin, int32_t,
         CNWSMessage*, CNWSPlayer*>(&HandleMapPinDestroyMapPinMessageHook);
     m_HandlePlayerToServerMapPinDestroyMapPinHook = hooker->FindHookByAddress(API::Functions::CNWSMessage__HandlePlayerToServerMapPinDestroyMapPin);
+
 }
 template <typename T>
 static T PeekMessage(CNWSMessage *pMessage, int32_t offset)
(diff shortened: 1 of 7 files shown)


Plugins/Events/Events/MapEvents.cpp
```diff
@@ -0,0 +1,146 @@
...
+
+    // Get number of bytes for the message
+    int len = PeekMessage<int32_t>(thisPtr, offset); offset += sizeof(len);
+
+    // Copy the string over
+    std::string note;
+    note.reserve(len+1);
+    std::memcpy(note.data(), thisPtr->m_pnReadBuffer + thisPtr->m_nReadBufferPtr + offset, len);
+    note[len] = '\0';
+    offset += len;
```
Daztek: `offset` is never used after this point?

Plugins/Events/Events/MapEvents.cpp
```diff
@@ -0,0 +1,147 @@
...
+    float y = PeekMessage<float>(thisPtr, offset); offset += sizeof(y);
+    float z = PeekMessage<float>(thisPtr, offset); offset += sizeof(z);
+
+    // Get number of bytes for the message
+    int len = PeekMessage<int32_t>(thisPtr, offset); offset += sizeof(len);
+
+    // Copy the string over
+    std::string note;
+    note.reserve(len+1);
+    note.assign(reinterpret_cast<const char*>(thisPtr->m_pnReadBuffer + thisPtr->m_nReadBufferPtr + offset), len);
```
mtijanic: Thanks ~~Obama~~ C++. Guess it'll have to do.

Plugins/Events/Events/MapEvents.cpp
```diff
@@ -0,0 +1,147 @@
...
+    note[len] = '\0';
+
+    Events::PushEventData("PIN_X", std::to_string(x));
+    Events::PushEventData("PIN_Y", std::to_string(y));
+    Events::PushEventData("PIN_NOTE", note);
+    
+    if (Events::SignalEvent("NWNX_ON_MAP_PIN_ADD_PIN", oidPlayer))
```
mtijanic: The Events convention is that you will fire both an `_BEFORE` and a `_AFTER` event.

PR #311, opened 2019-01-30
Add GetTotalEffectBonus()

plenarius:
So we have a custom implementation of Weird that ignores immunities to Fear but uses any fear bonuses in their Will and Fort calculations. `MySavingThrow` would spout out immunity feedback that was confusing to the end user so we needed to write our own calcs and we used `GetSavingThrowVersus` from the old nwnx. 

This exposes the actual bonus value beyond a player's base scores to attack, damage bonus, saves, skills ability scores,and touch attack provided by spells, equipment, potions etc.

Example usage:

```c
NWNX_Creature_GetTotalEffectBonus(oTarget, NWNX_EFFECT_TYPE_SAVING_THROW,
    oSaveVersus, 0, 0, nSavingThrow, SAVING_THROW_TYPE_FEAR);

NWNX_Creature_GetTotalEffectBonus(oTarget, NWNX_EFFECT_TYPE_SKILL,
    OBJECT_INVALID, 0, 0, -1, -1, SKILL_SPOT);

NWNX_Creature_GetTotalEffectBonus(oTarget, NWNX_EFFECT_TYPE_ABILITY,
    OBJECT_INVALID, 0, 0, -1, -1, -1, ABILITY_WISDOM);
```

### DIFF
--- a/Plugins/Creature/NWScript/nwnx_creature.nss
+++ b/Plugins/Creature/NWScript/nwnx_creature.nss
@@ -36,12 +36,12 @@ const int NWNX_CREATURE_CLERIC_DOMAIN_TRICKERY    = 19;
 const int NWNX_CREATURE_CLERIC_DOMAIN_WAR         = 20;
 const int NWNX_CREATURE_CLERIC_DOMAIN_WATER       = 21;
 
-const int NWNX_EFFECT_TYPE_ATTACK        = 1;
-const int NWNX_EFFECT_TYPE_DAMAGE        = 2;
-const int NWNX_EFFECT_TYPE_SAVING_THROW  = 3;
-const int NWNX_EFFECT_TYPE_ABILITY       = 4;
-const int NWNX_EFFECT_TYPE_SKILL         = 5;
-const int NWNX_EFFECT_TYPE_TOUCH_ATTACK  = 6;
+const int NWNX_CREATURE_BONUS_TYPE_ATTACK        = 1;
+const int NWNX_CREATURE_BONUS_TYPE_DAMAGE        = 2;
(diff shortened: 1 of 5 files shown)


Plugins/Creature/NWScript/nwnx_creature.nss
```diff
@@ -36,6 +36,13 @@ const int NWNX_CREATURE_CLERIC_DOMAIN_TRICKERY    = 19;
 const int NWNX_CREATURE_CLERIC_DOMAIN_WAR         = 20;
 const int NWNX_CREATURE_CLERIC_DOMAIN_WATER       = 21;
```
plenarius: Do you suppose I should rename these as `NWNX_EFFECT_BONUS_TYPE`?

Plugins/Creature/NWScript/nwnx_creature.nss
```diff
@@ -36,6 +36,13 @@ const int NWNX_CREATURE_CLERIC_DOMAIN_TRICKERY    = 19;
...
 
+const int NWNX_EFFECT_TYPE_ATTACK        = 1;
+const int NWNX_EFFECT_TYPE_DAMAGE        = 2;
+const int NWNX_EFFECT_TYPE_SAVING_THROW  = 3;
+const int NWNX_EFFECT_TYPE_ABILITY       = 4;
```
plenarius: Do you suppose I should rename these as `NWNX_EFFECT_BONUS_TYPE_`?

mtijanic: I think `NWNX_CREATURE_BONUS_TYPE_XXX`. So it's properly namespaced and doesn't collide with stuff defined in nwnx_effect.

Issue #31, opened 2018-01-29
NWNX_Structs


zunath:
https://github.com/NWNX/nwnx2-linux/blob/master/plugins/structs/nwnx_structs.nss

In the old NWNX_Structs plugin we had the ability to do a lot of things which manipulated internal structures. One feature in particular that I need is a way to add custom item properties onto items. I was able to do it before using code like this:

```

itemproperty ItemPropertyDirect (int nType, int nSubType, int nCostTable, int nCostValue, int nParamTable, int nParamValue)
{
    itemproperty ip = ItemPropertyAdditional(0);

    SetItemPropertyInteger(ip, 0, nType);
    SetItemPropertyInteger(ip, 1, nSubType);
    SetItemPropertyInteger(ip, 2, nCostTable);
    SetItemPropertyInteger(ip, 3, nCostValue);
    SetItemPropertyInteger(ip, 4, nParamTable);
    SetItemPropertyInteger(ip, 5, nParamValue);

    return ip;
}
```

Which created a normal item property and then turned around and overwrote its settings. It was a really good workaround for adding custom item properties to items.

Liareth:
For this to work, the core needs the ability to push and pop effects and item properties like it does the other primitive data types.

PR #315, opened 2019-02-02
ItemProperty plugin

mtijanic:
Just a utility pack/unpack like the effects, but most of the effect crap is irrelevant to IPs. Can still use the converter functions in nwnx_util to convert between the two if needed.

Finally fully implements #31


mtijanic:
And spell ID for #121

PR #324, opened 2019-02-08
Add ASSERT_OR_THROW()

### LINKED ISSUE #320: Assertion failures in functions called from nwscript should throw
We use `ASSERT(condition)` a lot to verify the user data is sane, which will give a nice message and stack trace. But bad data will still be used afterwards, leading to a crash.

We need a new type of assert (e.g. `ASSERT_THROW(condition)` that will generate a `std::runtime_exception` that we can use from the function handlers. An exception there will just jump out of the implementation back to nwscript, and all the results fetched afterwards will be invalid/default.

This makes the user less likely to crash the server with bad input - they'll just have bugs in their scripts.

Daztek:
Fixes #320

A nicer `ASSERT()` that will just jump out of function handlers with a runtime error exception.

Looks like this in the log:
```E [22:31:48] [NWNXLib] [{File}.cpp:{Line}] Plugin '{Plugin}' failed event '{Function Handler}'. Error: ASSERTION FAILURE: ({Condition}) in NWScript: {Script File}```

### DIFF
--- a/NWNXLib/Assert.hpp
+++ b/NWNXLib/Assert.hpp
@@ -31,7 +31,7 @@ namespace Assert {
     #define ASSERT_FAIL_MSG(format, ...) (void)0
 #endif
 
-    #define ASSERT_THROW(condition) \
+    #define ASSERT_OR_THROW(condition) \
         do \
         { \
             if(!(condition)) throw std::runtime_error("ASSERTION FAILURE: (" + std::string(#condition) + ") in NWScript: " + Utils::GetCurrentScript()); \
--- a/NWNXLib/Assert.hpp
+++ b/NWNXLib/Assert.hpp
@@ -31,12 +31,20 @@ namespace Assert {
     #define ASSERT_FAIL_MSG(format, ...) (void)0
 #endif
 
+    #define ASSERT_THROW(condition) \
+        do \
+        { \
+            if(!(condition)) ::NWNXLib::Assert::Throw((#condition)); \
+        } while (0)
+
 void Fail(const char* condition, const char* file, int line, const char* message);
 
 template <typename ... Args>
 void Fail(const char* condition, const char* file, int line, const char* format, Args ... args);
 
 void SetCrashOnFailure(bool crash);
+
+void Throw(const char* condition);
 
 #include "Assert.inl"
 
--- a/NWNXLib/Assert.hpp
+++ b/NWNXLib/Assert.hpp
@@ -1,5 +1,6 @@
 #pragma once
 
+#include "Utils.hpp"
 #include <cstdio>
 
 namespace NWNXLib {
--- a/NWNXLib/Assert.hpp
+++ b/NWNXLib/Assert.hpp
@@ -34,7 +34,7 @@ namespace Assert {
     #define ASSERT_THROW(condition) \
         do \
         { \
-            if(!(condition)) ::NWNXLib::Assert::Throw((#condition)); \
+            if(!(condition)) throw std::runtime_error("ASSERTION FAILURE: (" + std::string(#condition) + ") in NWScript: " + Utils::GetCurrentScript()); \
         } while (0)
 
 void Fail(const char* condition, const char* file, int line, const char* message);
@@ -43,8 +43,6 @@ template <typename ... Args>
 void Fail(const char* condition, const char* file, int line, const char* format, Args ... args);
 
 void SetCrashOnFailure(bool crash);
-
-void Throw(const char* condition);
 
 #include "Assert.inl"
 
--- a/Plugins/Creature/Creature.cpp
+++ b/Plugins/Creature/Creature.cpp
@@ -168,7 +168,7 @@ ArgumentStack Creature::AddFeatByLevel(ArgumentStack&& args)
         if (level > 0 && level <= pCreature->m_pStats->m_lstLevelStats.num)
         {
             auto *pLevelStats = pCreature->m_pStats->m_lstLevelStats.element[level-1];
-            ASSERT_OR_THROW(pLevelStats);
+            ASSERT(pLevelStats);
             pLevelStats->AddFeat(static_cast<uint16_t>(feat));
             pCreature->m_pStats->AddFeat(static_cast<uint16_t>(feat));
         }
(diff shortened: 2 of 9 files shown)

### CODE REVIEW

File: NWNXLib/Assert.hpp [hunk 1]
```diff
@@ -31,13 +31,21 @@ namespace Assert {
     #define ASSERT_FAIL_MSG(format, ...) (void)0
 #endif
 
+    #define ASSERT_THROW(condition) \
+        do \
+        { \
+            if(!(condition)) ::NWNXLib::Assert::Throw((#condition)); \
```
mtijanic: Just inline it directly. The function isn't adding anything useful, just an extra indirection step, and an extra step that has to be unwound when throwing.

Daztek: Inline'd! If this is good I'll start replacing all the `ASSERT()`s

mtijanic: Yes, very good! A most _minor_ nit is that I'd prefer the name `ASSERT_OR_THROW`, but I don't feel that strongly. Whatever you like better.

Daztek:
Should be good to merge now

Plugins/Creature/Creature.cpp
```diff
@@ -149,7 +149,7 @@ ArgumentStack Creature::AddFeat(ArgumentStack&& args)
     ArgumentStack stack;
     if (auto *pCreature = creature(args))
     {
-        const auto feat = Services::Events::ExtractArgument<int32_t>(args);  ASSERT(feat <= 65535);
+        const auto feat = Services::Events::ExtractArgument<int32_t>(args);  ASSERT_OR_THROW(feat <= 65535);
```
mtijanic: Can you add >= 0 as well.

Plugins/Creature/Creature.cpp
```diff
@@ -162,8 +162,8 @@ ArgumentStack Creature::AddFeatByLevel(ArgumentStack&& args)
     ArgumentStack stack;
     if (auto *pCreature = creature(args))
     {
-        const auto feat  = Services::Events::ExtractArgument<int32_t>(args);  ASSERT(feat <= 65535);
-        const auto level = Services::Events::ExtractArgument<int32_t>(args);  ASSERT(level <= 40);
+        const auto feat  = Services::Events::ExtractArgument<int32_t>(args);  ASSERT_OR_THROW(feat <= 65535);
```
mtijanic: `>= 0` as well

Plugins/Creature/Creature.cpp
```diff
@@ -162,8 +162,8 @@ ArgumentStack Creature::AddFeatByLevel(ArgumentStack&& args)
...
+        const auto level = Services::Events::ExtractArgument<int32_t>(args);  ASSERT_OR_THROW(level <= 40);
```
mtijanic: `>= 1` as well

Plugins/Creature/Creature.cpp
```diff
@@ -181,7 +181,7 @@ ArgumentStack Creature::RemoveFeat(ArgumentStack&& args)
     ArgumentStack stack;
     if (auto *pCreature = creature(args))
     {
-        const auto feat = Services::Events::ExtractArgument<int32_t>(args);  ASSERT(feat <= 65535);
+        const auto feat = Services::Events::ExtractArgument<int32_t>(args);  ASSERT_OR_THROW(feat <= 65535);
```
mtijanic: `>= 0`

Plugins/Creature/Creature.cpp
```diff
@@ -192,43 +192,43 @@ ArgumentStack Creature::RemoveFeat(ArgumentStack&& args)
 ArgumentStack Creature::GetKnowsFeat(ArgumentStack&& args)
 {
     ArgumentStack stack;
-    int32_t retval = 0;
+    int32_t retVal = 0;
     if (auto *pCreature = creature(args))
     {
-        const auto feat = Services::Events::ExtractArgument<int32_t>(args);  ASSERT(feat <= 65535);
+        const auto feat = Services::Events::ExtractArgument<int32_t>(args);  ASSERT_OR_THROW(feat <= 65535);
```
mtijanic: `>= 0`

Plugins/Creature/Creature.cpp
```diff
@@ -192,43 +192,43 @@ ArgumentStack Creature::RemoveFeat(ArgumentStack&& args)
...
 
 ArgumentStack Creature::GetFeatCountByLevel(ArgumentStack&& args)
 {
     ArgumentStack stack;
-    int32_t retval = -1;
+    int32_t retVal = -1;
     if (auto *pCreature = creature(args))
     {
-        const auto level = Services::Events::ExtractArgument<int32_t>(args);  ASSERT(level <= 40);
+        const auto level = Services::Events::ExtractArgument<int32_t>(args);  ASSERT_OR_THROW(level <= 40);
```
mtijanic: `>= 1`

Plugins/Creature/Creature.cpp
```diff
@@ -192,43 +192,43 @@ ArgumentStack Creature::RemoveFeat(ArgumentStack&& args)
...
 
         if (level > 0 && level <= pCreature->m_pStats->m_lstLevelStats.num)
```
mtijanic: huh, I see we already check for `level > 0`, so maybe not necessary. But I think it's still better to give a meaningful error than silently ignore. If you add `>= 1`, you can also remove this.

Plugins/Creature/Creature.cpp
```diff
@@ -192,43 +192,43 @@ ArgumentStack Creature::RemoveFeat(ArgumentStack&& args)
...
 
 ArgumentStack Creature::GetFeatByLevel(ArgumentStack&& args)
 {
     ArgumentStack stack;
-    int32_t retval = -1;
+    int32_t retVal = -1;
     if (auto *pCreature = creature(args))
     {
-        const auto level = Services::Events::ExtractArgument<int32_t>(args);  ASSERT(level <= 40);
+        const auto level = Services::Events::ExtractArgument<int32_t>(args);  ASSERT_OR_THROW(level <= 40);
```
mtijanic: etc. Not much point in making comments on each one of these...

Plugins/Creature/Creature.cpp
```diff
@@ -1385,7 +1385,7 @@ ArgumentStack Creature::SetBaseSavingThrow(ArgumentStack&& args)
     if (auto *pCreature = creature(args))
     {
         const auto which = Services::Events::ExtractArgument<int32_t>(args);
-        const auto value = Services::Events::ExtractArgument<int32_t>(args); ASSERT(value >= -128 && value <= 127);
+        const auto value = Services::Events::ExtractArgument<int32_t>(args); ASSERT_OR_THROW(value >= -128 && value <= 127);
```
mtijanic: It's generally better if we split and `&&` conditions into two separate asserts. That way when it fires, you know which one it was.

Daztek:
Should be good to merge now, part 2 :D

PR #326, opened 2019-02-09
WIP: Constants do-over

mtijanic:
A bit of namespacing, smaller headers, and better type safety for constants. WIP.

Also includes a ton of whitespace changes to make all files conform to single style.


Plugins/Events/Events/DMActionEvents.cpp
```diff
@@ -159,38 +160,58 @@ void DMActionEvents::HandleDMMessageHook(Services::Hooks::CallType type,
...
-        case 0x88:
+        case MessageDungeonMasterMinor::SetDate:
             event += "SET_DATE";
             break;
-
+        case MessageDungeonMasterMinor::SetFactionReputation:
+            event += "SET_FACTION)REPUTATION";
+            break;
+        case MessageDungeonMasterMinor::GetFactionReputation:
+            event += "GET_FACTION)REPUTATION";
```
Daztek: Set/Get Faction Reputation have a `)` instead of `_`

mtijanic:
Merging this half-baked to avoid conflicts with other work. Other than the war on magic numbers and whitespace, this massive PR also:

- Changes GetAsciiTable() so that first element is 1 instead of 32
- Adds a few more DM events
- Fixes a few copy paste errors where multiple actions fired the same nwnx event
- Changes some ASSERTs to ASSERT_OR_THROW where it makes sense
- Use `Utils` for casting instead of `static_cast` in a few places.

PR #332, opened 2019-02-11
add -fno-strict-aliasing

mtijanic:
Strict aliasing is a horrible concept that shouldn't ever exist anyway. But especially in NWNX that does a whole lot of hackery the compiler has no hope of understanding.

It must die.


mtijanic:
Someone who can explain why this sucks much more eloquently than I can: http://www.yodaiken.com/2018/06/07/torvalds-on-aliasing/

PR #335, opened 2019-02-11
resolves #168

### LINKED ISSUE #168: Change CMakeLists.txt to use CMAKE_CURRENT_BINARY_SOURCE.
From the project root CMakeLists.txt

set(CMAKE_RUNTIME_OUTPUT_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}/Binaries)
set(CMAKE_LIBRARY_OUTPUT_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}/Binaries)

Change those to CMAKE_CURRENT_BINARY_DIR, otherwise you cannot make a clean build out of source.

hendrikgit:
I wanted to help out and looked at the "good first issues" and this was one that came up. I put resolves #168 in the title because I read that is one of the keywords to later auto close the issue when this is merged. I'm not that confident it is resolved though. It is probably not as simple as making a search & replace or someone would have done it already. I have this as a starting point though and if someone can give me some pointers on what change is needed I'll try to get that done.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -42,8 +42,8 @@ function(add_plugin_windows target)
 endfunction()
 
 # Sets the output directory for the built targets.
-set(CMAKE_RUNTIME_OUTPUT_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}/Binaries)
-set(CMAKE_LIBRARY_OUTPUT_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}/Binaries)
+set(CMAKE_RUNTIME_OUTPUT_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR})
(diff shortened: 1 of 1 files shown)


mtijanic:
"resolves"! I was wondering why "fixes #XXX" works, but "implements #XXX" doesn't, and it always seemed weird to "fix" a feature request. Makes sense now, thanks!

Anyway, the build is failing because it looks like it is now putting files in the root unified/ directory, instead of unified/Binaries. I think `{CMAKE_CURRENT_BINARY_DIR}/Binaries` might be a solution, but there's probably a better one where we tell cmake what to set CMAKE_CURRENT_BINARY_DIR in the first place.

hendrikgit:
Closing this PR since I don't know enough about this to be of help here.

PR #354, opened 2019-02-20
Creature: add {Get/Set}OriginalName()

### LINKED ISSUE #353: Feature Request: Change name in .bic file
Currently the names plugin for PCs allows the masking of character name in the playerlist, and so on, but does not support changing the name in the .bic file. In NWNX2, there was the NWNX_Names_SetName and NWNX_Names_GetName functions; if those could be updated for EE, that would be very useful to replace some of the functionality our module had at that time.

Daztek:
Resolves #353 

Adds {Get/Set}OriginalName() to NWNX_Creature.

### DIFF
--- a/Plugins/Creature/Creature.cpp
+++ b/Plugins/Creature/Creature.cpp
@@ -1749,7 +1749,6 @@ ArgumentStack Creature::SetOriginalName(ArgumentStack&& args)
     if (auto *pCreature = creature(args))
     {
         const auto name = Services::Events::ExtractArgument<std::string>(args);
-          ASSERT_OR_THROW(!name.empty());
         const auto isLastName = Services::Events::ExtractArgument<int32_t>(args);
 
         CExoLocString locName;
@@ -1761,6 +1760,7 @@ ArgumentStack Creature::SetOriginalName(ArgumentStack&& args)
         }
         else
         {
+            ASSERT_OR_THROW(!name.empty());
             pCreature->m_pStats->m_lsFirstName = locName;
         }
     }
--- a/Plugins/Creature/Creature.cpp
+++ b/Plugins/Creature/Creature.cpp
@@ -123,6 +123,8 @@ Creature::Creature(const Plugin::CreateParams& params)
     REGISTER(GetFeatTotalUses);
     REGISTER(SetFeatRemainingUses);
     REGISTER(GetTotalEffectBonus);
+    REGISTER(SetOriginalName);
+    REGISTER(GetOriginalName);
 
 #undef REGISTER
 }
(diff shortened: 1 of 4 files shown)


Plugins/Creature/NWScript/nwnx_creature.nss
```diff
@@ -310,6 +310,15 @@ void NWNX_Creature_SetFeatRemainingUses(object creature, int feat, int uses);
 int NWNX_Creature_GetTotalEffectBonus(object creature, int bonusType=NWNX_CREATURE_BONUS_TYPE_ATTACK, object target=OBJECT_INVALID, int isElemental=0,
     int isForceMax=0, int savetype=-1, int saveSpecificType=-1, int skill=-1, int abilityScore=-1, int isOffhand=FALSE);
 
+// Set the original first or last name of creature
+//
+// For PCs this will persist to the .bic file if saved. Requires a relog to update.
+void NWNX_Creature_SetOriginalName(object creature, string name, int isLastName);
+
+// Get the original first or last name of creature
+string NWNX_Creature_GetOriginalName(object creature, int isLastName);
```
mtijanic: Getter already provided in the base game:
```C
// Set the name of oObject.
//
// - oObject: the object for which you are changing the name (area, creature, placeable, item, or door).
// - sNewName: the new name that the object will use.
// Note: SetName() does not work on player objects.
//       Setting an object's name to "" will make the object
//       revert to using the name it had originally before any
//       SetName() calls were made on the object.
string GetName(object oObject, int bOriginalName=FALSE);
```

Plugins/Creature/Creature.cpp
```diff
@@ -1739,4 +1741,60 @@ ArgumentStack Creature::GetTotalEffectBonus(ArgumentStack&& args)
...
 }
+
+ArgumentStack Creature::SetOriginalName(ArgumentStack&& args)
+{
+    ArgumentStack stack;
+
+    if (auto *pCreature = creature(args))
+    {
+        const auto name = Services::Events::ExtractArgument<std::string>(args);
+          ASSERT_OR_THROW(!name.empty());
```
mtijanic: Setting the last name to `""` should be okay.

Plugins/Creature/NWScript/nwnx_creature.nss
Daztek: True, but it returns the full name so you can't differentiate between the first or last name.

mtijanic: Yep, true. Both of those could have spaces. Let's leave it in. :)

Plugins/Creature/Creature.cpp
Daztek: Moved the assert to firstname only

PR #360, opened 2019-02-20
FindItemWithBaseItemID event, used by out of ammo search for next stack

### LINKED ISSUE #225: Feature request: Possibility to influence ammo reload selection (on reload event?)
The goal of this would be to allow the player to select to shoot only, for example, fire arrows and have only those being auto reloaded.

When the current equipped stack of arrows runs out and the game looks for more arrows it would be nice if there was an event or some way to call a script. In this nwscript the object to load next could be specified, by searching through the characters inventory for specific tags on arrow stacks or whatever means one chooses.

plenarius:
This function is used by the engine to search for the next stack of ammunition when the creature runs out of its existing stack. Resolves #225 

Does there need to be more sanity checking on the engine side?
* Make sure the item belongs to the creature
* Make sure the item still matches the base item
* Make sure they're allowed to equip the item

I'm not really sure the nTh instant of the baseitem is ever anything but 0 in its current state but should this function be used for more in the future it's there for completeness.

### DIFF
--- a/Plugins/Events/Events/ItemEvents.cpp
+++ b/Plugins/Events/Events/ItemEvents.cpp
@@ -229,13 +229,9 @@ uint32_t ItemEvents::FindItemWithBaseItemIdHook(CItemRepository* thisPtr, uint32
 
     if (Events::SignalEvent("NWNX_ON_ITEM_AMMO_RELOAD_AFTER", thisPtr->m_oidParent, &sAfterEventResult))
     {
-        retVal = m_FindItemWithBaseItemIdHook->CallOriginal<uint32_t>(thisPtr, baseItem, nTh);
-    }
-    else
-    {
-        retVal = stoul(sAfterEventResult, nullptr, 16);
-        if (ItemSanityCheck(retVal))
-            return retVal;
+        uint32_t result = stoul(sAfterEventResult, nullptr, 16);
+        if (ItemSanityCheck(result))
+            return result;
     }
 
     return retVal;
--- a/Plugins/Events/Events/ItemEvents.cpp
+++ b/Plugins/Events/Events/ItemEvents.cpp
@@ -177,9 +177,7 @@ uint32_t ItemEvents::FindItemWithBaseItemIdHook(CItemRepository* thisPtr, uint32
     // unless the player has > 255 stacks of ammo they can't equip
     // Suppose I could set this to a higher value but I think this is safe
     if(nTh > 255)
-    {
         return OBJECT_INVALID;
-    }
 
     auto ItemSanityCheck = [&](uint32_t objectId) -> bool {
         if (static_cast<Types::ObjectID>(objectId) == Constants::OBJECT_INVALID)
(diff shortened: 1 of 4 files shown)


Plugins/Events/Events/ItemEvents.cpp
```diff
@@ -149,4 +154,80 @@ void ItemEvents::RemoveItemHook(Services::Hooks::CallType type, CItemRepository*
...
+
+    std::string sBeforeEventResult;
+    std::string sAfterEventResult;
+
+    uint32_t retVal;
+
+    Events::PushEventData("BASE_ITEM_ID", std::to_string(baseItem));
+    Events::PushEventData("BASE_ITEM_NTH", std::to_string(nTh));
+
+    if (Events::SignalEvent("NWNX_ON_ITEM_FIND_ITEM_WITH_BASE_ITEMID_BEFORE", thisPtr->m_oidParent, &sBeforeEventResult))
```
mtijanic: This seems like an overly generic event on an overly generic function that just so happens to not be used much currently. I'm thinking just rename the event to `NWNX_ON_AMMO_RELOAD` or something, and add an early return to the function if the base item is not one of the ammo types.

That way, future changes to the game won't cause this to randomly fire in other cases.

Plugins/Events/Events/ItemEvents.cpp
```diff
@@ -149,4 +154,80 @@ void ItemEvents::RemoveItemHook(Services::Hooks::CallType type, CItemRepository*
...
+
+    // We're most assuredly in an infinite loop here caused by the scripting end
+    // unless the player has > 255 stacks of ammo they can't equip
+    // Suppose I could set this to a higher value but I think this is safe
+    if(nTh > 255)
+    {
+        return OBJECT_INVALID;
+    }
+
+    auto ItemSanityCheck = [&](uint32_t objectId) -> uint32_t {
```
mtijanic: A "sanity check" function should probably just return a bool, and then caller can CallOriginal() if the check failed.

Plugins/Events/Events/ItemEvents.cpp
```diff
@@ -149,4 +154,80 @@ void ItemEvents::RemoveItemHook(Services::Hooks::CallType type, CItemRepository*
...
+        return ItemSanityCheck(stoul(sBeforeEventResult, nullptr, 16));
+    }
+
+    Events::PushEventData("BASE_ITEM_ID", std::to_string(baseItem));
+    Events::PushEventData("BASE_ITEM_NTH", std::to_string(nTh));
+    Events::PushEventData("ACTION_RESULT", Utils::ObjectIDToString(retVal));
+
+    if (Events::SignalEvent("NWNX_ON_ITEM_FIND_ITEM_WITH_BASE_ITEMID_AFTER", thisPtr->m_oidParent, &sAfterEventResult))
+    {
+        retVal = m_FindItemWithBaseItemIdHook->CallOriginal<uint32_t>(thisPtr, baseItem, nTh);
```
mtijanic: Don't call this twice. We already know what it returned after the BEFORE call. The user can override it in the after, but if they don't, just use existing `retVal` value.

PR #369, opened 2019-02-21
Reveal Plugin

Silvard:
This plugin has functions to selectively reveal a stealthed character to another character (for the next detection check), or to toggle whether the stealthed character will always be revealed to members of their party.

### DIFF
--- a/Plugins/Reveal/Reveal.cpp
+++ b/Plugins/Reveal/Reveal.cpp
@@ -16,14 +16,12 @@ static ViewPtr<Reveal::Reveal> g_plugin;
 //key names for Per Object Storage
 const std::string revealKey = "REVEAL";
 
-
-
 NWNX_PLUGIN_ENTRY Plugin::Info* PluginInfo()
 {
     return new Plugin::Info
     {
         "Reveal",
-        "Functions to allow the selective revealing of a stealthing character to another character or their party.",
+        "Functions to allow the selective revealing of a stealthed character to another character or their party.",
         "Silvard",
         "jusenkyo at gmail.com",
         1,
@@ -51,7 +49,6 @@ Reveal::Reveal(const Plugin::CreateParams& params)
 #undef REGISTER
 
     GetServices()->m_hooks->RequestExclusiveHook<Functions::CNWSCreature__DoStealthDetection, int32_t,CNWSCreature*,CNWSCreature*, int32_t, int32_t*, int32_t*, int32_t>(&HookStealthDetection);
-
     m_DoStealthDetection = GetServices()->m_hooks->FindHookByAddress(Functions::CNWSCreature__DoStealthDetection);
 }   
 
@@ -74,14 +71,13 @@ int32_t Reveal::HookStealthDetection(NWNXLib::API::CNWSCreature* pObserverCreatu
             }
             if (static_cast<bool>(*g_plugin->GetServices()->m_perObjectStorage->Get<int>(pHidingCreature->m_idSelf, revealKey + Utils::ObjectIDToString(pObserverCreature->m_idSelf))))
             {
-                g_plugin->GetServices()->m_perObjectStorage->Remove(pHidingCreature->m_idSelf, revealKey + Utils::ObjectIDToString(pObserverCreature->m_idSelf));
+                g_plugin->GetServices()->m_perObjectStorage->Remove(pHidingCreature->m_idSelf, revealKey + Utils::ObjectIDToString(pObserverCreature->m_idSelf)); //remove mapping after first check
                 *bHeard = true;
                 return true;
             }
         }
     }
     return g_plugin->m_DoStealthDetection->CallOriginal<int32_t>(pObserverCreature, pHidingCreature, bClearLOS, bSeen, bHeard, bTargetInvisible);
-    
 }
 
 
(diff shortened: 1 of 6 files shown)


Plugins/Reveal/Reveal.cpp
```diff
@@ -0,0 +1,107 @@
...
+Reveal::~Reveal()
+{
+}
+int32_t Reveal::HookStealthDetection(NWNXLib::API::CNWSCreature* pObserverCreature, NWNXLib::API::CNWSCreature* pHidingCreature, int32_t bClearLOS, int32_t* bSeen, int32_t* bHeard, int32_t bTargetInvisible)
+{
+    if (pObserverCreature->m_bPlayerCharacter && pHidingCreature->m_bPlayerCharacter && pHidingCreature->m_nStealthMode)
+    {
+        if (pObserverCreature->GetArea() == pHidingCreature->GetArea())
+        {
+            if (static_cast<bool>(*g_plugin->GetServices()->m_perObjectStorage->Get<int>(pHidingCreature->m_idSelf, revealKey + "PARTY")))
```
mtijanic: Could you do a `pPOS->` thing here as well. Keep the column size to something reasonable.

mtijanic: Also, `static_cast<bool>` is just noise, you can check an int in an if.

Plugins/Reveal/Reveal.cpp
```diff
@@ -0,0 +1,107 @@
...
+            {
+                if (pObserverCreature->GetFaction()->GetLeader() == pHidingCreature->GetFaction()->GetLeader())
+                {
+                    *bHeard = true;
```
mtijanic: Why only bHeard, and not bSeen? (add a comment if there's a reason)

Plugins/Reveal/Reveal.cpp
```diff
@@ -0,0 +1,107 @@
...
+                    return true;
+                }
+            }
+            if (static_cast<bool>(*g_plugin->GetServices()->m_perObjectStorage->Get<int>(pHidingCreature->m_idSelf, revealKey + Utils::ObjectIDToString(pObserverCreature->m_idSelf))))
```
mtijanic: As above

Plugins/Reveal/Reveal.cpp
```diff
@@ -0,0 +1,107 @@
...
+            {
+                g_plugin->GetServices()->m_perObjectStorage->Remove(pHidingCreature->m_idSelf, revealKey + Utils::ObjectIDToString(pObserverCreature->m_idSelf)); //remove mapping after first check
```
mtijanic: So if you see someone, next time the check fires is when they go out of range again? If so, please add a comment.

Christian-Vassallo:
Actually this could be possible also with visibility plugin using some trick.

If you set visibile somebody, the object is visibile always (invisibility, hide, etc) or default or invisibile

Plugins/Reveal/Reveal.cpp
weavejester: I'm not the author of the PR, but IIRC a "heard" creature is rendered as semi-transparent, whereas a "seen" creature is rendered as opaque.

It would be nice to have heard/seen as an option to the plugin, so you can reveal a creature through sound or through vision. I don't think this is something that the visibility plugin covers, though I could be wrong.

Daztek: I'm not sure why it's even set because it returns true right after and never gets used.

mtijanic: It's an `out` parameter. The calling function uses it. And weavejester's explanation makes sense, but it'd be nice to document (or better yet expose)

Daztek:
Minor thing: update the changelog :D

Silvard:
Added the ability to specify whether the stealther is revealed through being seen or heard, and cleaned it up a bit.

Plugins/Reveal/Reveal.cpp
```diff
@@ -0,0 +1,128 @@
...
+}
+int32_t Reveal::HookStealthDetection(NWNXLib::API::CNWSCreature* pObserverCreature, NWNXLib::API::CNWSCreature* pHidingCreature, int32_t bClearLOS, int32_t* bSeen, int32_t* bHeard, int32_t bTargetInvisible)
+{
+    Services::PerObjectStorageProxy* pPOS = g_plugin->GetServices()->m_perObjectStorage.get();
+    
+    if (pObserverCreature->m_bPlayerCharacter && pHidingCreature->m_bPlayerCharacter && pHidingCreature->m_nStealthMode)
+    {
+        if (pObserverCreature->GetArea() == pHidingCreature->GetArea())
+        {
+            if (*pPOS->Get<int>(pHidingCreature->m_idSelf, revealKey + "PARTY"))
```
mtijanic: I'm not sure this is so well defined. `pPos->Get<int>(...)` will return a `Maybe<int>`, and `*` of that will be the int. But if the value does not exist, dereferencing it with * is undefined.

I think you want:
```C
auto reveal = pPOS->Get<int>(pHidingCreature->m_idSelf, revealKey + "PARTY");
if (reveal && *reveal)
{
    ...
```
And same for other cases

PR #372, opened 2019-02-22
Events: add EffectEvents

### LINKED ISSUE #350: Feature request: effect add/remove events for the Events plugin
The new Effect plugin allows a script to be tied to an effects expiration (#314), but it looks like a more general event is possible, that would trigger when *any* effect is added or removed.

May I suggest adding four new events to the Events plugin:

* `NWNX_ON_ADD_EFFECT_BEFORE`
* `NWNX_ON_ADD_EFFECT_AFTER`
* `NWNX_ON_REMOVE_EFFECT_BEFORE`
* `NWNX_ON_REMOVE_EFFECT_AFTER`

These would tie into the internal events:

* `CNWSEffectListHandler__OnEffectApplied`
* `CNWSEffectListHandler__OnEffectRemoved`

My reason for suggesting this is that it allows events to be triggered by effects, no matter what the source. For example, if a strength increase effect is applied to a player, maybe it triggers a bark, or perhaps NPCs around notice when a flashy effect is applied to a player.

More specifically, we were playing around with adding a damage increase to the dexterity modifier on our server, and it would be useful to catch effects globally so that any effect that increases or decreases dexterity can be accounted for.

Daztek:
Resolves #350

I think this should work okay, only handles temporary/permanent effects that aren't VisualEffects/ItemProperty effects

Test/Example:
```
#include "nwnx_events"
#include "nwnx_object"

void main()
{
    object oPlayer = OBJECT_SELF;
    string sEvent = NWNX_Events_GetCurrentEvent();
    int bBefore = FindSubString(sEvent, "_BEFORE") != -1;

    string effectID = NWNX_Events_GetEventData("ID");
    object oCreator = NWNX_Object_StringToObject(NWNX_Events_GetEventData("CREATOR"));
    string effectType = NWNX_Events_GetEventData("TYPE");

    if (FindSubString(sEvent, "NWNX_ON_EFFECT_APPLIED_") != -1)
    {
        SendMessageToPC(GetFirstPC(), (bBefore ? "BEFORE: " : "AFTER: '") + GetName(oPlayer) + "' effect applied!" +
                                                                            "\nID: " + effectID +
                                                                            "\nCreator: " +GetName(oCreator) +
                                                                            "\nType: " + effectType);
        if (bBefore)
        {
            if (StringToInt(effectType) == 36)
            {
                // 36 is ability increase, Param 1 is Type, Param 2 is Value
                // See: effect EffectAbilityDecrease(int nAbility, int nModifyBy)
                SendMessageToPC(oPlayer, "Applied: Ability: " + NWNX_Events_GetEventData("INT_PARAM_1") + ", Increase: " + NWNX_Events_GetEventData("INT_PARAM_2"));
            }
        }
    }

    if (FindSubString(sEvent, "NWNX_ON_EFFECT_REMOVED_") != -1)
    {
        SendMessageToPC(GetFirstPC(), (bBefore ? "BEFORE: " : "AFTER: '") + GetName(oPlayer) + "' effect removed!" +
                                                                            "\nID: " + effectID +
                                                                            "\nCreator: " +GetName(oCreator) +
                                                                            "\nType: " + effectType);

        if (!bBefore)
        {
            if (StringToInt(effectType) == 36)
            {
                // 36 is ability increase, Param 1 is Type, Param 2 is Value
                // See: effect EffectAbilityDecrease(int nAbility, int nModifyBy)
                SendMessageToPC(oPlayer, "Removed: Ability: " + NWNX_Events_GetEventData("INT_PARAM_1") + ", Increase: " + NWNX_Events_GetEventData("INT_PARAM_2"));
            }
        }
    }
}
```


weavejester:
Thanks! This works perfectly for me.

PR #375, opened 2019-02-22
Object: CheckFit()

### LINKED ISSUE #362: [Request] CheckFreeSpaceFitsInventory
CheckFreeSpaceFitsInventory (int Width, int Height)
Function checked the compatibility of the proportions size W and H and free space in inventory for this proportions?

int Width, int Height  i.e. InvSlotWidth and InvSlotHeight in baseitems.2da
To set the required int value.

plenarius:
Resolves #362. Given an object with inventory and a baseitem, checks if there's room to place an item of baseitem type into that object. Works for creatures, placeables and container items.


Plugins/Object/Object.cpp
```diff
@@ -329,4 +335,49 @@ ArgumentStack Object::GetHasVisualEffect(ArgumentStack&& args)
...
+        else if (auto *pItem = Utils::AsNWSItem(pObject))
+            pRepo = pItem->m_pItemRepository;
+        else
+        {
+            Services::Events::InsertArgument(stack, retVal);
+            return stack;
+        }
+        retVal = 0;
+        const auto baseitem = Services::Events::ExtractArgument<int32_t>(args);
+        static CNWSItem *tmp = new CNWSItem(Constants::OBJECT_INVALID);
```
mtijanic: I'll get around to fixing this, but for now "good enough"

PR #387, opened 2019-02-24
SpellEvents: Set|ClearMemorizedSpell

### LINKED ISSUE #366: Spellbook manipulation events
Fire events when a PC memorizes or unmemorizes a spell. Should be bypassable.

plenarius:
Resolves #366. The UI doesn't update if the event is bypassed, well it requires a refresh. Also should we do anything with the return value on the Set?

### DIFF
--- a/Plugins/Events/Events/SpellEvents.cpp
+++ b/Plugins/Events/Events/SpellEvents.cpp
@@ -1,19 +1,33 @@
 #include "Events/SpellEvents.hpp"
 #include "API/CNWSCreature.hpp"
+#include "API/CNWSCreatureStats.hpp"
+#include "API/CNWRules.hpp"
+#include "API/CNWSpellArray.hpp"
 #include "API/Functions.hpp"
+#include "API/Globals.hpp"
 #include "Events.hpp"
 #include "Utils.hpp"
 
 namespace Events {
 
 using namespace NWNXLib;
+using namespace NWNXLib::API;
+
 
 static Hooking::FunctionHook* m_SpellCastAndImpactHook = nullptr;
+static Hooking::FunctionHook* m_SetMemorizedSpellSlotHook = nullptr;
+static Hooking::FunctionHook* m_ClearMemorizedSpellSlotHook = nullptr;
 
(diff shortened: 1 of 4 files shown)


Plugins/Events/NWScript/nwnx_events.nss
```diff
@@ -326,6 +326,35 @@
         COUNTERING_SPELL        int         Returns TRUE if cast as counter else FALSE
         PROJECTILE_PATH_TYPE    int
         IS_INSTANT_SPELL        int         Returns TRUE if spell was instant else FALSE
+
+    NWNX_SET_MEMORIZED_SPELL_SLOT_BEFORE
+    NWNX_SET_MEMORIZED_SPELL_SLOT_AFTER
+
+    Usage:
+        OBJECT_SELF = The creature whose memorizing the spell
```
mtijanic: s/whose/who's/g

Plugins/Events/NWScript/nwnx_events.nss
```diff
@@ -326,6 +326,35 @@
...
+
+    Event data:
+        Variable Name           Type        Notes
+        SPELL_MULTICLASS        int
```
mtijanic: Think it's worth noting what this means (index of the class)

Plugins/Events/Events/SpellEvents.cpp
```diff
@@ -62,4 +79,72 @@ void SpellEvents::CastSpellHook
...
+    Events::PushEventData("SPELL_FROMCLIENT", std::to_string(fromClient));
+    Events::PushEventData("ACTION_RESULT", std::to_string(retVal));
+
+    Events::SignalEvent("NWNX_SET_MEMORIZED_SPELL_SLOT_AFTER", thisPtr->m_pBaseCreature->m_idSelf, &sAfterEventResult);
+
+    retVal = sAfterEventResult.empty() ? retVal : sAfterEventResult == "1";
+
+    if(retVal)
+    {
+        // Do we do anything if the user bypasses but returns a value?
```
mtijanic: Don't think so.

PR #396, opened 2019-03-03
[Deprecated] added NWNX_ON_ATTACK event and functions to modify damage dealt

henningkoehlernz:
Example usage:

```
#include "nwnx_events"
...
NWNX_Events_SubscribeEvent("NWNX_ON_ATTACK", "on_attack");
```
triggering
```
#include "nwnx_events"
#include "nwnx_object"

void main()
{
    object oAttacker = OBJECT_SELF;
    object oTarget = NWNX_Object_StringToObject(NWNX_Events_GetEventData("TARGET_OBJECT_ID"));
    string sAttackNr = NWNX_Events_GetEventData("ATTACK_NUMBER");
    string sDamage = NWNX_Events_GetEventData("ATTACK_DAMAGE_0");
    int nDamType;
    for ( nDamType = 1; nDamType < 13; nDamType++ )
    {
        sDamage += "/" + NWNX_Events_GetEventData("ATTACK_DAMAGE_" + IntToString(nDamType));
    }
    string sSneak = NWNX_Events_GetEventData("IS_SNEAK_ATTACK");
    string sOffhand = NWNX_Events_GetEventData("IS_OFFHAND_ATTACK");
    string sMessage = "Original damage = " + sDamage + " (Sneak=" + sSneak + ", Offhand=" + sOffhand + ")";

    SendMessageToPC(oAttacker, "You attack " + GetName(oTarget) + " (#" + sAttackNr + ").");
    SendMessageToPC(oAttacker, sMessage);
    SendMessageToPC(oTarget, GetName(oAttacker) + " attacks you (#" + sAttackNr + ").");
    SendMessageToPC(oTarget, sMessage);

    // increase damage if no sneak attack damage was dealt
    if ( sSneak == "0" && GetHasFeat(FEAT_SNEAK_ATTACK, oAttacker) )
        NWNX_Events_OnAttack_AddDamage(2);
}
```

### DIFF
--- a/Plugins/Events/Events/CombatEvents.cpp
+++ b/Plugins/Events/Events/CombatEvents.cpp
@@ -39,7 +39,7 @@ CombatEvents::CombatEvents(ViewPtr<HooksProxy> hooker, ViewPtr<EventsProxy> eve)
                 &StartCombatRoundHook
             );
     });
-    Events::InitOnFirstSubscribe("NWNX_ON_ATTACK", [hooker]() {
+    Events::InitOnFirstSubscribe("NWNX_ON_ATTACK_.*", [hooker]() {
         hooker->RequestSharedHook<
             API::Functions::CNWSCombatRound__SetCurrentAttack,
             int32_t,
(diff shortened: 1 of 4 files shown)


Plugins/Events/Events/CombatEvents.cpp
```diff
@@ -42,4 +63,56 @@ void CombatEvents::StartCombatRoundHook(
...
+        }
+        Events::PushEventData("ATTACK_RESULT", std::to_string(combatAttackData->m_nAttackResult));
+        Events::PushEventData("IS_SNEAK_ATTACK", std::to_string(combatAttackData->m_bSneakAttack || combatAttackData->m_bDeathAttack));
+        Events::PushEventData("IS_OFFHAND_ATTACK", std::to_string(combatAttackData->m_nWeaponAttackType == 2));
+        Events::SignalEvent("NWNX_ON_ATTACK", thisPtr->m_pBaseCreature->m_idSelf);
+        g_combatEvents->m_combatAttackData = nullptr;
+    }
+}
+
+ArgumentStack CombatEvents::AddAttackDamage(ArgumentStack&& args)
```
Daztek: This is better done with `SetEventResult()`, see [here](https://github.com/nwnxee/unified/blob/master/Plugins/Events/Events/StealthEvents.cpp#L50-L76), [here](https://github.com/nwnxee/unified/blob/master/Plugins/Events/Events/ClientEvents.cpp#L55-L84) and [here](https://github.com/nwnxee/unified/blob/master/Plugins/Events/Events/ItemEvents.cpp#L231-L242) for examples.

Plugins/Events/Events/CombatEvents.cpp
```diff
@@ -28,6 +39,16 @@ CombatEvents::CombatEvents(ViewPtr<HooksProxy> hooker)
                 &StartCombatRoundHook
             );
     });
+    Events::InitOnFirstSubscribe("NWNX_ON_ATTACK", [hooker]() {
+        hooker->RequestSharedHook<
+            API::Functions::CNWSCombatRound__SetCurrentAttack,
+            int32_t,
+            API::CNWSCombatRound*,
+            unsigned char>
```
Daztek: You can use an `uint8_t` here so you don't have to cast later on.

Plugins/Events/Events/CombatEvents.cpp
```diff
@@ -42,4 +63,56 @@ void CombatEvents::StartCombatRoundHook(
     Events::SignalEvent(before ? "NWNX_ON_START_COMBAT_ROUND_BEFORE" : "NWNX_ON_START_COMBAT_ROUND_AFTER" , thisPtr->m_pBaseCreature->m_idSelf);
 }
 
+void CombatEvents::AttackHook(
+    Hooks::CallType type,
+    API::CNWSCombatRound* thisPtr,
+    unsigned char attackNumber)
```
Daztek: Same as above.

Plugins/Events/Events/CombatEvents.cpp
```diff
@@ -42,4 +63,56 @@ void CombatEvents::StartCombatRoundHook(
...
+    {
+        uint16_t damageType = Services::Events::ExtractArgument<int32_t>(args);
+        int32_t damageAmount = Services::Events::ExtractArgument<int32_t>(args);
+        g_combatEvents->m_combatAttackData->AddDamage(damageType, damageAmount);
+    }
+
+    return stack;
+}
+
+ArgumentStack CombatEvents::SetAttackDamage(ArgumentStack&& args)
```
Daztek: Same as above.

Plugins/Events/Events/CombatEvents.hpp
```diff
@@ -20,7 +23,17 @@ class CombatEvents
         NWNXLib::API::CNWSCombatRound*,
         uint32_t
     );
-    static NWNXLib::ViewPtr<NWNXLib::Hooking::FunctionHook> m_hook;
+    static void AttackHook
+    (
+        NWNXLib::Services::Hooks::CallType,
+        NWNXLib::API::CNWSCombatRound*,
+        unsigned char
```
Daztek: `uint8_t` again!

Daztek:
It's best to use `SetEventResult()` to pass data back to the event and let the user write nwscript wrappers for it.

Plugins/Events/Events/CombatEvents.cpp
```diff
@@ -42,4 +63,56 @@ void CombatEvents::StartCombatRoundHook(
...
+        Events::PushEventData("ATTACK_NUMBER", std::to_string(static_cast<int>(attackNumber)));
+        for ( int i = 0; i < 13; i++ )
+        {
+            int16_t nDamage = combatAttackData->m_nDamage[i];
+            Events::PushEventData("ATTACK_DAMAGE_" + std::to_string(i), std::to_string(nDamage));
+        }
+        Events::PushEventData("ATTACK_RESULT", std::to_string(combatAttackData->m_nAttackResult));
+        Events::PushEventData("IS_SNEAK_ATTACK", std::to_string(combatAttackData->m_bSneakAttack || combatAttackData->m_bDeathAttack));
+        Events::PushEventData("IS_OFFHAND_ATTACK", std::to_string(combatAttackData->m_nWeaponAttackType == 2));
+        Events::SignalEvent("NWNX_ON_ATTACK", thisPtr->m_pBaseCreature->m_idSelf);
```
Daztek: We generally signal a *_BEFORE and *_AFTER event, see the other events for an example.

Plugins/Events/Events/CombatEvents.cpp
henningkoehlernz: The problem is that the result isn't a single integer, but an array of integers (one per damage type). So at a minimum we'd need to worry about serialization - are there functions readily available for this (similar to argument/return serialization for function calls)?

Plugins/Events/Events/CombatEvents.cpp
henningkoehlernz: Didn't do this since there doesn't seem to be any notable difference between the two - in both cases the script would run after damage has been calculated but before it is being dealt.

Plugins/Events/Events/CombatEvents.cpp
henningkoehlernz: Simplified the functionality to only deal with base damage. More restrictive but much simpler (using SetEventResult now), and probably all that's really needed (other damage type modifications can be handled via the Damage plugin).

Daztek:
This looks good, although I'm wondering if it might be better off in NWNX_Damage

henningkoehlernz:
Alternative implementation in damage plugin found here: https://github.com/nwnxee/unified/pull/404

Daztek:
Closing this in favor of #404

PR #398, opened 2019-03-04
added GetDamageImmunity function to Object plugin

henningkoehlernz:
Usage:
```
#include "nwnx_object"
int nImmunity = NWNX_Object_GetDamageImmunity(oTarget, DAMAGE_TYPE_PIERCING);
```

### DIFF
--- a/Plugins/Object/NWScript/nwnx_object.nss
+++ b/Plugins/Object/NWScript/nwnx_object.nss
@@ -135,6 +135,8 @@ int NWNX_Object_GetHasVisualEffect(object obj, int nVFX);
 // Return TRUE if an item of baseitem type can fit in object's inventory
 int NWNX_Object_CheckFit(object obj, int baseitem);
 
+// Return damage immunity (in percent) against given damage type
+int NWNX_Object_GetDamageImmunity(object obj, int damageType);
 
 const string NWNX_Object = "NWNX_Object";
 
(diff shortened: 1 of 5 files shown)


Plugins/Object/NWScript/nwnx_object.nss
```diff
@@ -135,6 +135,8 @@ int NWNX_Object_GetHasVisualEffect(object obj, int nVFX);
 // Return TRUE if an item of baseitem type can fit in object's inventory
 int NWNX_Object_CheckFit(object obj, int baseitem);
 
+// Return damage immunity (in percent) against given damage type
+int NWNX_Object_GetDamageImmunity(object obj, int damageType);
```
Daztek: Should probably mention here that damageType takes DAMAGE_TYPE_* constants and what it actually returns.

Plugins/Object/Object.cpp
```diff
@@ -346,4 +347,18 @@ ArgumentStack Object::CheckFit(ArgumentStack&& args)
...
 }
+
+ArgumentStack Object::GetDamageImmunity(ArgumentStack&& args)
+{
+    ArgumentStack stack;
+    int32_t retVal = -1;
+
+    if (auto *pObject = object(args))
+    {
+        const uint16_t damageFlags = Services::Events::ExtractArgument<int32_t>(args);
```
Daztek: Minor thing, but you can use `auto` instead of `uint16_t` and let the compiler figure out what's best.

Since `GetDamageImmunityByFlags()` takes an unsigned int, I'd also throw in a `ASSERT_OR_THROW(damageFlags >= 0);`

Daztek:
Looks good aside from what I mentioned above.

Also, please add it to CHANGELOG.md and if possible write a small unit test for `nwnx_object_t.nss`!

Plugins/Object/NWScript/nwnx_object.nss
henningkoehlernz: What's unclear about the return value description?

Daztek: I totally read over the `(in percent)`, sorry :D

PR #417, opened 2019-03-17
Util: IsValidResRef()

plenarius:
Fun time figuring out the constants for ResRef type. The default type was added for convenience but maybe it should be ITEM, or maybe there just shouldn't be a default?

### DIFF
--- a/Plugins/Util/NWScript/nwnx_util.nss
+++ b/Plugins/Util/NWScript/nwnx_util.nss
@@ -1,5 +1,17 @@
 #include "nwnx"
 
+const int NWNX_UTIL_RESREF_TYPE_AREA_GIT  = 2012;
+const int NWNX_UTIL_RESREF_TYPE_ITEM      = 2025;
+const int NWNX_UTIL_RESREF_TYPE_CREATURE  = 2027;
+const int NWNX_UTIL_RESREF_TYPE_DIALOG    = 2029;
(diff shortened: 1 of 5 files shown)


Plugins/Util/NWScript/nwnx_util.nss
```diff
@@ -18,6 +30,8 @@ effect NWNX_Util_ItemPropertyToEffect(itemproperty ip);
 string NWNX_Util_GenerateUUID();
 // Strip any color codes from a string
 string NWNX_Util_StripColors(string str);
+// Is a valid resref
+int NWNX_Util_IsValidResRef(string filename, int type=NWNX_UTIL_RESREF_TYPE_CREATURE);
```
Daztek: Nitpick: I'd use `resref` instead of `filename`

Even smaller nitpick: spaces before and after the `=` for type :D

Daztek:
I assume it works fine for custom objects, too?

Plugins/Util/NWScript/nwnx_util.nss
plenarius: I copied how the default was done in `NWNX_Util_GetCurrentScriptName(int depth=0)` so blame sherincall (fixed that one too)

plenarius:
Yes, works with custom objects.

PR #424, opened 2019-03-24
Update the environment variable names in the Redis plugin documentation

hendrikgit:
I just set up the Redis plugin and noticed that these are the environment variable names to use. I tried it for all except NWNX_REDIS_PUBSUB_CHANNELS but I assume it follows the same schema as the rest.


Daztek:
If you could fix the ones under `Getting started with PubSub` too, that'd be cool!

PR #437, opened 2019-04-03
Core: Module Shutdown Script + Docker Clean Shutdown

Daztek:
Adds `NWNX_CORE_SHUTDOWN_SCRIPT` as an environment variable which lets you set a nwscript that gets executed when the server is shutdown.

Also changes some docker stuff to correctly propagate the SIGTERM from `docker stop` and `docker-compose down` to nwserver, resulting in a clean shutdown.

The docker changes appear to be working fine for me, but some more testing would be great.


Dockerfile
```diff
@@ -7,14 +7,20 @@ RUN Scripts/buildnwnx.sh -j $(nproc)
 FROM beamdog/nwserver
 RUN mkdir /nwn/nwnx
 COPY --from=builder /nwnx/home/Binaries/* /nwn/nwnx/
+
+# Copy our modified run-server.sh
+COPY --from=builder /nwnx/home/Scripts/Docker/run-server.sh /nwn/
```
jakkn: Keep in mind that overwriting this file means it has to be maintained if the original should change. I'm not sure how likely it is that this will happen, but as long as it's possible the frequency is irrelevant imo. Any wrong behavior could pass by silently as the build is not bound to fail, and chasing down the root cause might also be hard.

An alternative would be to check in only the output of `diff run-server.sh.orig run-server.sh.new > run-server.patch` and to do
```Dockerfile
COPY --from=builder /nwnx/home/Scripts/Docker/run-server.patch /nwn/
RUN patch /nwnx/run-server.sh < /nwn/run-server.patch
```
This way `run-server.sh` continues to get updates from upstream, and a failed patch will show up as a failed build on hub with a reference to the conflicting line.

Daztek: Good point, now with patching!

PR #460, opened 2019-04-18
added NWNX_Creature_GetHighestLevelOfFeat

henningkoehlernz:
Exposes another NWN function. Needed to get correct number of total/remaining uses for feats with multiple levels, like barbarian rage.


Daztek:
A test in nwnx_creature_t.nss would be nice!

PR #475, opened 2019-05-11
added DESTROY_OBJECT and DECREMENT_STACKSIZE hooks to Events/ItemEvents

henningkoehlernz:
Can e.g. be used to create infinite ammunition:

on module load:
```
#include "nwnx_events"
void main()
{
    NWNX_Events_SubscribeEvent("NWNX_ON_ITEM_DECREMENT_STACKSIZE_BEFORE", "on_decr_stack");
}
```
on_decr_stack:
```
#include "nwnx_events"
void main()
{
   if ( IsMarkedAsInfinite(OBJECT_SELF) )
       NWNX_Events_SkipEvent();
}
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,7 +21,7 @@ NOTICE: The ABI has changed, please make sure to update your nwnx.nss and recomp
 - Events: Added On{Un}Polymorph events as PolymorphEvents
 - Events: Added Event Data for SpawnObject, GiveItem, GiveAlignment, Heal, Kill, ToggleInvulnerabilty, ForceRest, Limbo, ToggleAI, ToggleImmortal, Goto, Possess, PossessFullPower, ToggleLock, DisableTrap, JumpToPoint, JumpTargetToPoint, JumpAllPlayersToPoint, ChangeDifficulty, ViewInventory, SpawnTrapOnObject events in DMActionEvents
 - Events: Added OnClientConnect events that fire before the player sees the server vault.
-- Events: Added ItemInventory{Open/Close}, ItemInventory{Add/Remove}Item and ScrollLearn events to ItemEvents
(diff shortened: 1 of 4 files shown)


Daztek:
Might be good to mention that this conflicts with the (Item)ObjectEventHandler target of NWNX_Profiler. ObjectEventHandler profiling is off by default so it's not a super big deal, but adding a note is probably good.

CHANGELOG.md
```diff
@@ -110,6 +110,7 @@ The following plugins were added:
 - Encounter: GetResetTime()
 - Encounter: SetResetTime()
 - Events: GetCurrentEvent()
+- Events: ON_ITEM_EVENT_{BEFORE|AFTER}
```
Daztek: Minor, this should be added to `Events: Added ItemInventory{Open/Close}, ItemInventory{Add/Remove}Item and ScrollLearn events to ItemEvents` under the `Added` header in the changelog.

Plugins/Events/NWScript/nwnx_events.nss
```diff
@@ -729,6 +739,11 @@ const int NWNX_EVENTS_TIMING_BAR_LOCK          = 8;
 const int NWNX_EVENTS_TIMING_BAR_CUSTOM        = 10;
 */
 
+/*
+const int NWNX_EVENTS_ITEM_EVENT_DESTROY_OBJECT = 11;
+const int NWNX_EVENTS_ITEM_EVENT_DECREMENT_STACKSIZE = 16;
+*/
```
Daztek: Are these the only events Items have?

henningkoehlernz: These are the only ones I've managed to trigger. The 1.69 decompiled code also handles EVENT_TIMED_EVENT, EVENT_APPLY_EFFECT and EVENT_REMOVE_EFFECT, but at least the latter two are not handled anymore (signature changed and ApplyEffectToObject doesn't trigger anything).

PR #490, opened 2019-06-04
Util: Get2DARowCount()

plenarius:
Will be required for upcoming Race plugin to efficiently loop through the race and racialmod rows.

### DIFF
--- a/Plugins/Util/Util.cpp
+++ b/Plugins/Util/Util.cpp
@@ -2,6 +2,7 @@
 
 #include "API/Constants.hpp"
 #include "API/Globals.hpp"
+#include "API/C2DA.hpp"
 #include "API/CResRef.hpp"
 #include "API/CExoResMan.hpp"
 #include "API/CExoString.hpp"
@@ -64,6 +65,7 @@ Util::Util(const Plugin::CreateParams& params)
     REGISTER(GetMinutesPerHour);
     REGISTER(SetMinutesPerHour);
     REGISTER(EncodeStringForURL);
+    REGISTER(GetTwoDARowCount);
 
 #undef REGISTER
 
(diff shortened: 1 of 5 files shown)


Daztek:
My OCD would prefer for the NWScript function to be `Get2DARowCount()` so it's in line with `Get2DAString()` :D

Plugins/Util/Util.cpp
```diff
@@ -258,4 +260,14 @@ ArgumentStack Util::EncodeStringForURL(ArgumentStack&& args)
     return stack;
 }
 
+ArgumentStack Util::GetTwoDARowCount(ArgumentStack&& args)
+{
+    ArgumentStack stack;
+    const auto twodaRef = Services::Events::ExtractArgument<std::string>(args);
+    auto *twoda = new C2DA(CResRef(twodaRef.c_str()), 0);
```
mtijanic: This is leaking the memory. If you're using `new`, you need to use `delete` at the end. Or, better yet, you can use `unique_ptr<C2DA>` and it will automatically release it when the function exits (this will happen even if you exit due to an exception, etc)

mtijanic: Or, better-better-yet, just use a local variable:
```C
C2DA twoda(CResRef(twodaRef.c_str()), 0);
twoda.Load2DArray();
... twoda.m_nNumRows...
````

Comment (plenarius) [reply Plugins/Util/Util.cpp]: As per discord discussion, we used `Globals::Rules()->m_p2DArrays->GetCached2DA(twodaRef.c_str(), true);` as the most efficient means to retrieve the info.

File: Plugins/Util/Util.cpp [hunk 2]
```diff
@@ -258,4 +262,13 @@ ArgumentStack Util::EncodeStringForURL(ArgumentStack&& args)
     return stack;
 }
 
+ArgumentStack Util::Get2DARowCount(ArgumentStack&& args)
+{
+    ArgumentStack stack;
+    const auto twodaRef = Services::Events::ExtractArgument<std::string>(args);
+    auto twoda = Globals::Rules()->m_p2DArrays->GetCached2DA(twodaRef.c_str(), true);
+    Services::Events::InsertArgument(stack, twoda->m_nNumRows);
```
Daztek: This crashes if you pass an empty string or a 2da that doesn't exist. This should fix it:

`Services::Events::InsertArgument(stack, !twoda ? 0 : twoda->m_nNumRows);`

mtijanic: Sure, just please write it as `twoda ? twoda->m_nNumRows : 0`. :)

PR #516, opened 2019-06-21
Player: GetQuestCompleted() Fix missing semi colon

plenarius:


Plugins/Player/NWScript/nwnx_player.nss
```diff
@@ -168,7 +168,7 @@ void NWNX_Player_SetPlaceableNameOverride(object player, object placeable, strin
 
 // Get's whether a quest has been completed by a player
```
Daztek: Gets :P

PR #522, opened 2019-07-04
Chat: SetChatHearingDistance()

### LINKED ISSUE #499: Allow modifying a creatures hearing distance for chat/whisper
This is mostly hardcoded in `CNWSMessage::SendServerToPlayerChatMessage`, but might be a few other spots. 

Should have an option to get/set a creatures hearing distance in the POS, and then use that instead of the default 20.0f/3.0f values if given.
Will likely require reimplementing the whole function.

plenarius:
Resolves #499 - We might want to consider doing the opposite where a PC's voice can carry beyond the hardcoded limit.


Plugins/Chat/Chat.cpp
```diff
@@ -5,14 +5,22 @@
 #include "API/CNWSDungeonMaster.hpp"
 #include "API/CNWSMessage.hpp"
 #include "API/CNWSPlayer.hpp"
+#include "API/CNWSObject.hpp"
 #include "API/Constants.hpp"
+#include "API/CServerInfo.hpp"
 #include "API/CServerExoApp.hpp"
+#include "API/CServerExoAppInternal.hpp"
+#include "API/CServerExoAppInternal.hpp"
```
Daztek: Once is probably enough :P

PR #528, opened 2019-07-17
New race plugin features:

Morderon:
When effect vs parent race is applied, applies the effect vs each child race as well. This includes weapon properties.

Child races can wear gear restricted to a parent race, as well as gear restricted to their own.


Daztek:
This seems to have some errant commits?

Morderon:
Oh yeah sorry about that, should be all good now.

PR #534, opened 2019-07-23
Object: add AddIconEffect and RemoveIconEffect

weavejester:
This adds two new functions to the object plugin:

* `NWNX_Object_AddIconEffect(object obj, int nIcon, float fDuration=0.0)`
* `NWNX_Object_RemoveIconEffect(object obj, int nIcon)`

These functions add effect icons without the associated effect. This will be considerably more useful in the next version of NWN, when the hardcoded limit for effect icons is removed, but if there are some effects you know won't be used, you can replace their icons with custom ones in the `effecticons.2da`.

Icon effects work with `ApplyEffectToObject` and `RemoveEffect`, but don't seem to work with many other effect functions, such as `TagEffect`, `EffectLinkEffects`, `GetFirstEffect` and `GetNextEffect`. For this reason it seems reasonable to hide the implementation behind functions, and not to expose the actual `effect` pointer in NWScript.

Adding an icon effect overwrites any icon effect with the same icon already on the object. This means there's some code in common between `AddIconEffect` and `RemoveIconEffect` that I wasn't sure what to do with. Current the lines of code in common are duplicated, but they could be factored out into a common private function if that would be better design.

### DIFF
--- a/Plugins/Object/Object.cpp
+++ b/Plugins/Object/Object.cpp
@@ -35,8 +35,6 @@ using namespace NWNXLib::API;
 
 static ViewPtr<Object::Object> g_plugin;
 
-const auto MODULE_OID = 0;
-
 NWNX_PLUGIN_ENTRY Plugin::Info* PluginInfo()
 {
     return new Plugin::Info
@@ -642,7 +640,7 @@ ArgumentStack Object::AddIconEffect(ArgumentStack&& args)
         }
 
         auto *effIcon = new API::CGameEffect(true);
-        effIcon->m_oidCreator = MODULE_OID;
+        effIcon->m_oidCreator = 0;
         effIcon->m_nType      = Constants::EffectTrueType::Icon;
         effIcon->m_nSubType   = Constants::EffectSubType::Supernatural;
         effIcon->m_bShowIcon  = true;
--- a/Plugins/Object/Object.cpp
+++ b/Plugins/Object/Object.cpp
@@ -35,6 +35,8 @@ using namespace NWNXLib::API;
 
 static ViewPtr<Object::Object> g_plugin;
 
+const auto MODULE_OID = 0;
+
 NWNX_PLUGIN_ENTRY Plugin::Info* PluginInfo()
 {
     return new Plugin::Info
@@ -85,6 +87,8 @@ Object::Object(const Plugin::CreateParams& params)
     REGISTER(SetAutoRemoveKey);
     REGISTER(GetTriggerGeometry);
     REGISTER(SetTriggerGeometry);
+    REGISTER(RemoveIconEffect);
+    REGISTER(AddIconEffect);
 
 #undef REGISTER
 }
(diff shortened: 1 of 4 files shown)


Plugins/Object/Object.cpp
```diff
@@ -594,4 +598,72 @@ ArgumentStack Object::SetTriggerGeometry(ArgumentStack&& args)
...
+    {
+        const auto nIcon = Services::Events::ExtractArgument<int32_t>(args);
+
+        for (int i = 0; i < pObject->m_appliedEffects.num; i++)
+        {
+            auto *eff = pObject->m_appliedEffects.element[i];
+
+            if (eff->m_sCustomTag == "NWNX_Object_IconEffect" && eff->m_nParamInteger[0] == nIcon)
+            {
+                pObject->RemoveEffect(eff);
```
Daztek: Since these icon effects are unique, you can add a `break;` here to exit the for loop when the effect has been found

Plugins/Object/Object.cpp
```diff
@@ -594,4 +598,72 @@ ArgumentStack Object::SetTriggerGeometry(ArgumentStack&& args)
...
+    return stack;
+}
+
+ArgumentStack Object::AddIconEffect(ArgumentStack&& args)
+{
+    ArgumentStack stack;
+
+    if (auto *pObject = object(args))
+    {
+        const auto nIcon = Services::Events::ExtractArgument<int32_t>(args);
```
Daztek: An `ASSERT_OR_THROW(nIcon > 0);` would be nice here

Plugins/Object/Object.cpp
```diff
@@ -594,4 +598,72 @@ ArgumentStack Object::SetTriggerGeometry(ArgumentStack&& args)
...
+        const auto fDuration = Services::Events::ExtractArgument<float>(args);
+
+        for (int i = 0; i < pObject->m_appliedEffects.num; i++)
+        {
+            auto *eff = pObject->m_appliedEffects.element[i];
+
+            if (eff->m_sCustomTag == "NWNX_Object_IconEffect" && eff->m_nParamInteger[0] == nIcon)
+            {
+                pObject->RemoveEffect(eff);
```
Daztek: `break;` again!

Plugins/Object/Object.cpp
```diff
@@ -594,4 +598,72 @@ ArgumentStack Object::SetTriggerGeometry(ArgumentStack&& args)
...
+            }
+        }
+
+        auto *effIcon = new API::CGameEffect(true);
+        effIcon->m_oidCreator = MODULE_OID;
```
Daztek: You can just use `0` here or if you want to be fancy, `Utils::GetModule()->m_idSelf;`

I'm debating just adding a module oid constant to the global constants

Plugins/Object/Object.cpp
weavejester: Should I also add it to `RemoveIconEffect`?

weavejester:
Changes should all be made now.

PR #538, opened 2019-07-29
Initialize CombatModes fields

Qowyn:
Otherwise CombatModes plugin may prevent all changes to combat modes if used without nwnx_event plugin.


Daztek:
Whoops, nice catch!

PR #540, opened 2019-08-03
Tweak: Dead creature fires on area exit

### LINKED ISSUE #493: Tweak: Run OnAreaExit event for dead characters too
`CNWSArea::RemoveObjectFromArea()` has an explicit check to see if the leaving creature is dead, and doesn't run the scripts in that case.
Should have a tweak that eliminates the check. The function is tiny and can just be reimplented entirely.

plenarius:
Includes the addition of nwn event related constants and some functions to execute some ScriptEvents through Utils

This resolves #493


Daztek:
Pls fix conflicts :D

PR #554, opened 2019-08-22
Player: SetPersistentLocation

plenarius:
This function is mainly used so after a server restart the PC doesn't need to load the starting area before being jumped to their persistent location.

It is probably best executed `OnModuleLoad`.  Builders create a waypoint where they want a PC to start then key that waypoint by cd key (or community name depending on vault config) and bic file. 

Example usage in `OnModuleLoad` of creating a persistent location for any PCs that have logged on in the last four weeks.
```c
    string sSQL = "SELECT public_cdkey, bic_file, start_loc FROM player_tracking WHERE TIMESTAMPDIFF(DAY,last_seen,NOW()) < 28";
    NWNX_SQL_PrepareQuery(sSQL);
    NWNX_SQL_ExecutePreparedQuery();
    while (NWNX_SQL_ReadyToReadNextRow())
    {
        NWNX_SQL_ReadNextRow();
        string sKey = NWNX_SQL_ReadDataInActiveRow(0);
        string sResRef = NWNX_SQL_ReadDataInActiveRow(1);
        location lLoc = NWNXStringToLocation(NWNX_SQL_ReadDataInActiveRow(2));
        object oWP = CreateObject(OBJECT_TYPE_WAYPOINT, "nw_waypoint001", lLoc, FALSE);
        if (GetArea(oWP) != OBJECT_INVALID)
            NWNX_Player_SetPersistentLocation(sKey, sResRef, oWP);
    }
```


plenarius:
Another and perhaps better option for placing the waypoints would be in the `NWNX_ON_CLIENT_CONNECT_AFTER` for each of that cdkey/player’s characters.

Plugins/Player/Player.cpp
```diff
@@ -1064,4 +1071,74 @@ ArgumentStack Player::GetQuestCompleted(ArgumentStack&& args)
...
+            auto pCreature = Utils::AsNWSCreature(Utils::GetGameObject(pPlayer->m_oidNWSObject));
+            pCreature->m_oidDesiredArea = pWP->m_oidArea;
+            pCreature->m_vDesiredAreaLocation = pWP->m_vPosition;
+            pCreature->m_bDesiredAreaUpdateComplete = false;
+            pCreature->m_vOrientation = pWP->m_vOrientation;
+            pPlayer->m_bFromTURD = true;
+            // We don't need the WP any more if we're only setting the location on first connect
+            if (bFirstConnectOnly)
+            {
+                pWP->RemoveFromArea();
```
Daztek: Mind, this doesn't actually destroy the WP, it just lives somewhere in "limbo". I wonder if you could do a `delete pWP;` after `RemoveFromArea()` without the game exploding.

Daztek:
Also, if you feel like it, you could add your examples from above to the plugin's `README.md`!

Energia00:
So I can't transfer a newly created character if it has certain conditions? I have a test area and whoever creates a character named "test" I transfer him to that area. Can I do this for new characters?

PR #565, opened 2019-09-08
Event: Surface Material Change

plenarius:
Sends an event whenever the surface material under a creature changes.

Caveat: `SetPosition` doesn't fire when first entering an area, only after the creature moves the slightest will it then fire. So if a creature transitions from a grass area directly into a snowy area it won't change until they take their first step.

### DIFF
--- a/Plugins/Events/Events/MaterialChangeEvents.cpp
+++ b/Plugins/Events/Events/MaterialChangeEvents.cpp
@@ -28,6 +28,8 @@ void MaterialChangeEvents::SetPositionHook(Services::Hooks::CallType type, API::
     if (thisPtr->m_nObjectType == API::Constants::ObjectType::Creature)
     {
         auto oArea = API::Globals::AppManager()->m_pServerExoApp->GetAreaByGameObjectID(thisPtr->m_oidArea);
+        if (oArea == nullptr)
+            return;
         auto iMat = oArea->GetSurfaceMaterial(vPos);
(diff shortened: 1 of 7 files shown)


Plugins/Events/Events/MaterialChangeEvents.cpp
```diff
@@ -0,0 +1,41 @@
...
+                API::CNWSObject*, API::Vector, int32_t>(&SetPositionHook);
+    });
+}
+
+void MaterialChangeEvents::SetPositionHook(Services::Hooks::CallType type, API::CNWSObject* thisPtr, API::Vector vPos, int32_t)
+{
+    const bool before = type == Services::Hooks::CallType::BEFORE_ORIGINAL;
+    if (thisPtr->m_nObjectType == API::Constants::ObjectType::Creature)
+    {
+        auto oArea = API::Globals::AppManager()->m_pServerExoApp->GetAreaByGameObjectID(thisPtr->m_oidArea);
```
Daztek: Add a `if (pArea) {}` to be safe

Plugins/Events/NWScript/nwnx_events.nss
```diff
@@ -791,6 +791,16 @@ const string NWNX_Events = "NWNX_Events";
         POS_Y                   float
         POS_Z                   float
         RUN_TO_POINT            int         TRUE if player is running, FALSE if player is walking (eg when shift clicking)
+////////////////////////////////////////////////////////////////////////////////
+    NWNX_ON_MATERIALCHANGE_BEFORE
+    NWNX_ON_MATERIALCHANGE_AFTER
+
```
Daztek: Maybe add the caveat here as a note

PR #579, opened 2019-09-28
Rename: Refactoring

### LINKED ISSUE #572: [Feature Request] Reset global overrides for NWNX_Rename
For the disguise system on our PW we would like to be able to remove player-specific overrides if the global override changes; i.e., a new global override would "reset" all player-specific overrides.

plenarius:
* Add `ClearPCNameOverride()`
* Change from POS to std containers
* Allow an additional player name override "ANONYMOUS" which can be defined with an ENV VAR
* Use less confusing vars in the nss
* Fix a couple issues with DMs
Resolves #572


plenarius:
Forgot to mention this resolves #572

PR #580, opened 2019-09-29
Minor and style issues

jd28:
Handful of more minor stuff.

### DIFF
--- a/NWNXLib/Services/PerObjectStorage/PerObjectStorage.cpp
+++ b/NWNXLib/Services/PerObjectStorage/PerObjectStorage.cpp
@@ -51,7 +51,7 @@ void PerObjectStorage::Set(API::CGameObject *pGameObject, const std::string& key
 {
     if (auto *pOS = GetObjectStorage(pGameObject))
     {
-        pOS->GetStringMap()[key] = value;
+        pOS->GetStringMap().emplace(key, std::move(value));
     }
 }
 void PerObjectStorage::Set(API::CGameObject *pGameObject, const std::string& key, void *value, CleanupFunc cleanup)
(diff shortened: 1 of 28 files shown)


NWNXLib/Services/Events/Events.cpp
```diff
@@ -138,9 +138,8 @@ void EventsProxy::ClearEvent(const std::string& eventName)
         throw std::runtime_error("Tried to clear unrecognised event.");
     }
 
-    Events::RegistrationToken concreteToken = std::move(*token);
+    m_proxyBase.ClearEvent(*token);
```
mtijanic: I have a vague recollection that these had to be done in the old order, but I don't remember why at all, and looking at the code now it doesn't seem like it. *shrug*

Did you run the unit test suite with these changes?

NWNXLib/Services/PerObjectStorage/PerObjectStorage.cpp
```diff
@@ -33,28 +33,28 @@ PerObjectStorage::ObjectStorage* PerObjectStorage::GetObjectStorage(API::Types::
...
     {
         pOS->GetFloatMap()[key] = value;
     }
 }
-void PerObjectStorage::Set(API::CGameObject *pGameObject, std::string key, std::string value)
+void PerObjectStorage::Set(API::CGameObject *pGameObject, const std::string& key, std::string value)
 {
     if (auto *pOS = GetObjectStorage(pGameObject))
     {
         pOS->GetStringMap()[key] = value;
```
mtijanic: Since you're already cleaning this up, might as well move the value in (`emplace()` method, IIRC)

NWNXLib/Services/Events/Events.cpp
jd28: I'll just drop this commit for now.  I really don't understand enough about the code base yet.

Daztek: I don't think `ClearEvent()` is actually used anywhere right now.

NWNXLib/Services/PerObjectStorage/PerObjectStorage.cpp
jd28: Will do!

PR #590, opened 2019-10-04
Area: add TestDirectLine

weavejester:
A function for testing whether there is a direct walkable line between two points in an area. Takes into account placeables and creatures as well as terrain.

My use-case for this is supporting @Daztek's "knockback" effect/hack that is stopped by placeables as well as terrain.

### DIFF
--- a/Plugins/Area/Area.cpp
+++ b/Plugins/Area/Area.cpp
@@ -73,6 +73,7 @@ Area::Area(const Plugin::CreateParams& params)
     REGISTER(CreateTransition);
     REGISTER(GetTileAnimationLoop);
     REGISTER(SetTileAnimationLoop);
+    REGISTER(TestDirectLine);
 
 #undef REGISTER
 }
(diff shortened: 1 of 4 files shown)


Plugins/Area/Area.cpp
```diff
@@ -723,4 +724,31 @@ ArgumentStack Area::SetTileAnimationLoop(ArgumentStack&& args)
...
+          ASSERT_OR_THROW(fStartY >= 0.0f);
+        const auto fEndX = Services::Events::ExtractArgument<float>(args);
+          ASSERT_OR_THROW(fEndX >= 0.0f);
+        const auto fEndY = Services::Events::ExtractArgument<float>(args);
+          ASSERT_OR_THROW(fEndY >= 0.0f);
+        const auto fPerSpace = Services::Events::ExtractArgument<float>(args);
+          ASSERT_OR_THROW(fPerSpace >= 0.0f);
+        const auto fHeight = Services::Events::ExtractArgument<float>(args);
+            ASSERT_OR_THROW(fHeight >= 0.0f);
+        const auto bIgnoreDoors = Services::Events::ExtractArgument<int32_t>(args);
```
Daztek: I'd do `const auto bIgnoreDoors = Services::Events::ExtractArgument<int32_t>(args) != 0;` here so it's always 1 or 0.

PR #593, opened 2019-10-06
nwnxlib/logging: A couple (maybe) optimizations

jd28:
Haven't really tested this, just an exploratory PR.  There are some other issues with the logging system viz. user-defined types have to be coerced to const char* long before the loglevel is determined.  So I'm sending this out for feedback, it's all minor stuff.

Change LogLevelMap to an adhoc flat map.  There's no point in potentially allocating and hashing a string for every log call when in production most plugins/core are at the default log level.  The advantage here is that in production the map would be (basically) empty.

Don't set log level if it's equal to default loglevel.

### DIFF
--- a/NWNXLib/Log.cpp
+++ b/NWNXLib/Log.cpp
@@ -5,8 +5,10 @@
 #include "API/Globals.hpp"
 #include "API/CExoBase.hpp"
 
+#include <string>
 #include <cstring>
-#include <unordered_map>
+#include <vector>
+#include <algorithm>
 
 namespace NWNXLib {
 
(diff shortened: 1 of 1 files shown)


NWNXLib/Log.cpp
```diff
@@ -72,21 +74,32 @@ void InternalTrace(Channel::Enum channel, Channel::Enum allowedChannel, const ch
...
-    auto entry = s_LogLevelMap.find(plugin);
+    auto entry = std::find_if(std::begin(s_LogLevelMap), std::end(s_LogLevelMap),
+        [plugin](const auto& pair) { return pair.first == plugin; });
     return entry == std::end(s_LogLevelMap) ? Channel::SEV_NOTICE : entry->second;
 }
 
 void SetLogLevel(const char* plugin, Channel::Enum logLevel)
 {
-    s_LogLevelMap[plugin] = logLevel;
+    // No point in adding plugin to log level map if it's requesting the default log level.
```
mtijanic: The default log level for every plugin is the core's default level (which, if not set, is notice). _Ideally_, we'd check against a variable rather than a constant, but it really doesn't matter.

NWNXLib/Log.cpp
```diff
@@ -72,21 +74,32 @@ void InternalTrace(Channel::Enum channel, Channel::Enum allowedChannel, const ch
...
+    if (logLevel != Channel::SEV_NOTICE) { return; }
```
mtijanic: style nit: return in new line, braces not necessary.

jd28:
the performance improvement would really only be when logging doesn't occur.  It is minor but I'm guessing just by a search there are a fair amount of LOG_DEBUGs in main code paths.  With the move to x64 short string optimization would make it pretty moot tho.

moving the log level check wouldn't fix stuff like this: https://github.com/nwnxee/unified/blob/4a087dcf4ea4f8c48f8e0e71feab39cd9ff0a716/NWNXLib/Services/Events/Events.inl#L37-L38

tinyformat would be a nice improvement and probably fix the above.  I'm assuming we could just write operator<< for these type of types.  if you're cool with that I'll work on it.

mtijanic:
> I'm assuming we could just write operator<< for these type of types

Yes, tinyformat will use `<<` when you want to %s print something, whatever it is, so to make a custom print for our custom type, we'd just override the operator.

I don't have a strong preference on the tinyformat lib, so if you have better solutions, I'm all ears. I do have a strong preference for printf-style API, rather than using ostream directly though.

jd28:
I'm closing this one and reopening new one.

PR #597, opened 2019-10-07
Fix spells known for Bard/Sorcs in MaxLevel

plenarius:
A rather large misunderstanding in the Spell Known logic for Bards/Sorcs has been corrected.

### DIFF
--- a/Plugins/MaxLevel/MaxLevel.cpp
+++ b/Plugins/MaxLevel/MaxLevel.cpp
@@ -210,21 +210,21 @@ void MaxLevel::LoadSpellGainTableHook(Services::Hooks::CallType type, CNWClass* 
     if (type == Services::Hooks::CallType::BEFORE_ORIGINAL)
         return;
 
-    auto *twoda = new C2DA(CResRef(pTable->CStr()), 1);
-    twoda->Load2DArray();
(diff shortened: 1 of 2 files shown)


Plugins/MaxLevel/MaxLevel.cpp
```diff
@@ -268,24 +269,60 @@ uint8_t MaxLevel::GetSpellGainHook(CNWClass *pClass, uint8_t nLevel, uint8_t nSp
...
 }
 
-// If the player is at level 40 or lower we get the spell gain for that class from the normal array,
+// After the server loads 1-40 we populate our map with the values for 41+
+void MaxLevel::LoadSpellKnownTableHook(Services::Hooks::CallType type, CNWClass* pClass, CExoString *pTable)
+{
+    if (type == Services::Hooks::CallType::BEFORE_ORIGINAL)
+        return;
+
+    auto *twoda = new C2DA(CResRef(pTable->CStr()), 1);
```
Daztek: You should just do `C2DA twoda(pTable->CStr(), true);` here, no need to allocate it on the heap, especially since you don't free it when you're done.

Daztek: Same for `LoadSpellGainTableHook()` which I guess I missed last time

PR #619, opened 2019-10-19
Utils add some string functions

jd28:


NWNXLib/Utils/String.hpp
```diff
@@ -0,0 +1,30 @@
+#pragma once
+
+#include <string>
+#include <vector>
+
+namespace NWNXLib {
+namespace Utils {
+
+// String converstions...
```
plenarius: Nit: typo

mtijanic:
Soeme of these conversion functions throw exceptions, which I'm not entirely a fan of. But at the very least, we should document that they do, because exceptions out of nwscript context (e.g. In hooks) are fatal.

Preferably (to me) , they'd return Maybe/std::optional. :)

jd28:
I'll fix the type and do the optional/Maybe.  It's a pity that [std::from_chars](https://en.cppreference.com/w/cpp/utility/from_chars) isn't more supported yet.  These functions could be a very thin veneer over that.

jd28:
I redid this with Maybe, std::optional wasn't available with the current docker g++.  The code would be cleaner if it was, since it has implicit constructors and Maybe does not.  Can't say I'm thrilled with it but genericity at this level can lead to some serious simplifications throughout the codebase.

I also fixed the typo.

jd28:
And the semantics of this could be argued still.  like should "123heythere" be valid conversion, i dunno.

jd28:
Changed this to allow converting numbers with surrounding whitespace.  Not the pretty is stuff in the world.
Added `bool trimmed = true` param to split so that "1, 2, 3" wouldn't result in ["1", " 2", " 3"]

Issue #459, opened 2019-04-16
NWNX:EE-64bit TODO


Daztek:
### CORE:
  - ~~Add an environment variable(default ON) to disable NWNX calls from ExecuteScriptChunk()~~

### PLUGINS:
- **Administration:**
  - ~~REMOVE: BootPCWithMessage()~~
  - ~~CHANGE: Have OnDeleteTURD() return TRUE on successful deletion~~
  - ~~CHANGE: Clean up the nwnx_admin.nss~~
- **Data:**
  - ~~CHANGE: PR #427~~
- **ELC:**
  - ~~CHANGE: Fix ELC :(~~
- **Effect:**
  - ~~REMOVE: EffectID from Pack/Unpack Functions/Struct~~
- **Events:**
  - INVESTIGATE: Event Name / Event Data Tag standardization.
  - ~~CHANGE: InventoryEvent Event Names, see PR #529~~
  - ~~CHANGE: Refactor `Events.cpp` to be more inline with the other plugins~~ 
- **ItemProperty:**
  - ~~REMOVE: EffectID from Pack/Unpack Functions/Struct~~
- **Mono**
  - INVESTIGATE: Replace Mono with .NET Core or add a separate plugin for it
- **Object:**
  - ~~REMOVE: {Get|Set}EventHandler(), {Get|Set}Portrait()~~
- **Player:**
  - ~~REMOVE: {Get|Set}VisibilityOverride()~~ 
  - ~~CHANGE: Remove the try/catch block around StartGuiTimingBar()'s type variable~~
- **SQL**
  - ~~CHANGE: SQLite currently uses the `.sqlite3` extension, this will conflict with the CampaignDB sqlite implementation. NWNX should use something else.~~
- **Weapon:**
  - INVESTIGATE: Move the Devastating Critical event to NWNX_Events

Baaleos:
Can I suggest that if we do get 64bit support - .Net Core would become a potential replacement for MONO.
This was suggested previously, but the hurdle at the time was that .Net Core was not 32bit capable.

mtijanic:
Yes. Whether it is a replacement or a separate plugin remains to be seen, but it .NET Core support is planned. Mono sucks terribly.

jd28:
[...]

PR #631, opened 2019-11-03
WIP: nwnx64

mtijanic:
NWNX 64 bit update for build 8192, big code drop

Unfortunately, this PR is effectively unreviewable, and we've had some issues with duplicated commits and other git history snafus. **We've decided to leave the messed up history in**. Sorry for anyone who goes spelunking.

Changes specific to nwnx64:

- cmake changes to build as 64bit
- Fixed all code assumptions that `sizeof(int)==sizeof(void*)` and similar
- API headers update
    - No longer using `NWNXLib::API` namespace
    - `CExoArrayList` (and others TBD) are using custom implementation now
- API .cpp files are replaced with `FunctionsLinux.cpp` which adds the implementation
    - This allows us to use virtual methods and a few other goodies.
- The `FunctionsLinux.hpp` constants have been changed to use mangled names
- All globals are now queried from nwserver instead of using hardcoded addresses
- Bea engine has been replaced with ~~ZyDis+Subhook~~ Funchook
- The `Patching` service has been removed. We no longer support inline assembly at the core level
     - individual plugins can do it if they want, but core makes no guarantees about cross-plugin functionality in that case

For other changes/features included in this branch, see #459, or CHANGELOG.md


jd28:
What a breath of fresh air this is.  I probably have been a jerk on discord, i apologize for that.  When I see this I think 'finally'. i proposed custom implementations of templated stuff like CExoArrayList to thes old maintainers like 4 years ago: https://gist.github.com/jd28/31fec93616f982fb9e57

PR #633, opened 2019-11-07
Purge JVM and Mono from the repo; remove -full docker builds.

mtijanic:
I think this does it. I also think I didn't break anything regarding the regular (i.e. non -full) docker. @jakkn or @urothis can you please confirm?


jakkn:
Yep, looks good to me.

urothis:
Agreed. lol

PR #648, opened 2019-11-09
services: Remove ServicesBase

jd28:
It seems to serve no purpose.  I can't imagine runtime polymorphism being very useful here.  I could be wrong tho.


mtijanic:
The best code is deleted code.

PR #650, opened 2019-11-09
Remove win32 CMake logic in CMakeLists

plenarius:


mtijanic:
Does anyone even use clang?

plenarius:
I don't believe anyone uses clang, no.

PR #665, opened 2019-11-17
NWNX_DotNET plugin

### LINKED ISSUE #634: [new plugin] .NET Core
Since we got rid of mono, and since we're at 64bit now, we can finally add a proper .NET core runtime.

(I'll be handling this, just wanted an issue here for a public record)

mtijanic:
Still WIP, but putting it up for feedback, primarily from @zunath .

Resolves #634


zunath:
This looks great - nice job!

There's a few pain points I've had to deal with. Not sure how much time/effort you want to put into this but here's some of my recommendations/nice to haves:

1.) The Object class in the NWN wrapper clashes with the .NET Object class. Recommend changing to NWGameObject or something similar to differentiate.

2.) The comments on the NWN functions aren't picked up by Visual Studio. They require some specific styling like the following:

```
        /// <summary>
        /// Cause the action subject to play an animation
        ///  - nAnimation: ANIMATION_*
        ///  - fSpeed: Speed of the animation
        ///  - fDurationSeconds: Duration of the animation (this is not used for Fire and
        ///    Forget animations)
        /// </summary>
        /// <param name="nAnimation"></param>
        /// <param name="fSpeed"></param>
        public static void ActionPlayAnimation(AnimationFireForget nAnimation, float fSpeed = 1.0f)
        {
            ActionPlayAnimation((int)nAnimation, fSpeed, 0.0f);
        }
```

If you auto-genned the wrappers, it might not be too difficult to get the comments formatted like this. Parameter information is a luxury, so long as the main comment info is picked up by auto-complete.

3.) Along the same lines, I've switched the constants over to proper C# enumerations and have found it cleans up the code a lot. Likely too much work for a general-purpose release but I thought I'd mention it. My example code is here: https://github.com/zunath/NWN.FinalFantasy/tree/master/NWN.FinalFantasy.Core/NWScript/Enumerations

4.) A question to make sure I'm understanding correctly: Are script events you set in the toolset firing the OnRunScript event or is it ONLY the ones you mark with SetEventScript? The comment "(e.g. via SetEventScript)" makes it sound like only the latter happens but the first comment line "This is called every time a named script is scheduled to run." seems to imply both. To reduce confusion I'd recommend specifying both methods as the examples given.

5.) Is oidSelf a necessary parameter on OnRunScript? Can we not use NWN.Object.OBJECT_SELF?

6.) How has your experience been with null references? In the mono plugin they would bring down the entire server, even if they were handled in a try/catch. Is that still the case here?

I'll try to do some testing on this tomorrow evening. Really appreciate it and looking forward to tinkering with it.

mtijanic:
> 1.) The Object class in the NWN wrapper clashes with the .NET Object class. Recommend changing to NWGameObject or something similar to differentiate.

Do you think we should be offering the whole class tree (NWObject->NWCreature, etc) here? Also, should we expose the property wrappers for nwscript calls, or should that be left to the user?

> 2.) The comments on the NWN functions aren't picked up by Visual Studio. They require some specific styling like the following:

I'll update it to include the `<summary>`. Adding params is a bit too much work right now, but if someone wants to pick it up later, great.
One thing that is missing here is some default values for Object types - I see nwnx_mono had the same issue - did you have any problems with that?

> 3.) Along the same lines, I've switched the constants over to proper C# enumerations and have found it cleans up the code a lot. Likely too much work for a general-purpose release but I thought I'd mention it. My example code is here: https://github.com/zunath/NWN.FinalFantasy/tree/master/NWN.FinalFantasy.Core/NWScript/Enumerations

I have plenty of those as well. Maybe we should include it. Related to 1).. I want to offer both a minimal lib that you need to get started, and a good premade base with convenience wrappers. I'm just not sure if the second one belongs in nwnxee/unified repo.

> 4.) A question to make sure I'm understanding correctly: Are script events you set in the toolset firing the OnRunScript event or is it ONLY the ones you mark with SetEventScript? The comment "(e.g. via SetEventScript)" makes it sound like only the latter happens but the first comment line "This is called every time a named script is scheduled to run." seems to imply both. To reduce confusion I'd recommend specifying both methods as the examples given.

Yes, both run. "Named scripts" means it doesn't trigger for `DelayCommand(..., () => {lambda})`. It also does not trigger for any events where the handler script is `""` - this is an explicit choice, and if that's a usecase anyone cares about, I can add a toggle.

> 5.) Is oidSelf a necessary parameter on OnRunScript? Can we not use NWN.Object.OBJECT_SELF?

Does it hurt any? It's always the same as `NWN.Object.OBJECT_SELF` (in `NWN.Internal.OnRunScript` we assign `OBJECT_SELF = oidSelf`), but it felt nicer to take it as an argument.

> 6.) How has your experience been with null references? In the mono plugin they would bring down the entire server, even if they were handled in a try/catch. Is that still the case here?

This PR will crash, but I have a fix for that so that try/catch work. Uncaught NullReferenceException will still crash the server.
Also, if you're running GDB, caught NRE will still generate a trap into the debugger. I'm not sure what is desirable here, and I'm not sure if we can control it even.

zunath:
> 
> Do you think we should be offering the whole class tree (NWObject->NWCreature, etc) here? Also, should we expose the property wrappers for nwscript calls, or should that be left to the user?

When I started SWLOR I went in the direction of "let's make this as OOP as possible" but found it made things harder to understand and new contributors got confused. So now I'm in favor of making it look like NWScript as close as possible. For that reason I'd recommend a general purpose release NOT including the full object hierarchy. I think it confuses too many people who may be new to C#.

An "advanced" optional version of the API with property wrappers, full class tree, and enumerations would be fine but I think that should be left up to the end user.
 
> I'll update it to include the `<summary>`. Adding params is a bit too much work right now, but if someone wants to pick it up later, great.

That's fine. The summary alone will save a lot of time.

> One thing that is missing here is some default values for Object types - I see nwnx_mono had the same issue - did you have any problems with that?

I don't recall specifically having any problems with them but if there's ways to improve it, then by all means do so.

> I have plenty of those as well. Maybe we should include it. Related to 1).. I want to offer both a minimal lib that you need to get started, and a good premade base with convenience wrappers. I'm just not sure if the second one belongs in nwnxee/unified repo.

Per our Discord convo I think I prefer having it in the nwnxee/unified repo simply because it's one central location for all things NWNX_DotNet. However you made a good point that stuff unrelated to NWNXEE would send notifications. I don't feel strongly either way, just a slight preference to the same repo.

> Yes, both run. "Named scripts" means it doesn't trigger for `DelayCommand(..., () => {lambda})`. It also does not trigger for any events where the handler script is `""` - this is an explicit choice, and if that's a usecase anyone cares about, I can add a toggle.

No that makes sense. The Lambda calls shouldn't execute that event.

> Does it hurt any? It's always the same as `NWN.Object.OBJECT_SELF` (in `NWN.Internal.OnRunScript` we assign `OBJECT_SELF = oidSelf`), but it felt nicer to take it as an argument.

Nah, doesn't matter in the grand scheme of things. Just was one small redundancy / point of confusion people might have. Just trying to preempt the "Do I need to pass oidSelf through all script calls?" question.

> This PR will crash, but I have a fix for that so that try/catch work. Uncaught NullReferenceException will still crash the server.
> Also, if you're running GDB, caught NRE will still generate a trap into the debugger. I'm not sure what is desirable here, and I'm not sure if we can control it even.

As long as we can try/catch I think I'm satisfied. Tracking down NREs was probably the biggest pain in Mono.

mtijanic:
Added `<summary>` to all nwscript functions, moved them over into a subfolder. Also, try/catch now works with NRE. This is nearing a plausible first release, once I type up the docs.
Got several perf improvements to do afterwards.

Plugins/DotNET/NWN/NWScript/Functions.cs
```diff
@@ -0,0 +1,11839 @@
...
+        ///    "AssignCommand failed."
+        ///    (If the object doesn't exist, nothing happens.)
+        ///  Delay aActionToDelay by fSeconds.
+        ///  * No return value, but if an error occurs, the log file will contain
+        ///    "DelayCommand failed.".
+        ///  It is suggested that functions which create effects should not be used
+        ///  as parameters to delayed actions.  Instead, the effect should be created in the
+        ///  script and then passed into the action.  For example:
+        ///  effect eDamage = EffectDamage(nDamage, DAMAGE_TYPE_MAGICAL);
+        ///  DelayCommand(fDelay, ApplyEffectToObject(DURATION_TYPE_INSTANT, eDamage, oTarget);
```
Daztek: Some misplaced text!

mtijanic:
Added readme, removed WIP tag. I'll do a little bit more cleanup tonight before submitting.

PR #671, opened 2019-11-21
Docker: update Dockerfile for 64bit

Daztek:
Adds the .NET SDK

~~It's failing to run, due to the following error:~~

~~```./nwserver-linux: /usr/lib/x86_64-linux-gnu/libstdc++.so.6: version `CXXABI_1.3.11' not found (required by /nwn/nwnx/NWNX_Core.so)```~~

~~I think because the beamdog/nwserver image is on debian-stretch, while we build nwnx on debian-buster.~~

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -10,10 +10,10 @@ COPY --from=builder /nwnx/home/Binaries/* /nwn/nwnx/
 
 # Install plugin run dependencies
 RUN runDeps="hunspell \
-    libmariadbclient18 \
+    libmariadb3 \
     libpq-dev \
     libsqlite3-dev \
-    libruby2.3 \
+    libruby2.5 \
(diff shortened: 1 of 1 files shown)


Daztek:
Niv kindly updated the beamdog/nwserver to debian-buster, so it no longer crashes.

* Had to update `libmariadbclient18` to `libmariadb3` since the former doesn't exist on buster, haven't tested if MySQL works.
* Also had to update `libruby2.3` to `libruby2.5`, no idea if that breaks anything.

Merging this will still not build our image, for that we'd have to change `FROM beamdog/nwserver` to `FROM beamdog/nwserver:8192`, do we want to do that?

Daztek:
NWNX_SQL with MySQL/MariaDB works. Hurray.

PR #672, opened 2019-11-22
Util: GetEnvironmentVariable: check for nullptr

Daztek:


Daztek:
It wasn't crashing, just throwing an error in the log when you query an unset env var. And since you made errors red they bother me :D 

```
Plugin 'NWNX_Util' failed event 'GetEnvironmentVariable'. Error: basic_string::_M_construct null not valid
Plugin 'NWNX_Util', event 'GetEnvironmentVariable': Tried to get a return value when one did not exist.```

mtijanic:
ACK.

PR #673, opened 2019-11-24
SQL: Cleanup

Daztek:


Plugins/SQL/NWScript/nwnx_sql.nss
```diff
@@ -100,7 +100,7 @@ int NWNX_SQL_GetPreparedQueryParamCount();
 
 int NWNX_SQL_PrepareQuery(string query)
 {
-    string sFunc = "PREPARE_QUERY";
+    string sFunc = "OnPrepareQuery";
```
mtijanic: Get rid of the `On` as well?

Daztek: Done.

mtijanic: Ship it!

PR #699, opened 2019-12-07
NWN.Object -> uint and OnClosure fix

nwnstuff:


Plugins/DotNET/NWN/Internal/Internal.cs
```diff
@@ -64,25 +65,25 @@ public static void OnClosure(ulong eid, uint oidSelf)
             Closures.Remove(eid);
```
mtijanic: It would be nice to clear it back to OBJECT_INVALID, so we catch these in the future.

Plugins/DotNET/Readme.md
```diff
@@ -100,7 +100,7 @@ This event runs whenever a named script - i.e. event scripts set in the toolset,
 
 If you do not want to handle this script, and let the regular nwscript do it (e.g. for `nw_`,`x0_` and `x2_` scripts), you should just return `-1` or `SCRIPT_NOT_HANDLED`. Otherwise, it's up to you to dispatch however you want, and return `>=0`. The return value is used in case of `StartingConditional` scripts.
 
-`oidSelf` is the handle of the object running the script. You can also access this object with `NWN.Object.OBJECT_SELF`.
+`oidSelf` is the handle of the object running the script. You can also access this object with `Internal.OBJECT_SELF`.
```
Liareth: This should probably be `NWN.OBJECT_SELF`. It seems wrong for code to have to touch something in the `Internal.` namespace to me.

mtijanic: I agree, but, is that possible? Don't you need `Namespace.Class.Field`? I guess `NWN.NWN.OBJECT_SELF` could work, and then you drop the first one since you're already in that NS?

Liareth: I would put it in the same class as all the scripting functions. So `NWN.NWScript.OBJECT_SELF` rather than `NWN.Internal.OBJECT_SELF`, if possible. The nwscript one can forward the internal one.

Liareth: `OBJECT_INVALID` will also need to be in the same place.

nwnstuff: Added `public static uint OBJECT_SELF => Internal.OBJECT_SELF;` to NWScript and updated docs. Much better place for it, NWScript methods still refer to `Internal.OBJECT_SELF` rather than just `OBJECT_SELF`, which should be fine.

Plugins/DotNET/NWN/Internal/Internal.cs
nwnstuff: Something like `finally { OBJECT_SELF = OBJECT_INVALID; }`?

mtijanic: `finally` is generally useless if the `catch` branch doesn't re-throw. I'd do it as:

```csharp
uint old = OBJECT_SELF;
OBJECT_SELF = oidSelf;
try {...} catch(...) {...}
Closures.Remove(eid);
OBJECT_SELF = old;
```

PR #703, opened 2019-12-08
Events stuff

jd28:
Few small Events related things.

### DIFF
--- a/NWNXLib/Services/Events/Events.inl
+++ b/NWNXLib/Services/Events/Events.inl
@@ -64,6 +64,12 @@ void Events::InsertArgument(ArgumentStack& stack, T&& arg)
     stack.emplace(std::forward<T>(arg));
 }
 
+template <typename... Args>
+void Events::InsertArguments(ArgumentStack& stack, Args&&... args)
+{
+    (InsertArgument(stack, std::forward<Args>(args)), ...);
+}
+
 template <typename T>
 T Events::ExtractArgument(ArgumentStack& arguments)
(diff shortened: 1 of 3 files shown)


NWNXLib/Services/Events/Events.inl
```diff
@@ -59,9 +59,15 @@ template<> std::optional<std::string>&          Events::Argument::Get<std::strin
...
+void Events::InsertArgument(ArgumentStack& stack, T&& arg)
 {
-    stack.push(Events::Argument(arg));
+    stack.emplace(std::forward<T>(arg));
+}
+
+template <typename... Args>
+void Events::InsertArguments(ArgumentStack& stack, Args&&... args)
+{
+    (InsertArgument(stack, std::forward<Args>(args)), ...);
```
mtijanic: I don't think I've ever seen this syntax. How does this even work?

Also, do we ever use this version of `InsertArgument`? Can we just remove it?

mtijanic: Ah, I see now you just added this. Looks reasonable. I still have no clue how this works and would like an explanation, but assuming it does, I'll merge it.

jd28: See `push_back_vec` example:
https://en.cppreference.com/w/cpp/language/fold it's new in C++17

I think everything uses InsertArgument since ArgumentStack is just an alias for std::stack.  Whether that's worth having or not, I doubt it.

PR #704, opened 2019-12-08
Update contributing.md

mtijanic:
Expanded a bit more and fixed some guidelines that just don't match reality.


mtijanic:
Cool. Merging, if anyone has any comments we can easily address them later.

PR #723, opened 2019-12-17
Core: check if the build revision is correct

Daztek:


mtijanic:
Add this also to `#define NWNX_EXPECT_VERSION(version)` and the two/three places that is used.

Daztek:
Done

PR #745, opened 2020-01-20
Adding GetSurfaceMaterial

Baaleos:
The new OnSurfaceMaterialChanged event is good - if you want to know when your character changes material - it does not however help if you are trying to plug it into point and click feat / use item property etc.
Adding GetSurfaceMaterial to areas plugin - it seemed appropriate.
Note - I still need to do some tests to verify this all builds and is functional - but in the mean time would appreciate feedback - as this is my first attempt at adding a nwnx contribution under the EE code base.


mtijanic:
This was added in base NWN nwscript with EE:
```C

// Get the surface material at the given location. (This is
// equivalent to the walkmesh type).
// Returns 0 if the location is invalid or has no surface type.
int GetSurfaceMaterial(location at);
```

Baaleos:
My bad-  closing 🗡 
I haven't explored all the EE functions yet.

mtijanic:
In terms of learning to add new functions, you did everything right. There's only style issues with your code: 
1) Make sure you set your editor to use spaces for indentation instead of tabs
2) Take a look at existing code if in doubt - e.g. we put `} else {` on three separate lines.

See here for more details: https://github.com/nwnxee/unified/blob/master/CONTRIBUTING.md

PR #748, opened 2020-01-22
Fix sneak attacks not working on flanked targets

### LINKED ISSUE #747: NWNX_TWEAKS_SNEAK_ATTACK_IGNORE_CRIT_IMMUNITY - No Flanking Sneak/Death attacks while enabled
Noticed since `NWN:EE build 8193.4` (never tested it on previous releases)

- Enabling `NWNX_TWEAKS_SNEAK_ATTACK_IGNORE_CRIT_IMMUNITY `makes both creatures and players unable to land Sneak/Death Attacks while flanking an opponent.
- Only flat-footed creatures/players can be subject to Sneak/Death Attacks while this Environment Variable is enabled.
- Disabling it restored the default situation and both creatures and players were able to Sneak/Death attack while flanking.

GoLoT:
Fixes #747

Can't test the changes at the moment because I need some sleep. I'll try to set up a testing scenario when I wake up, unless someone wants to give it a go before I do.

PS: Some changes are due to VS's auto-formatting and I only noticed them after pushing the changes. If it breaks consistency I can roll those changes back.

### DIFF
--- a/Plugins/Tweaks/Tweaks/SneakAttackCritImmunity.cpp
+++ b/Plugins/Tweaks/Tweaks/SneakAttackCritImmunity.cpp
@@ -225,43 +225,45 @@ void SneakAttackCritImmunity::CNWSCreature__ResolveDeathAttack_hook(CNWSCreature
         v.x -= pTarget->m_vPosition.x;
         v.y -= pTarget->m_vPosition.y;
         v.z -= pTarget->m_vPosition.z;
-        fDistance = v.x*v.x + v.y*v.y + v.z*v.z;
+        fDistance = v.x * v.x + v.y * v.y + v.z * v.z;
+        if (fDistance >= SNEAK_ATTACK_DISTANCE)
+            return;
     }
 
     bool isDeathAttack = false;
 
-    auto *pVisNode = pTarget->GetVisibleListElement(pThis->m_idSelf);
+    auto* pVisNode = pTarget->GetVisibleListElement(pThis->m_idSelf);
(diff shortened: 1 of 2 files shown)


Plugins/Tweaks/Tweaks/SneakAttackCritImmunity.cpp
```diff
@@ -223,58 +225,60 @@ void SneakAttackCritImmunity::CNWSCreature__ResolveDeathAttack_hook(CNWSCreature
...
-            CNWCCMessageData *pData = new CNWCCMessageData;
+            CNWCCMessageData* pData = new CNWCCMessageData;
             pData->SetObjectID(0, pTarget->m_idSelf);
             pData->SetInteger(0, 134);
             pAttackData->m_alstPendingFeedback.Add(pData);
         }
         else
         {
-            pAttackData->m_bDeathAttack = 1;
+            pAttackData->m_bSneakAttack = 1;
```
Daztek: Copy/paste error?

GoLoT: Yeah, thanks for catching it. Kinda pushed that commit quickly so I wouldn't forget about the death attack.

GoLoT:
Now that I had access to a second account (thanks to @mtijanic ) I was able to test the changes and for me it worked fine. If flat footed it allows sneak attacks unless the defender has uncanny dodge II and the attacker doesn't meet the level requirement (defender level + 4). Removing the untested tag.

PR #750, opened 2020-01-24
Events: add OnUUIDCollision to UUIDEvents

Daztek:


Plugins/Events/NWScript/nwnx_events.nss
```diff
@@ -893,6 +893,19 @@ _______________________________________
...
+    - NWNX_ON_UUID_COLLISION_BEFORE
+    - NWNX_ON_UUID_COLLISION_AFTER
+
+    `OBJECT_SELF` = The object that caused the UUID collision
+
+    Event Data Tag        | Type   | Notes
+    ----------------------|--------|-------
+    UUID                  | string | The UUID
+
+    Note: To get the existing object with `UUID` you can use GetObjectByUUID()
```
mtijanic: IIUC, this will fire before the object is actually added to the world, so many functions won't work (like `GetArea(OBJECT_SELF)`). If that's the case, would be nice to note.

Issue #715, opened 2019-12-11
Regarding C++ guidelines #4


jd28:
"4.  Never use non-const references as arguments. If a function will change the value of an object, pass it as a pointer."

I'm curious why this would be codified.  Personally, I think Config, for example (I have a pr gathering dust), would have been better off using a templated non-const reference parameter rather than template specializations on Maybe/std::optional return types.  Pointers could be used in that scenario too, but I don't see introducing a possible `null` as pure win.

jd28:
On second thought there would have been no need for templates in the inout parameter scenario, bog standard function overloading would have worked just fine.

mtijanic:
> I'm curious why this would be codified

The reason it is codified is the case of:
```C
int n = 3;
foo(n);
assert(n == 3);
```
When reading the code and come across `foo(n)`, there is no way to know whether it's passed by (non-const) reference or by value, and as such no way to know if n changes without checking the declaration of `foo`. This is extreme mental overhead, and I've had bugs with such code before. It applies to other types as well, objects or PODs.
When you see `foo(&n)` it is very obvious that it's going to change the value.

Overloads beat template functions for e.g. `Config::Set()`, but the `std::optional`-returning `Config::Get()` can't be overloaded, since it differs only in return type. I would not at all be opposed to changing `auto thingy = config->Get<std::string>("THINGY")` to `auto thingy = config->GetString("THINGY")`, but I didn't feel to strongly about it (and didn't think too hard about it) to include it in the guidelines.

jd28:
Fair enough.  I was thinking of this kind of use case (pointers could work here too of course) that's nicer (imo) than template specializations on return types:

```c++
bool GetConfig(const string& plugin, const string& key, int& inout); // etc... 
```

But it's really neither here nor there really,, it's not what my PR will be about.

jd28:
Oops, meant to close this.

jd28:
Re-opening this because there are examples in the codebase that do not follow this guideline.  Viz Events::InsertArgument(s).

mtijanic:
Yep. I'd actually prefer if we just changed all the code to be
```C
return Events::ArgumentStack{arg1, arg2, arg3};
```
just need to have a strong guarantee of left-to-right or right-to-left - whatever people feel more intuitive.

jd28:
i like it but -- and i am sliding seriously off topic -- but i don't really get the value of destroying the argstack storage every call/return.

mtijanic:
[...]

PR #755, opened 2020-01-26
Simplify ArgumentStack usage from plugins

mtijanic:
800 LOC less. Also paves the way to some optimizations. Related to #715

### DIFF
--- a/Plugins/Time/Time.cpp
+++ b/Plugins/Time/Time.cpp
@@ -55,58 +55,44 @@ Time::~Time()
 
 ArgumentStack Time::GetTimeStamp(ArgumentStack&&)
 {
-    ArgumentStack stack;
-
     auto now = std::chrono::system_clock::now();
     auto dur = now.time_since_epoch();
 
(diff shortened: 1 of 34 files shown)


Plugins/Time/Time.cpp
```diff
@@ -55,58 +55,44 @@ Time::~Time()
...
-
     auto now = std::chrono::system_clock::now();
     auto dur = now.time_since_epoch();
 
     auto count = std::chrono::duration_cast<std::chrono::microseconds>(dur).count();
 
-    Services::Events::InsertArgument(stack, (int32_t)(count / 1000000));
-    Services::Events::InsertArgument(stack, (int32_t)(count % 1000000));
-    return stack;
+    return Services::Events::Arguments((int32_t)(count % 1000000));
```
Daztek: This should return 2 values

`E [17:32:23] [NWNX_Core] Plugin 'NWNX_Time', event 'GetHighResTimeStamp': Tried to get a return value when one did not exist.` :(

mtijanic: Ouch. Thanky.

PR #756, opened 2020-01-26
ServerLogRedirector: fix timestamp trimming

### LINKED ISSUE #753: ServerLogRedirector discards chat [prefixes]
```
eval PrintString("[foo] aaa");
I [20:24:54] [NWNX_Core] Executing console command: 'eval' with args: PrintString("[foo] aaa");
I [20:24:54] [NWNX_ServerLogRedirector] (Server) aaa
```
https://github.com/nwnxee/unified/blob/master/Plugins/ServerLogRedirector/ServerLogRedirector.cpp#L52-L62

Code assumes people are using `WriteTimestampedLogEntry()` which adds a timestamp, but if you use PrintString and add a `[tag]` manually, it gets removed when printed to NWNX log.

Daztek:
Fixes #753

### DIFF
--- a/Plugins/ServerLogRedirector/ServerLogRedirector.cpp
+++ b/Plugins/ServerLogRedirector/ServerLogRedirector.cpp
@@ -1,6 +1,4 @@
 #include "ServerLogRedirector.hpp"
-#include "API/Version.hpp"
-#include "API/Functions.hpp"
 #include "API/CExoString.hpp"
 #include "Services/Config/Config.hpp"
 #include "Services/Hooks/Hooks.hpp"
@@ -34,6 +32,8 @@ using namespace NWNXLib;
 using namespace NWNXLib::API;
 using namespace NWNXLib::Services;
 
+static bool m_WriteTimestampedLogEntry;
+
 ServerLogRedirector::ServerLogRedirector(const Plugin::CreateParams& params)
     : Plugin(params)
 {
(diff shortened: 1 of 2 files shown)


Plugins/ServerLogRedirector/ServerLogRedirector.cpp
```diff
@@ -43,6 +43,9 @@ ServerLogRedirector::ServerLogRedirector(const Plugin::CreateParams& params)
 
     GetServices()->m_hooks->RequestSharedHook<Functions::_ZN17CExoDebugInternal16WriteToErrorFileERK10CExoString,
         void, CExoDebugInternal*, CExoString*>(&WriteToErrorFileHook);
+
+    GetServices()->m_hooks->RequestSharedHook<Functions::_ZN25CNWVirtualMachineCommands27ExecuteCommandPrintLogEntryEii,
+        int32_t>(&ExecuteCommandPrintLogEntry);
```
mtijanic: Could just do a tiny inline lambda :)

Plugins/ServerLogRedirector/ServerLogRedirector.cpp
```diff
@@ -34,6 +32,8 @@ using namespace NWNXLib;
 using namespace NWNXLib::API;
 using namespace NWNXLib::Services;
 
+static bool m_WriteTimestampedLogEntry;
```
mtijanic: Nitiest of nits: `m_` is used for member fields. Statics sometimes use `s_`, or just no prefix at all.

Plugins/ServerLogRedirector/ServerLogRedirector.cpp
Daztek: Good idea :D

PR #762, opened 2020-01-30
Optimizations: ridiculously low level GameObjectArray

mtijanic:


mtijanic:
Moved the optimizations out of experimental into a new plugin. I'll add a few more things to it.

PR #765, opened 2020-01-31
Tweak to revert cost of items with charges to 1.69 values

### LINKED ISSUE #761: Setting Item Charge Max to 50
Spoke with sherincall on the NWNX Discord, and he mentioned it might be possible for NWNX to set the item charge limit back to 50. This would be nice for servers relying on the old calculation of an item's value and/or for those relying on the value that is shown for these items as calculated in the toolset at the moment.

Example: In the toolset, an item with 15 charges that has Cast Spell: Gate (17) 5 Charges/Use shows as a value of 51638 which is an ILR of 16. But when created in-game it has a value of 3099 and an ILR of 4.

GoLoT:
Fixes #761 

I finally discarded the exclusive hook because it didn't make sense for this use case, maybe it could be useful to add cost calculations events to the events plugin or for some other uses but not for this.

It's working as expected but I don't know where the env var values should be documented. Any ideas?


Daztek:
> It's working as expected but I don't know where the env var values should be documented. Any ideas?

Somewhere in the readme would be best

GoLoT:
Done. Does that look OK?

mtijanic:
This is great, thanks!

PR #767, opened 2020-02-01
POS: use count instead of size() when serializing strings and floats

Daztek:
This fixes the following error:
`E [2020-01-30 20:07:20] [NWNXLib] [312] Serialized POS corrupted at '                    ', expected to parse '<%d>'. Aborting`


mtijanic:
Sanity restored.

PR #774, opened 2020-02-11
Adjust caster level when using ResistSpell() / DispelMagic() / GetCasterLevel()

GoLoT:
This tweak will increase the caster level based on the (Div|Arc)SpellLvlMod field of the character's multiclasses when applicable. This is done by checking classes.2da and using the Arcane, SpellCaster and (Div|Arc)SpellLvlMod to create a list of compatible classes and using that list to change the caster level when needed.

The formula used is `(level - 1) / spellLvlMod + 1` which results in adding 1 extra caster level for each `spellLvlMod` levels of the prestige class, same formula used by the engine to calculate the effective caster level for daily spells / spell slots.

This adjustment is only applied when `ResistSpell()`, `GetCasterLevel()`, `CNWSEffectListHandler::OnApplyDispelAllMagic()`, `CNWSEffectListHandler::OnApplyDispelBestMagic()` or `CGameEffect::SetCreator()` are called so it shouldn't affect other functions negatively and it should work around the issue of custom prestige classes not affecting spell penetration / dispel checks.

I want to add support for a custom nwscript callback to allow more precise control over the caster level on a per-module basis (i.e. increasing the effective caster level with feats)  but I don't know what the best approach would be. I will figure it out eventually. For now the PR can be merged if there aren't any issues with it.


Daztek:
Any more than a toggle moves it out of "Tweak" territory imo, and would probably be better off in its own plugin.

GoLoT:
Well, I guess it could be left as it is without the nwscript customization and if someone has a need for the callback (which I don't think is needed but I wanted to add it for completeness) I could maybe create a new plugin with that feature. Or add it to a different plugin.

Plugins/Tweaks/Tweaks/AdjustCasterLevel.cpp
```diff
@@ -0,0 +1,175 @@
...
+    hooker->RequestSharedHook<Functions::_ZN11CGameEffect10SetCreatorEj, void>
+        (&CGameEffect__SetCreator);
+}
+
+void AdjustCasterLevel::LoadClasses()
+{
+    s_classCasterType.clear();
+    s_divModClasses.clear();
+    s_arcModClasses.clear();
+    auto p2DA = Globals::Rules()->m_p2DArrays->GetCached2DA("classes", 1);
```
mtijanic: Use true/false instead of 1/0

Plugins/Tweaks/Tweaks/AdjustCasterLevel.cpp
```diff
@@ -0,0 +1,175 @@
...
+    auto spellCasterCol = p2DA->GetColumnIndex(CExoString("SpellCaster"));
+    auto arcaneCol = p2DA->GetColumnIndex(CExoString("Arcane"));
+    auto arcaneModCol = p2DA->GetColumnIndex(CExoString("ArcSpellLvlMod"));
+    auto divineModCol = p2DA->GetColumnIndex(CExoString("DivSpellLvlMod"));
+    for (int i = 0; i < p2DA->m_nNumRows; i++)
+    {
+        int spellCaster, arcane;
+        if (p2DA->GetINTEntry(i, spellCasterCol, &spellCaster) && spellCaster && p2DA->GetINTEntry(i, arcaneCol, &arcane))
+        {
+            s_classCasterType.push_back(!arcane + 1);
```
mtijanic: Use constants instead of 0/1/2.

Plugins/Tweaks/Tweaks/AdjustCasterLevel.cpp
```diff
@@ -0,0 +1,175 @@
...
+#include "API/C2DA.hpp"
+
+
+namespace Tweaks {
+
+using namespace NWNXLib;
+using namespace NWNXLib::API;
+
+std::vector<uint8_t> AdjustCasterLevel::s_classCasterType;
+std::vector<std::tuple<uint8_t, uint8_t>> AdjustCasterLevel::s_divModClasses;
```
mtijanic: This can just be a `std::vector<uint8_t>` for the mods, and avoid the extra logic of tuples.
Or even just `uint8_t[Constants::MAX_CLASSES]`. Same for other vectors.

GoLoT:
I will put this PR on hold for now until I find the best way to implement the suggestions, possibly moving it entirely to NWNX_Creature and adding more customization.

I'll leave it open in case someone wants to add some feedback.

GoLoT:
Quick update: Moved from Tweaks to Creature plugin. Added two nwscript functions to set a caster level override or a caster level modifier for a creature.

GoLoT:
I set up a simple test scenario using a 5 Wizard / 10 Pale Master character. It added the (10/2) levels and my custom +5 caster level properly for a total caster level of 15. It seemed to work for AoE effects too so the basic functionality seems to be working.

Not having access to a live server, setting up and testing more scenarios would be too time consuming so I'll let server owners do their own testing and report any issues they find with the plugin.

The code is ready for review and merge if everything is right.

TotallyNotLoinne:
If this comes to the live, I can test it, as some of my future work can and probably will heavily rely on this plugin.
Can the auto-2DA method be used with the conjunction of manual nwscript overriding? As I see it in the code, it calculates ManualOverride->ManualModifier->2daModifier, this order?

GoLoT:
Yes. It will use the override value and if it doesn't exist it will use the default value (class level) + manually set modifier (if any) + `ceil(prestigeClassLevel/(Arc|Div)SpellLvlMod)` if the 2DA option is enabled.

The level override can be removed by setting it to any value < 1 and the modifier is removed by setting it to 0.

mtijanic:
I'm not a fan of the organization here. It doesn't belong in tweaks, but also it doesn't belong in creature. And the additional config toggles for creature are weird.

I think that you should get rid of the base toggle: Either hook the functions the first time anyone calls any of the 4 functions form nwscript, or just hook them on load (I prefer if we adopted the first approach).

The 2DA stuff could be a tweak, as it is independent of the rest. It's a bit messy with the exclusive hook, but can be done with the messagebus or callbacks. Or possibly just shared hooks and temporarily modifying the level.

GoLoT:
Moved the 2DA logic to Tweaks as requested and added a fix for existing spell effects losing their caster level when the effect creator is removed/ceases to exist.

I repeated my previous tests and the plugin is working for those specific cases. The level override feature depends on the order in which the shared hooks are called, I don't see a way to ensure that the creature plugin is the last one to modify `m_nLevel` without using events/messages but that feels a bit overkill. I can add it if needed though.

Also I can't completely trust the new shared hook logic. I tried to make it as solid as possible but I expect problems in corner cases (modifiers being removed in a different order than they were applied and causing an under/overflow of the `m_nLevel` variable). Saving the previous value on `before` and restoring it on `!before` feels more robust but it also requires the after calls to be done in reverse order to the before calls. I'm hoping that an under/overflow in those cases will be reverted when the other modifier is removed but the `uint32_t -> uint8_t` implicit casts may not work the way I expect.

Plugins/Tweaks/Tweaks/AddPrestigeclassCasterLevels.cpp
```diff
@@ -0,0 +1,137 @@
...
+
+AddPrestigeclassCasterLevels::AddPrestigeclassCasterLevels(Services::HooksProxy* hooker)
+{
+    hooker->RequestSharedHook<Functions::_ZN17CNWSCreatureStats13GetClassLevelEhi, uint8_t>(&CNWSCreatureStats__GetClassLevel);
+    hooker->RequestSharedHook<Functions::_ZN25CNWVirtualMachineCommands28ExecuteCommandGetCasterLevelEii, int32_t>(&CNWVirtualMachineCommands__ExecuteCommandGetCasterLevel);
+    hooker->RequestSharedHook<Functions::_ZN25CNWVirtualMachineCommands25ExecuteCommandResistSpellEii, int32_t>(&CNWVirtualMachineCommands__ExecuteCommandResistSpell);
+    hooker->RequestSharedHook<Functions::_ZN11CGameEffect10SetCreatorEj, void>(&CGameEffect__SetCreator);
+    hooker->RequestSharedHook<Functions::_ZN8CNWRules13LoadClassInfoEv, void>(&CNWRules__LoadClassInfo);
+
+    for (int i = 0; i <= NWNXLib::API::Constants::ClassType::MAX; i++)
```
mtijanic, line 35: Just FYI, but any **global** variable that is not initialized explicitly is always automatically initialized to zeros. Doesn't really hurt tho.

GoLoT: I wasn't sure about arrays, I'll keep it in mind. Want me to remove it?

mtijanic, line 35: Also, a standard way to get the number of elements in an  array would be:
```
for (int i = 0; i < sizeof(s_classCasterType)/sizeof(s_classCasterType[0]); i++
```

this will work if you modify the size above, but forget to modify the loop.

Plugins/Tweaks/Tweaks/FixDispelEffectLevels.cpp
```diff
@@ -0,0 +1,166 @@
...
+{
+    hooker->RequestExclusiveHook<Functions::_ZN21CNWSEffectListHandler21OnApplyDispelAllMagicEP10CNWSObjectP11CGameEffecti>
+        (&CNWSEffectListHandler__OnApplyDispelAllMagic);
+    hooker->RequestExclusiveHook<Functions::_ZN21CNWSEffectListHandler22OnApplyDispelBestMagicEP10CNWSObjectP11CGameEffecti>
+        (&CNWSEffectListHandler__OnApplyDispelBestMagic);
+}
+
+int32_t FixDispelEffectLevels::CNWSEffectListHandler__OnApplyDispelAllMagic(CNWSEffectListHandler* thisPtr, CNWSObject* pObject, CGameEffect* pEffect, BOOL bLoadingGame)
+{
+    (void)thisPtr;
```
mtijanic, line 34: You can skip the explicit name in the arg list, and then don't need this.

Plugins/Tweaks/Tweaks/FixDispelEffectLevels.cpp
```diff
@@ -0,0 +1,166 @@
...
+    if (nDispelLevel < 1)
+    {
+        if (auto* pCreature = Utils::AsNWSCreature(pCreatorObject))
+            nDispelLevel = pCreature->m_pStats->GetLevel(0);
+        else
+            return 1;
+    }
+
+    std::vector<uint64_t> nIgnoreEffectIDs;
+    auto* pMessageData = new CNWCCMessageData();
```
mtijanic, line 48: Don't use `new` for objects unless NWN has to take ownership of it. Either `unique_ptr<CNWCCMessageData>`, or allocate it on the stack directly. That way you can't forget to do a delete. Stack alloc is also marginally faster.

Issue #776, opened 2020-02-13
Possible bug / Issue with Events Stealth hook


Baaleos:
Scenario: I want to handle stealth in my module, when a Vampire or god goes into stealth mode - the enter stealth event is skipped, and instead they receive greater sanctuary and nwnx visibility kicks in to make them invisible to players.

This worked well on 1.69 - the logic I had there was a single method that took the player, and int (boolean) to indicate whether stealth was turning on or off.

Code kinda looked like this (updated to be in .Net Core on EE)

```
public static void ToggleGodInvisibility(uint oGod, bool enabled)
        {

            if (!enabled )
            {
                Effect e = NWScript.GetFirstEffect(oGod);
                while (NWScript.GetIsEffectValid(e) == 1)
                {
                    if (NWScript.GetEffectType(e) == NWScript.EFFECT_TYPE_SANCTUARY || NWScript.GetEffectType(e) == NWScript.EFFECT_TYPE_ETHEREAL)
                    {
                        NWScript.RemoveEffect(oGod, e);
                    }
                    e = NWScript.GetNextEffect(oGod);
                }
                NWNX.Visibility.NWNX_Visibility_SetVisibilityOverride(oGod, NWScript.OBJECT_INVALID, -1);
                NWNX.Visibility.NWNX_Visibility_SetVisibilityOverride(oGod, oGod, -1);
                NWScript.SendMessageToPC(oGod, "You are now visible to all");
            }
            else
            {
                Effect eVis = NWScript.EffectVisualEffect(2155);
                Effect eSanc = NWScript.EffectSanctuary(255);
                Effect eEth = NWScript.EffectEthereal();
                Effect eLink = NWScript.EffectLinkEffects(eVis, eSanc);
                Effect eVsi2 = NWScript.EffectVisualEffect(1616);
                Vector pos = NWScript.GetPosition(oGod);
                pos.z = pos.z + 1.20f;
                Location l = NWScript.Location(NWScript.GetArea(oGod), pos, NWScript.GetFacing(oGod));
                NWScript.ApplyEffectAtLocation(NWScript.DURATION_TYPE_INSTANT, eVsi2, l);
                NWScript.ApplyEffectToObject(NWScript.DURATION_TYPE_PERMANENT, eLink, oGod);
                NWScript.ApplyEffectToObject(NWScript.DURATION_TYPE_PERMANENT, eEth, oGod);
[...]

PR #778, opened 2020-02-13
Adding suggested improvement from GoLot - to handle exit event firing…

Baaleos:
… when skip is called when not desired.
Discussed in detail here:
https://github.com/nwnxee/unified/issues/776

Essentially - skipping the onEnter event, was calling the onExit event - when not necessarily desired.
This could cause recursive loops if both events skipped.


GoLoT:
I think you mixed the fix with your own project, leading to your personal code being added to the PR. You need to create a new branch for each PR and only push changes related to the PR to that branch.

Baaleos:
Ah crap - forgot to change branch- will fix it soon

PR #802, opened 2020-02-29
Events: add UnsubscribeEvent()

Daztek:


Plugins/Events/Events.cpp
```diff
@@ -256,6 +257,27 @@ ArgumentStack Events::SubscribeEvent(ArgumentStack&& args)
...
+      ASSERT_OR_THROW(!event.empty());
+    const auto script = Services::Events::ExtractArgument<std::string>(args);
+      ASSERT_OR_THROW(!script.empty());
+
+    auto& eventVector = m_eventMap[event];
+    auto it = std::find(std::begin(eventVector), std::end(eventVector), script);
+
+    if (it == std::end(eventVector))
+    {
+        throw std::runtime_error("Attempted to unsubscribe from an event with a script that is not subscribed!");
```
mtijanic: (Belated, but can be done in followup)
In general for cases like these I think it's better to just do a LOG_NOTICE (or LOG_WARNING at most) and not throw the exception. That could also print both arguments given to catch typos.

Daztek: Yeah, I just copied from SubscribeEvent(), I'll fix them both soonish.

PR #805, opened 2020-03-01
optimizations: Fix GOA optimization messing with NWNX functionality

### LINKED ISSUE #801: NWNX_OPTIMIZATIONS_GAME_OBJECT_LOOKUP breaks NWNX_Util_GetLastCreatedObject() and NWNX_ON_DM_SPAWN_*
Both `GetLastCreatedObject()` and `NWNX_ON_DM_SPAWN_*` use `GetObjectArray()->m_nNextObjectArrayID[0]` to get the next ObjectID, since the `GameObjectLookup` optimization has its own implementation of `m_nNextObjectArrayID[]` these functions do not work if the optimization is enabled.

mtijanic:
Fixes #801  (well, I only tested the nwnx_utils part of it because that was possible from console, but should fix the other one)

### DIFF
--- a/Plugins/Optimizations/Optimizations/GameObjectLookup.cpp
+++ b/Plugins/Optimizations/Optimizations/GameObjectLookup.cpp
@@ -313,7 +313,7 @@ uint8_t GameObjectLookup::GetGameObject(void*, uint32_t id, CGameObject** ptr)
 
 void GameObjectLookup::SyncWithGameGOA()
 {
-    if (auto pGameObjectArray = Globals::AppManager()->m_pServerExoApp->GetObjectArray())
+    if (auto* pGameObjectArray = Globals::AppManager()->m_pServerExoApp->GetObjectArray())
     {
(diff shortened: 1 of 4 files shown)


Plugins/Optimizations/Optimizations/GameObjectLookup.cpp
```diff
@@ -304,4 +311,15 @@ uint8_t GameObjectLookup::GetGameObject(void*, uint32_t id, CGameObject** ptr)
     return BadId;
 }
 
+void GameObjectLookup::SyncWithGameGOA()
+{
+    if (auto pGameObjectArray = Globals::AppManager()->m_pServerExoApp->GetObjectArray())
```
Daztek: Nitpick: missing the `*` :P

mtijanic: I copied it from your code. Fixed in both places :)

PR #809, opened 2020-03-03
Events: add ResourceEvents

Daztek:
Seems to work okay. I don't know if there's a better place to hook than the main loop, I have it poll every second but if there's a better place to hook that runs semi frequently I'm open to suggestions.
```
D [11:11:48] File modified: es_s_travel.nss
D [11:11:48] Pushing event data: 'RESREF' -> 'es_s_travel'.
D [11:11:48] Pushing event data: 'TYPE' -> '2009'.
D [11:11:48] Dispatching notification for event 'NWNX_ON_RESOURCE_MODIFIED' to script 'es_s_subsysman'.
D [11:11:48] Getting event data: 'RESREF' -> 'es_s_travel'.
D [11:11:48] Getting event data: 'TYPE' -> '2009'.
I [11:11:48] (Server) ResRef: 'es_s_travel' -> 2009
```

### DIFF
--- a/Plugins/Events/Events/ResourceEvents.cpp
+++ b/Plugins/Events/Events/ResourceEvents.cpp
@@ -1,0 +1,161 @@
+#include "Events/ResourceEvents.hpp"
+#include "API/CExoBase.hpp"
+#include "API/CExoAliasList.hpp"
+#include "API/CNWSModule.hpp"
+#include "API/CExoResMan.hpp"
+#include "API/Functions.hpp"
+#include "API/Constants.hpp"
+#include "Events.hpp"
+#include "Utils.hpp"
+
+#include <sys/inotify.h>
+#include <unistd.h>
+#include <poll.h>
+#include <climits>
+#include <fcntl.h>
+
+namespace Core {
+    extern bool g_CoreShuttingDown;
+}
+
+namespace Events {
+
+using namespace NWNXLib;
+using namespace NWNXLib::API;
+using namespace NWNXLib::API::Constants;
+
+#define EVENT_SIZE          ( sizeof (struct inotify_event) )
+#define EVENT_BUF_LEN       ( 128 * ( EVENT_SIZE + NAME_MAX + 1) )
+
+static bool processEvents;
+static int selfPipeFds[2];
+std::unique_ptr<std::thread> ResourceEvents::m_pollThread;
+
+ResourceEvents::ResourceEvents(Services::TasksProxy* tasks)
+{
+    Events::InitOnFirstSubscribe("NWNX_ON_RESOURCE_.*", [tasks]()
+    {
+        int ret = pipe2(selfPipeFds, O_NONBLOCK);
+
+        if (ret == -1)
+        {
+            LOG_ERROR("Failed to initialize self-pipe with error: %s", std::strerror(errno));
+            return;
+        }
+
+        int pipeReadFd = selfPipeFds[0];
+        CExoString exoPath = Globals::ExoBase()->m_pcExoAliasList->GetAliasPath("NWNX");
+        std::string path = exoPath.CStr();
+
+        m_pollThread = std::make_unique<std::thread>([tasks, path, pipeReadFd]()
+        {
+            struct pollfd fds[2];
+            char buffer[EVENT_BUF_LEN];
+
+            fds[0].fd = pipeReadFd;
+            fds[0].events = POLLIN;
+
+            int fd = inotify_init1(IN_NONBLOCK), wd = 0;
+
+            if (fd != -1)
+            {
+                fds[1].fd = fd;
+                fds[1].events = POLLIN;
+
+                wd = inotify_add_watch(fd, path.c_str(), IN_CREATE | IN_DELETE | IN_MODIFY);
+
+                if (wd != -1)
+                {
+                    processEvents = true;
+                }
+            }
+
+            auto ProcessEvent = [tasks](const std::string &eventType, const std::string &file) -> void
+            {
+                tasks->QueueOnMainThread([eventType, file]()
+                {
+                    if (Core::g_CoreShuttingDown)
+                        return;
+
+                    LOG_DEBUG("EventType: %s, File: %s", eventType, file);
+
+                    uint16_t resType = Globals::ExoResMan()->GetResTypeFromFile(file);
+
+                    if (resType != 65535)
+                    {
+                        CResRef resRef;
+                        Globals::ExoResMan()->GetResRefFromFile(resRef, file);
+
+                        Events::PushEventData("RESREF", resRef.GetResRefStr());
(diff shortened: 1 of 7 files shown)


mtijanic:
You shouldn't be polling in the first place. Fork off a new thread that will do a `select()` on the inotify fd, which will block it until there's something available. At which point, you QueueOnMainThread() the event you want to run.

Daztek:
Can't say I'm not learning stuff. Moved everything to its own thread and poll now blocks until something is available and then queues it on the main thread.

Plugins/Events/Events/ResourceEvents.cpp
```diff
@@ -0,0 +1,161 @@
...
+    extern bool g_CoreShuttingDown;
+}
+
+namespace Events {
+
+using namespace NWNXLib;
+using namespace NWNXLib::API;
+using namespace NWNXLib::API::Constants;
+
+#define EVENT_SIZE          ( sizeof (struct inotify_event) )
```
mtijanic: `constexpr` instead of `#define` whenever possible. Where `constexpr` won't work (will here), try `const`. `#define` should be reserved for generating code only (such as the REGISTER() macro).

Plugins/Events/Events/ResourceEvents.cpp
```diff
@@ -0,0 +1,161 @@
...
+#define EVENT_BUF_LEN       ( 128 * ( EVENT_SIZE + NAME_MAX + 1) )
+
+static bool processEvents;
```
mtijanic: This needs to be marked as `volatile`. Otherwise, the side thread might load the value in a register and keep checking the register instead of loading from memory every time.

Plugins/Events/Events/ResourceEvents.cpp
```diff
@@ -0,0 +1,161 @@
...
+        int ret = pipe2(selfPipeFds, O_NONBLOCK);
+
+        if (ret == -1)
+        {
+            LOG_ERROR("Failed to initialize self-pipe with error: %s", std::strerror(errno));
+            return;
+        }
+
+        int pipeReadFd = selfPipeFds[0];
+        CExoString exoPath = Globals::ExoBase()->m_pcExoAliasList->GetAliasPath("NWNX");
```
mtijanic: Should it also watch DEVELOPMENT:?

Plugins/Events/Events/ResourceEvents.cpp
```diff
@@ -0,0 +1,161 @@
...
+                            {
+                                auto *event = reinterpret_cast<inotify_event *>(&buffer[i]);
+
+                                if (event->len)
+                                {
+                                    if (event->mask & IN_MODIFY)
+                                    {
+                                        ProcessEvent("MODIFIED", event->name);
+                                    }
+                                    else
```
mtijanic: `else if` same line please.

Plugins/Events/Events/ResourceEvents.cpp
```diff
@@ -0,0 +1,161 @@
...
+
+                if (pollNum > 0)
+                {
+                    if (fds[1].revents & POLLIN)
+                    {
+                        size_t length = read(fd, buffer, EVENT_BUF_LEN);
+
+                        if (length > 0)
+                        {
+                            for (size_t i = 0; i < length;)
```
mtijanic: Personal nit, but for loops with a missing part are very confusing I'd just do:
```C
size_t i = 0;
while (i < length) {...}
```

or even:
```C
while (length > 0) {
  ...
  length -= (EVENT_SIZE + event->len);
}
```

Plugins/Events/Events/ResourceEvents.cpp
```diff
@@ -0,0 +1,161 @@
...
+
+                if (wd != -1)
+                {
+                    processEvents = true;
+                }
+            }
+
+            auto ProcessEvent = [tasks](const std::string &eventType, const std::string &file) -> void
+            {
+                tasks->QueueOnMainThread([eventType, file]()
```
mtijanic: Fun fact: NWN can take up to a second to actually register the resource after it's been copied in (due to lack of cross-platform inotify API). So it's very likely that on any event here you will actually see the stale data if you attempt to read the resource.

The easy way to solve this would be to just delay queuing here by 1 second. A better way might be to push all these events into a buffer, and also hook `CExoKeyTable::CheckForModificationsInStorage`. That thing runs every second when NWN checks the folders. After it has run, you can queue the event on the main thread and it'll guarantee the resources are visible.

Plugins/Events/Events/ResourceEvents.cpp
```diff
@@ -0,0 +1,161 @@
...
+    });
+}
+
+ResourceEvents::~ResourceEvents()
+{
+    processEvents = false;
+
+    if (m_pollThread)
+    {
+        int ret = write(selfPipeFds[1], "a", 1);
```
mtijanic: You can just do `(void)write(...);`

Plugins/Events/Events/ResourceEvents.cpp
Daztek: Shouldn't be too hard to add, personally I only use NWNX: for my stuff. If you think it's useful I can add a watch for it.

Plugins/Events/Events/ResourceEvents.cpp
Daztek: I tried that at first but my compiler still complained. Maybe it's just extra fussy.

Plugins/Events/Events/ResourceEvents.cpp
Daztek: Added DEVELOPMENT: after all!

Plugins/Events/Events/ResourceEvents.cpp
Daztek: I did the easy way and just sleep the thread for a second, how would the better way best be done?

mtijanic: > how would the better way best be done?

1) Define an event structure that contains `enum { ADDED, REMOVED, MODIFIED} type; string resource;`. Define a global `vector<eventstruct>` that stores all events.
2) shared-hook `CheckForModificationInStorage`. If your global vector is not empty, `QueueOnMainThread` a function that signals the event, once for every vector element, and then clears the vector.

This is still a bit racey, so _ideally_ there is a binary semaphore that your shared hook acquires when it runs the QueueOnMainThread, and the main thread lambda releases it when all events are signaled. Since that's a hassle, it's enough to just use a `std::mutex` and acquire it in the main thread lambda when iterating over the array, as well as the side thread when inserting into it.

All that said, this works well enough for me, so if you just wanna merge, go ahead.

Daztek: I'll give the shared hook a try sometime and do a followup PR if that works well

PR #810, opened 2020-03-03
Player: Journal funcs

### LINKED ISSUE #807: AddCustomJournalEntry  - to player plugin
Just placeholder issue to link against

Currently building method to add custom journal entries to a player via NSS / C#
Without having to use Custom Tokens or modify the JRL file.

Baaleos:
Journal modification functionality
As demonstrated here:
https://youtu.be/ZeR-9S6WAwc

resolves #807


Plugins/Player/Player.cpp
```diff
@@ -1336,4 +1344,171 @@ ArgumentStack Player::SetResManOverride(ArgumentStack&& args)
     return Services::Events::Arguments();
 }
 
+
+CExoLocString CreateCExoLocString(const std::string& str)
+{
+    CExoLocString locStr;
+    locStr.AddString(0,CExoString(str.c_str()),0);
+    return locStr;
+}
```
Daztek: This should go in NWNXLib/Utils.hpp/cpp as the following:

```c
CExoLocString CreateLocString(const std::string& str, int32_t nID = 0, uint8_t bGender = 0);

CExoLocString CreateLocString(const std::string& str, int32_t nID, uint8_t bGender)
{
    CExoLocString locStr;
    locStr.AddString(nID, CExoString(str.c_str()), bGender);
    
    return locStr;
}
```

Plugins/Player/Player.cpp
```diff
@@ -1336,4 +1344,171 @@ ArgumentStack Player::SetResManOverride(ArgumentStack&& args)
...
+            auto entries = pCreature->m_pJournal->m_lstEntries;
+            
+            const auto questName = Services::Events::ExtractArgument<std::string>(args);
+            const auto questText = Services::Events::ExtractArgument<std::string>(args);
+            const auto tag = Services::Events::ExtractArgument<std::string>(args);
+            const auto state = Services::Events::ExtractArgument<int32_t>(args);
+            const auto priority = Services::Events::ExtractArgument<int32_t>(args);
+            const auto completed = Services::Events::ExtractArgument<int32_t>(args);
+            const auto diplayed = Services::Events::ExtractArgument<int32_t>(args);
+            const auto updated = Services::Events::ExtractArgument<int32_t>(args);
```
Daztek: Some `ASSERT_OR_THROW()`s would be nice here. Especially for the `SJournalEntry` stuff that takes an unsigned value.

Also have you tested what happens when `questName`, `questText` or `tag` are empty strings?

Plugins/Player/Player.cpp
```diff
@@ -1336,4 +1344,171 @@ ArgumentStack Player::SetResManOverride(ArgumentStack&& args)
...
+            auto silentUpdate = Services::Events::ExtractArgument<int32_t>(args);
+            
+            //If server owner leaves this 0 - the entry will be added with todays date
+            if (calDay == 0)
+            {
+                uint32_t year = Globals::AppManager()->m_pServerExoApp->GetWorldTimer()->GetWorldTimeYear();
+                uint32_t month = Globals::AppManager()->m_pServerExoApp->GetWorldTimer()->GetWorldTimeMonth();
+                uint32_t day = Globals::AppManager()->m_pServerExoApp->GetWorldTimer()->GetWorldTimeDay();
+                calDay = Globals::AppManager()->m_pServerExoApp->GetWorldTimer()->ConvertToCalendarDay(year, month, day);
+            }
```
Daztek: I think you can just do `calDay = Globals::AppManager()->m_pServerExoApp->GetWorldTimer()->GetWorldTimeCalendarDay();`

Plugins/Player/Player.cpp
```diff
@@ -1336,4 +1344,171 @@ ArgumentStack Player::SetResManOverride(ArgumentStack&& args)
...
+            //If server owner leaves this 0 - the entry will be added with now() time
+            if(timeDay == 0)
+            {
+                uint32_t hour = Globals::AppManager()->m_pServerExoApp->GetWorldTimer()->GetWorldTimeHour();
+                uint32_t minute = Globals::AppManager()->m_pServerExoApp->GetWorldTimer()->GetWorldTimeMinute();
+                uint32_t second = Globals::AppManager()->m_pServerExoApp->GetWorldTimer()->GetWorldTimeSecond();
+                uint32_t ms = Globals::AppManager()->m_pServerExoApp->GetWorldTimer()->GetWorldTimeMillisecond();
+                timeDay = Globals::AppManager()->m_pServerExoApp->GetWorldTimer()->ConvertToTimeOfDay(hour, minute, second, ms);
+            }
```
Daztek: And `timeDay = Globals::AppManager()->m_pServerExoApp->GetWorldTimer()->GetWorldTimeTimeOfDay();` here

Plugins/Player/Player.cpp
Baaleos: Have moved it to Utils as requested

Plugins/Player/Player.cpp
Baaleos: Added ASSERT OR THROW on the tag - my tests showed that the tag is the only parameter that is actually required for success.

Plugins/Player/Player.cpp
Baaleos: Done

Plugins/Player/Player.cpp
Baaleos: Done

Plugins/Player/Player.cpp
Daztek:
```c
    uint32_t nCalendarDay;
    uint32_t nTimeOfDay;
    uint32_t nState;
    uint32_t nPriority;
    uint16_t nPictureIndex;
```

These are all unsigned, should have an `ASSERT_OR_THROW(var >= 0)`

Baaleos: added extra combined assert or throw-  now they are all checked to be >= 0

Daztek: Haha, sorry D: but we prefer them not combined so it's easier to see which assert is being thrown.

Baaleos: lol - done
Reason I kept the 'tag' assertion higher up - is because it is the bare minimum required for the method to function

Baaleos: actually done now: not sure what happened last night - must have forgot to push my changes

Plugins/Player/Player.cpp
```diff
@@ -1336,4 +1344,167 @@ ArgumentStack Player::SetResManOverride(ArgumentStack&& args)
...
+            newJournal.nPictureIndex= 0; // Not implemented by bioware/beamdog  
+            newJournal.bQuestCompleted= completed; 
+            newJournal.bQuestDisplayed= diplayed; 
+            newJournal.bUpdated     = updated; 
+                   
+            auto *pMessage = static_cast<CNWSMessage*>(Globals::AppManager()->m_pServerExoApp->GetNWSMessage());       
+            if (pMessage)
+                {
+                 
+                    int overwrite = -1;
```
GoLoT: It looks like you are using overwrite as a bool instead of an index, the `overwrite = i;` isn't used. I'd rewrite the loop as:

``` cpp
int i = 0;
for (i = 0; i < entries.num; i++)
{
    auto* pEntry = entries.element[i];
    if (pEntry->szPlot_Id.CStr() == tag)
    {
        pCreature->m_pJournal->m_lstEntries[i] = newJournal;
        break;
    }
}
if(i == entries.num)
    pCreature->m_pJournal->m_lstEntries.Add(newJournal);

```

Baaleos: Had some issues implementing the approach above 'exactly'
Merged the main principle described above into the existing for loop structure.

Plugins/Player/Player.cpp
```diff
@@ -1336,4 +1344,163 @@ ArgumentStack Player::SetResManOverride(ArgumentStack&& args)
...
+        auto *pCreature = Globals::AppManager()->m_pServerExoApp->GetCreatureByGameObjectID(pPlayer->m_oidNWSObject);
+        if (pCreature && pCreature->m_pJournal)
+        {
+            auto entries = pCreature->m_pJournal->m_lstEntries;
+            const auto tag = Services::Events::ExtractArgument<std::string>(args);
+            ASSERT_OR_THROW(tag.size() > 0);
+            if (entries.num > 0)
+            {
+                auto pEntry = entries.element;
+                for (int i = 0; i < entries.num; i++, pEntry++)
```
mtijanic: Please no pointer arithmetic :)

`CExoArrayList<>` actually supports modern C++ iterators, so you can probably do:
```C
for (auto& entry : entries)
{
  ...
}
```

Plugins/Player/Player.cpp
```diff
@@ -1336,4 +1344,163 @@ ArgumentStack Player::SetResManOverride(ArgumentStack&& args)
...
+                {
+                    if (pEntry->szPlot_Id.CStr() == tag)
+                        {
+                            lastJournalEntry = entries[i];
+                            found = i;
```
mtijanic: Should we abort if we find the tag, or do we need to look at the last entry with the same tag?
If there's only one with the tag, maybe move the big return here and then you don't need the `found` variable at all.

Plugins/Player/Player.cpp
```diff
@@ -1336,4 +1344,163 @@ ArgumentStack Player::SetResManOverride(ArgumentStack&& args)
...
+            ASSERT_OR_THROW(silentUpdate >= 0);
+            
+            
+            //If server owner leaves this 0 - the entry will be added with todays date
+            if (calDay <= 0)
+            {
+                calDay = Globals::AppManager()->m_pServerExoApp->GetWorldTimer()->GetWorldTimeCalendarDay();
+            }
+            //If server owner leaves this 0 - the entry will be added with now() time
+            if(timeDay <= 0)
```
mtijanic: A few generic style nits throughout the change:
- space between keyword and paren: `if (x)`, not `if(x)`.
- space around assignment operator `=` on both sides:  `a = x;` not `a=x`, `a= x`, `a =x`.
- Avoid superfluous newlines, such as empty lines after `{`

Plugins/Player/Player.cpp
Baaleos: Default journal entries add multiple entries to the player with the same tag - the last one is (as far as I can tell) the one that is generally presented/visible to the player. Even when you remove a journal entry via native RemoveJournalEntry NWScript method - it only ever removes the last entry - leaving some empty journal entries in the journal list on the player. (Somewhat memory-leakish behaviour in the base game)

The AddCustomJournal entry method attempts to avoid that sort of behaviour by overwriting entries with the same tag. This Get method should be compatible with both native journal entries and the custom entries.

Also - the found variable is used so I can return some sort of indication to the NWScript context that no valid journal entry was found - eg: no need to do more popping of return variables if the journal entry was not found etc.
Seemed more efficient to pop one return variable with -1 to indicate to the nwscript that there is no further variables to be expected etc.
If there is a better way of handling that - happy to include it.

mtijanic: In that case, it would be better to start at the last element and iterate backwards until you find the first tag.

And what I mean is if you add the full data return here, you don't need the `found` variable. As soon as you hit the end of the function, you know it's not found, so you can just return `-1` directly.

Baaleos: Ok - will try to refactor to start at entries.num -1 and work down towards 0 etc
Will also add the return directly to the loop

mtijanic:
Is this ready for merge? (i didn't re-eeview yet since last time)

Baaleos:
> Is this ready for merge? (i didn't re-eeview yet since last time)

Not quite - It still needs to have the style nits fixed and needs to be retested against the latest server binary. I will try to get some time put aside for this at the weekend.

plenarius:
Reviving this as I took my time to do my own review and fixed the merge conflicts and wouldn't mind seeing this added. Will spend some time testing a bit.

Baaleos:
> Reviving this as I took my time to do my own review and fixed the merge conflicts and wouldn't mind seeing this added. Will spend some time testing a bit.

Happy for someone to take it over - I've been pre-occupied with RL stuff at the moment.
From what I recall - everything was working - but then I implemented some of the refactoring that was recommended, such as iterating through the list backwards etc for efficiency. Just never got a chance to test that part etc.

PR #818, opened 2020-03-10
Add an iRanged flag to NWNX_Damage_DealDamage()

GideonCrawle:
This change allows the scripter to specify that damage dealt is ranged only. Default value of FALSE in NWScript so shouldn’t break compatibility.

### DIFF
--- a/Plugins/Damage/Damage.cpp
+++ b/Plugins/Damage/Damage.cpp
@@ -238,6 +238,7 @@ ArgumentStack Damage::DealDamage(ArgumentStack&& args)
         positive[k] = vDamage[k] > 0;
     }
     int damagePower = Services::Events::ExtractArgument<int32_t>(args);
+    int range = Services::Events::ExtractArgument<int>(args);
 
     CNWSCreature *pSource = Globals::AppManager()->m_pServerExoApp->GetCreatureByGameObjectID(oidSource);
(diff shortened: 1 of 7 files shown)


Plugins/Damage/Damage.cpp
```diff
@@ -238,6 +238,7 @@ ArgumentStack Damage::DealDamage(ArgumentStack&& args)
         positive[k] = vDamage[k] > 0;
     }
     int damagePower = Services::Events::ExtractArgument<int32_t>(args);
+    int range = Services::Events::ExtractArgument<int>(args);
```
mtijanic: Maybe this should be in a try/catch, so that if something is not pushed (because of old nss file) it doesn't break existing code, but just print a warning and assume 0?

GideonCrawle:
```suggestion
    int range = 0;
    try {
        range = Services::Events::ExtractArgument<int>(args);
    }
    catch(std::exception e) {
        LOG_WARNING("NWNX_Damage_DealDamage() called from NWScript without final parameter. Please download the latest versions of NWNX scripts.");
    }
```

Is there any sense to outputting the actual exception text, since there's more specific information available and the problem is readily apparent?

Daztek: Should be fine without the actual exception text

PR #830, opened 2020-03-31
Add GetFeatGrantLevel, GetInternalObjectType, PluginExists, GetUserDirectory

Jorteck:
**Creature - GetFeatGrantLevel**
When a feat is removed from a creature, there is no easy API to determine which level this feat was removed from, if we need to restore it later using AddFeatByLevel.

**Object - GetInternalObjectType**
Modules, Areas and a few other object types do not have their own OBJECT_TYPE_* constant that is returned from GetObjectType. NWNX knows this exact type, and this would be super handy to have for type safety in a managed implementation (e.g. DotNet).

**Util - PluginExists**
Currently there is no way to check if a plugin exists (e.g. to check if a plugin is an optional dependency) without throwing an error. I'm not sure if this might make more sense as a core function.

**Util - GetUserDirectory**
When using other plugins for databases and persistent data (e.g. EF in DotNet), this can be useful for getting a location to store such data.

### DIFF
--- a/Plugins/Creature/Creature.cpp
+++ b/Plugins/Creature/Creature.cpp
@@ -62,6 +62,7 @@ Creature::Creature(const Plugin::CreateParams& params)
     REGISTER(GetKnowsFeat);
     REGISTER(GetFeatCountByLevel);
     REGISTER(GetFeatByLevel);
+    REGISTER(GetFeatGrantLevel);
     REGISTER(GetFeatCount);
     REGISTER(GetFeatByIndex);
     REGISTER(GetMeetsFeatRequirements);
@@ -264,6 +265,34 @@ ArgumentStack Creature::GetFeatByLevel(ArgumentStack&& args)
         }
     }
     return Services::Events::Arguments(retVal);
+}
+
+ArgumentStack Creature::GetFeatGrantLevel(ArgumentStack&& args)
+{
+    ArgumentStack stack;
+    if (auto* pCreature = creature(args))
+    {
+        const auto feat = Services::Events::ExtractArgument<int32_t>(args);
+        ASSERT_OR_THROW(feat >= Constants::Feat::MIN);
+        ASSERT_OR_THROW(feat <= Constants::Feat::MAX);
+        const auto uFeat = static_cast<uint16_t>(feat);
+
+        for (int32_t i = 0; i < pCreature->m_pStats->GetLevel(); i++)
+        {
+            auto* pLevelStats = pCreature->m_pStats->m_lstLevelStats.element[i];
+            ASSERT_OR_THROW(pLevelStats);
+
+            for (int32_t j = 0; j < pLevelStats->m_lstFeats.num; j++)
+            {
+                if (pLevelStats->m_lstFeats.element[j] == uFeat)
+                {
+                    return Services::Events::Arguments(i + 1);
+                }
+            }
+        }
+    }
+
+    return Services::Events::Arguments(0);
 }
 
 ArgumentStack Creature::GetFeatCount(ArgumentStack&& args)
--- a/Plugins/Object/NWScript/nwnx_object.nss
+++ b/Plugins/Object/NWScript/nwnx_object.nss
@@ -16,6 +16,28 @@ const int NWNX_OBJECT_LOCALVAR_TYPE_OBJECT   = 4;
 const int NWNX_OBJECT_LOCALVAR_TYPE_LOCATION = 5;
 /// @}
 
+/// @anchor object_internal_types
+/// @name Internal Object Types
+/// @{
+const int NWNX_OBJECT_TYPE_INTERNAL_INVALID = -1;
+const int NWNX_OBJECT_TYPE_INTERNAL_GUI = 1;
+const int NWNX_OBJECT_TYPE_INTERNAL_TILE = 2;
+const int NWNX_OBJECT_TYPE_INTERNAL_MODULE = 3;
+const int NWNX_OBJECT_TYPE_INTERNAL_AREA = 4;
+const int NWNX_OBJECT_TYPE_INTERNAL_CREATURE = 5;
+const int NWNX_OBJECT_TYPE_INTERNAL_ITEM = 6;
+const int NWNX_OBJECT_TYPE_INTERNAL_TRIGGER = 7;
+const int NWNX_OBJECT_TYPE_INTERNAL_PROJECTILE = 8;
+const int NWNX_OBJECT_TYPE_INTERNAL_PLACEABLE = 9;
+const int NWNX_OBJECT_TYPE_INTERNAL_DOOR = 10;
+const int NWNX_OBJECT_TYPE_INTERNAL_AREAOFEFFECT = 11;
+const int NWNX_OBJECT_TYPE_INTERNAL_WAYPOINT = 12;
+const int NWNX_OBJECT_TYPE_INTERNAL_ENCOUNTER = 13;
+const int NWNX_OBJECT_TYPE_INTERNAL_STORE = 14;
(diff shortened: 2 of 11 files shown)


Plugins/Util/NWScript/nwnx_util.nss
```diff
@@ -163,6 +163,15 @@ int NWNX_Util_RegisterServerConsoleCommand(string sCommand, string sScriptChunk)
 /// @param sCommand The name of the command.
 void NWNX_Util_UnregisterServerConsoleCommand(string sCommand);
 
+/// @brief Determines if the given plugin exists and is enabled.
+/// @param sPlugin The name of the plugin to check.
```
Daztek, line 167: Should probably note that it expects the plugin name as `NWNX_{Name}`, case-sensitive and all.

Unless you want to go the extra mile and strip the prefix only when needed and use `GetServices()->m_plugins->GetCanonicalPluginName()` to deal with case-sensitivity. :P

Jorteck: I will add a comment ;)

Removing the ` NWNX_*` prefix might be a bit much, I think it makes sense to match the plugin naming used in `NWNX_Core`.

Plugins/Creature/Creature.cpp
```diff
@@ -266,6 +267,38 @@ ArgumentStack Creature::GetFeatByLevel(ArgumentStack&& args)
...
+        for (int32_t i = 0; i < pCreature->m_pStats->GetLevel() - 1; i++)
+        {
+            auto* pLevelStats = pCreature->m_pStats->m_lstLevelStats.element[i];
+            ASSERT_OR_THROW(pLevelStats);
+
+            for (int32_t j = 0; j < pLevelStats->m_lstFeats.num; j++)
+            {
+                if (pLevelStats->m_lstFeats.element[j] != uFeat)
+                {
+                    continue;
```
mtijanic, line 289: Style nit: rather than `if (x != y) continue;`, it's clearer when you do `if (x == y) { do_thing(); }`.

Plugins/Creature/Creature.cpp
```diff
@@ -266,6 +267,38 @@ ArgumentStack Creature::GetFeatByLevel(ArgumentStack&& args)
...
+                }
+
+                Services::Events::InsertArgument(stack, i + 1);
+                return stack;
```
mtijanic, line 293: Style nit: rest of this file uses oneline `return Services::Events::Arguments(i+1);`

Plugins/Object/NWScript/nwnx_object.nss
```diff
@@ -299,6 +321,11 @@ void NWNX_Object_DeleteVarRegex(object oObject, string sRegex);
 /// @return TRUE if vPosition is inside oTrigger's geometry.
 int NWNX_Object_GetPositionIsInTrigger(object oTrigger, vector vPosition);
 
+/// @brief Gets the given object's internal type (NWNX_OBJECT_TYPE_INTERNAL_*)
+/// @param oObject The object.
+/// @return The object's type (NWNX_OBJECT_TYPE_INTERNAL_*)
+int NWNX_Object_GetInternalObjectType(object oObject);
```
mtijanic, line 327: See https://github.com/nwnxee/unified/blob/master/Plugins/Tweaks/Tweaks/StringToIntBaseToAuto.cpp

This is prior art for just modifying existing nwscript functions to do something else. I'd rather just extend the default `GetObjectType()` to return `OBJECT_TYPE_AREA` and `OBJECT_TYPE_MODULE` in addition to the other stuff.

Also, this is already possible to do in pure nwscript, by just checking something like:
```C
int myGetObjectType(object oObj) {
  if (oObj == GetModule())
    return OBJECT_TYPE_MODULE;
  if (GetIsAreaAboveGround(oObj) != AREA_INVALID);
    return OBJECT_TYPE_AREA;
  return GetObjectType(oObj);
}
```

Plugins/Object/Object.cpp
```diff
@@ -839,4 +840,22 @@ ArgumentStack Object::GetPositionIsInTrigger(ArgumentStack&& args)
     return Services::Events::Arguments(retVal);
 }
 
+ArgumentStack Object::GetInternalObjectType(ArgumentStack&& args)
+{
+    const auto objectId = Services::Events::ExtractArgument<Types::ObjectID>(args);
+    ArgumentStack stack;
+
+    if (objectId == Constants::OBJECT_INVALID)
```
mtijanic, line 848: style: invert logic.
```C
{
    int retVal = -1;
    const auto objectId = ...;
    if (auto* go = Utils::GetGameObject(objectId))
    {
        retVal = go->m_nObjectType;
    }
    return Services::Events::Arguments(retVal);
}

File: Plugins/Util/Util.cpp [hunk 6]
```diff
@@ -548,4 +552,31 @@ ArgumentStack Util::UnregisterServerConsoleCommand(ArgumentStack&& args)
...
+ArgumentStack Util::PluginExists(ArgumentStack&& args)
+{
+    std::string pluginName = Services::Events::ExtractArgument<std::string>(args);
+    std::string pluginNameWithoutPrefix = pluginName.substr(5, pluginName.length() - 5);
+
+    ArgumentStack stack;
+
+    if (GetServices()->m_plugins->FindPluginByName(pluginNameWithoutPrefix))
+    {
+        Services::Events::InsertArgument(stack, 1);
```
mtijanic, line 564: Style: As above, you can invert the logic to simplify it.

Plugins/Util/Util.cpp
```diff
@@ -548,4 +552,31 @@ ArgumentStack Util::UnregisterServerConsoleCommand(ArgumentStack&& args)
...
+        Services::Events::InsertArgument(stack, 0);
+    }
+
+    return stack;
+}
+
+ArgumentStack Util::GetUserDirectory(ArgumentStack&&)
+{
+    ArgumentStack stack;
+    char* retval = Globals::ExoBase()->m_sUserDirectory.CStr();
```
mtijanic, line 577: Style: Oneliner:
```C
return Services::Events::Arguments(Globals::ExoBase()->m_sUserDirectory.CStr());
```

Plugins/Util/NWScript/nwnx_util.nss
```diff
@@ -163,6 +163,15 @@ int NWNX_Util_RegisterServerConsoleCommand(string sCommand, string sScriptChunk)
...
+/// @return TRUE if the plugin exists and is enabled, otherwise FALSE.
+int NWNX_Util_PluginExists(string sPlugin);
```
mtijanic, line 170: I'm not thrilled that this is part of util, but probably good enough for now.

Jorteck:
I've fixed up the points above and squashed those commits :)

> See https://github.com/nwnxee/unified/blob/master/Plugins/Tweaks/Tweaks/StringToIntBaseToAuto.cpp
> 
> This is prior art for just modifying existing nwscript functions to do something else. I'd rather just extend the default `GetObjectType()` to return `OBJECT_TYPE_AREA` and `OBJECT_TYPE_MODULE` in addition to the other stuff.
> 
> Also, this is already possible to do in pure nwscript, by just checking something like:
> 
> ```c
> int myGetObjectType(object oObj) {
>   if (oObj == GetModule())
>     return OBJECT_TYPE_MODULE;
>   if (GetIsAreaAboveGround(oObj) != AREA_INVALID);
>     return OBJECT_TYPE_AREA;
>   return GetObjectType(oObj);
> }
> ```

What would be the nicest way to add these other object types to GetObjectType without breaking existing scripts? I'm also not sure if there's a way to determine if something is a valid sound object for `SoundObjectPlay()`

mtijanic:
Merging now, and we can discuss about extending GetObjectType later. Thanks!

PR #852, opened 2020-04-14
Large scale refactor, part 1

### LINKED ISSUE #868: [RFC] Merge all the inl files into headers.
They just make the codebase harder to navigate, imo.

### LINKED ISSUE #921: Remove all heap allocations from the crash handler path
Otherwise in case of heap corruption we get a nested segfault which gives no useful info - see logs in #920

mtijanic:
Various misc improvements:

* Exclusive hooks can now return the hook, no need for two calls
* use `s_` for static variables everywhere
* minor copy optimizations
* Get rid of all the `NWNXLib::API::Types` stuff
* Get rid of some build configs that were never used
* Collapse multiple namespaces into nested ones
* Remove .inl files that are just noise (resolves #868)
* Handle errors that happen during stack trace printing (resolves #921)
* Get rid of PluginInfo and a ton of plugin boilerplate

...to be continued..


mtijanic:
Okay, if I continue this will become absolutely unmanageable (it already is), so I've cut off at some point, and the rest will be separate PRs.

PR #880, opened 2020-05-03
tweaks: Static member variables already zero initialized

jd28:
From standard: “Variables with static storage duration (3.7.1) or thread storage duration (3.7.2) shall be zero-initialized (8.5) before any other initialization takes place”


jd28:
There coudld be an issue with this PR if classes.2da were modified and `NWNX_Administration_ReloadRules` was called, I just realized that.

GoLoT:
@mtijanic mentioned in his review that https://github.com/nwnxee/unified/pull/880/files#diff-98ae8517219b13a69eec672945e0c63dL35 isn't needed. I wasn't aware that array contents were initialized to 0 by default, so feel free to remove it.

The other block is needed to reset the values when 2das are reloaded as you said.

jd28:
I reverted the deletion of the second block.

PR #881, opened 2020-05-03
services/config: Use if constexpr and Utils::from_string to simplify

jd28:
One (potentially) nice fallout is anything that specializes Utils::from_string could be used as a configuration type.  Other than that, just cleanup.


mtijanic:
Is this still WIP?

jd28:
Well... I haven't worked on it for awhile besides testing, so I figured best to let anyone that might have feedback the opportunity to share it.

jd28:
Removed WIP tag.  I think it's as good as it can be without adding any needless disruptions to the rest of the code base.

PR #882, opened 2020-05-04
dockerignore: Add lines from gitignore

jd28:
Avoid copying unnecessary crap to docker image.  I admit a lack of experience with docker, so if this is wrong please advise and I will close PR.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -8,3 +8,11 @@ LICENSE
 *.md
 .git*
 .circleci
+.idea/
+.vscode/
+Binaries/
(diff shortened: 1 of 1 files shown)


mtijanic:
This looks sane to me but I don't know much about docker. @jakkn @urothis any concerns?

urothis:
Looks good to me.

jakkn:
Should work, yeah. Can also flip it if that makes sense.
```
*
!Core
!NWNXLib
!Plugins
!Scripts
```

PR #890, opened 2020-05-09
Some minor things

jd28:
Only https://github.com/nwnxee/unified/commit/a4f0b3ad114ed09bf01f00d964c19d1a66ab1074 probably needs some review.  On it's face I'm not sure how/if it worked?


NWNXLib/Encoding.cpp
```diff
@@ -180,7 +180,7 @@ std::string ToBase64(const std::vector<uint8_t>& in)
     {
         out.push_back('=');
     }
-    return std::move(out);
+    return out;
```
mtijanic, line 183: What is the actual pesimization here?

mtijanic: Disassembly to the rescue. Nevermind.

PR #896, opened 2020-05-11
[WIP] Scriptable special attacks

GoLoT:
It's feedback time. I decided to go with a new plugin because this didn't fit any of the existing plugins and I have to add other combat engine related functions later on.

I still have to do thorough testing and the custom scripts only have access to the attacker oid (`OBJECT_SELF`) at the moment. I will expose the attacked object and other relevant information (damage, feat ID...) soon.

If there is any other info that could be useful to expose to the custom scripts or any other features that someone wants to be added, leave a comment here.

### DIFF
--- a/Plugins/Combat/CMakeLists.txt
+++ b/Plugins/Combat/CMakeLists.txt
@@ -1,0 +1,2 @@
+add_plugin(Combat
+    "Combat.cpp")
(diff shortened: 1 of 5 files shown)


GoLoT:
Added functions to get the feat ID, target oid and script type.

Results from the example script in readme.md:

image

Issue #820, opened 2020-03-12
[Feature request] Get/SetCriticalMultiplier(object oPlayer, int nBaseItem)


silent-rock:
I'm not sure if it's possible, but it would be an awesome feature to have to be able to modify weapon's critical hit multiplier via nwnx_weapon plugin like the Weapon Master's Increased multiplier feat does.

Baaleos:
While not technically the same as a crit multiplier - the nwnx_damage plugin could adjust the damage to your hearts content, based on the same criteria.

`/// @struct NWNX_Damage_AttackEventData
/// @brief Attack Event Data
struct NWNX_Damage_AttackEventData
{
    object oTarget; ///< The target who took the damage
    int iBludgeoning; ///< Bludgeoning damage
    int iPierce; ///< Piercing damage
    int iSlash; ///< Slashing damage
    int iMagical; ///< Magical damage
    int iAcid; ///< Acid damage
    int iCold; ///< Cold damage
    int iDivine; ///< Divine damage
    int iElectrical; ///< Electrical damage
    int iFire; ///< Fire damage
    int iNegative; ///< Negative damage
    int iPositive; ///< Positive damage
    int iSonic; ///< Sonic damage
    int iBase; ///< Base damage
    int iAttackNumber; ///< 1-based index of the attack in current combat round
    int iAttackResult; ///< 1=hit, 3=critical hit, 4=miss, 8=concealed
    int iAttackType;   ///< 1=main hand, 2=offhand, 3-5=creature, 6=haste
    int iSneakAttack;  ///< 0=neither, 1=sneak attack, 2=death attack, 3=both
};`

You could catch the attack data, see if it is a critical hit, see if you have the required feat and then adjust the damage to match whatever you want.

Eg: If the players has the FIRE_WEAPON_FEAT (I just made that up)
The weapon has fire damage + it is a critical hit, then you double the fire damage etc.

mtijanic:
Should be easy. Just an exclusive hook on `CNWSCreatureStats::GetCriticalHitMultiplier`, call original, then read additional modifier from POS.
There's also `CNWSCreatureStats::GetCriticalHitRoll` for the threat range.

jd28:
Anyone interested in implementing something like this might well look at how Acaos did it https://github.com/NWNX/nwnx2-linux/tree/master/plugins/weapons approx. 12 years ago.

jd28:
[...]

PR #897, opened 2020-05-12
NWNX_Creature: Get/Set Critical Multiplier/Range Override/Modifier

WilliamDraco:
Generally resolves issue #820 - Implementation for specific BaseItems can be managed by NWScript.

PR adds a series of functions to:
- Get or Set
- Critical Multiplier, or the Critical Hit Range
- as either an Override, or a modifier of the default
- for either both hands, Main-hand only, or offhand only.

The character sheet adjusts appropriate when an appropriate trigger is called, but reacts poorly beyond reasonable values (which still function in combat).

Persistent via POS, but does require a Set Multiplier, and Set Range (either mod or ovr) to first-hook each of the functions - An issue which I'm told requires a more general solution.


Plugins/Creature/Creature.cpp
```diff
@@ -2147,12 +2157,342 @@ ArgumentStack Creature::JumpToLimbo(ArgumentStack&& args)
...
     return Services::Events::Arguments();
 }
 
+void Creature::InitCriticalMultiplierHook()
+{
+    static NWNXLib::Hooking::FunctionHook* pGetCriticalHitMultiplier_hook;
+    if (!pGetCriticalHitMultiplier_hook)
+    {
+        g_plugin->GetServices()->m_hooks->RequestExclusiveHook<Functions::_ZN17CNWSCreatureStats24GetCriticalHitMultiplierEi>(
+            +[](CNWSCreatureStats *pThis, bool bOffHand = false) -> int32_t
```
Daztek, line 2175: The `BOOL` type in the API is actually an int32_t, so `bool bOffHand = false` should be `int32_t bOffHand = false`.

Plugins/Creature/Creature.cpp
```diff
@@ -2147,12 +2157,342 @@ ArgumentStack Creature::JumpToLimbo(ArgumentStack&& args)
...
+    return Services::Events::Arguments(retVal);
+}
+
+void Creature::InitCriticalRangeHook()
+{
+    static NWNXLib::Hooking::FunctionHook* pGetCriticalHitRoll_hook;
+    if (!pGetCriticalHitRoll_hook)
+    {
+        g_plugin->GetServices()->m_hooks->RequestExclusiveHook<Functions::_ZN17CNWSCreatureStats18GetCriticalHitRollEi>(
+            +[](CNWSCreatureStats *pThis, bool bOffHand = false) -> int32_t
```
Daztek, line 2340: Same for this bool.

jd28:
For players these functions would really need to be gated by (Un)Equip, LevelUp/Down, (Un)ApplyEffect, etc (depending on how ambitious people want to get) events or they will be exploited.  So the persistence aspect might cause confusion.   Even the usage of POS over local vars I wonder about cause for NPCs it's easier to just set them in the toolset.  But these are stray thoughts, so feel free to ignore.

mtijanic:
This has been up for too long waiting for me to review. The API is fine, so I'm gonna merge it, and any changes to the backend can be done as a follow on.

Issue #606, opened 2019-10-13
Generating a NWNX:EE Test Module during build.


jd28:
@mtijanic suggested a creating an issue for this.  I don't know anything about CircleCI, so I'm not sure if anything I'm going to suggest here (or if spinning up nwserver is at all feasible) to cause a build to fail if test scripts failed.

Anyway, some ideas:
* Add a `NWNX_Assert(int result, string message);`
* Change test scripts from `void main` to `int StartingConditional` so that some test runner could at least inspect whether or not tests succeed or failed.

mtijanic:
I don't know how feasible it is to run nwserver in CircleCI, but I wouldn't count on it. As for the options, NWNX_Assert seems like the way to go, but we might be able to do it without C++ side changes. I'm not sure if it is possible to run StartingConditional scripts outside of a conversation with a PC.

Should change those `report()` calls into `NWNX_Assert()`, streamline the prints and make it obvious whether an error has happened or not. Also report all errors at the end of the process.

Can use NWNX_Utils to register a pre-module-load script, which will use resman to query all `nwnx_*_t.ncs` files and execute them. That way, we can run the unit tests on any module by just dropping NCS files somewhere in resman.

I'm tagging this as help wanted and a good first issue because most of the work is just on the nwscript side of things, so this is a good opportunity for people who wanted to help but don't have the necessary C++ experience.

jd28:
I'm pretty sure some people used to do the `int StartingConditional` thing in the old days:

https://github.com/NWNX/nwnx2-linux/blob/fc49bb4abd064d74bca3f29d218bc62a4e048a37/plugins/events/HookFunc.cpp#L599-L607

I'll open a separate issue for this if you think it's worthwhile, unrelated to this issue, as adding it would be a good first issue too.

PR #904, opened 2020-05-15
[wip/poc] Add test infrastructure

jd28:
So.. I was thinking more about https://github.com/nwnxee/unified/issues/606
This is just a basic proof-of-concept structured as a plugin using [catch2](https://github.com/catchorg/Catch2).  There is a bit of a compile time hit for the cpp file with the actual test runner.

If people think it would be of any value, it would be refactored into Core with a `NWNX_Core_RunTests` event and a basic `nwnx_test.mod` so that:

- Plugins could add their own test cases as they saw fit.
- Server hosts could change their module to `nwnx_test.mod` in their own testing environments and ensure all the plugins they use are working as expected prior to upgrading.

But I'm curious if there would be much interest or if anyone had any better ideas?

### DIFF
--- a/Plugins/Test/CMakeLists.txt
+++ b/Plugins/Test/CMakeLists.txt
@@ -1,0 +1,4 @@
+add_plugin(Test
+    Test.cpp
+    Test2.cpp
+)
--- a/Plugins/Test/Test.cpp
+++ b/Plugins/Test/Test.cpp
@@ -1,0 +1,61 @@
+#include "Test.hpp"
+#include "Utils/String.hpp"
+
+#define CATCH_CONFIG_RUNNER
+#include "catch.hpp"
+
+static Test::Test* g_plugin;
+
+using namespace NWNXLib;
+using namespace NWNXLib::API;
+
+NWNX_PLUGIN_ENTRY Plugin::Info* PluginInfo()
+{
+    return new Plugin::Info
+            {
+                    "test",
+                    "Test some stuff",
+                    "jmd",
+                    "jmd2028@gmail.com",
+                    1,
+                    true,
+                    0
+            };
+}
+
+NWNX_PLUGIN_ENTRY Plugin* PluginLoad(Plugin::CreateParams params)
+{
+    g_plugin = new Test::Test(params);
+    return g_plugin;
+}
+
+namespace Test
+{
+
+Test::Test(const Plugin::CreateParams& params)
+        : Plugin(params)
+{
+    LOG_NOTICE("Running Tests...");
+    int result = Catch::Session().run();
+    if(result) 
+    {
+        LOG_FATAL("Tests failed");
+    }
+    else 
+    {
+        LOG_NOTICE("Tests passed");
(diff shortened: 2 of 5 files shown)


mtijanic:
I'm not sure testing on the cpp side of things is very valuable. It's not part of a public API and has in the past gone through several refactors. Tests in these cases would have been just a hassle, as you would have to update them each time.

What we should be looking into is testing at the API level. Basically, streamline the nwnx_foo_t.nss files, and add a switch to the core to check for all loaded plugins and run their tests, collect the results and display it somehow.
The NSS tests, crappy as they are, have been incredibly valuable in the past (most notably with the switch to 64bit).

The testing framework (i.e. catch2 equivalent) should be on the nss side.

jd28:
I agree that there'd be limited value for most NWNX related things.

On the other hand I don't see much practical difference between
```
    NWNX_Administration_SetPlayerPassword(sString);
    NWNX_Tests_Report("NWNX_Administration", "{Set/Get}PlayerPassword", NWNX_Administration_GetPlayerPassword() == sString);
 ```
and say something like:

```c++
TEST_CASE("SetPlayerPassword", "[NWNX_Administration]") {
    std::string pw{"FuzzyKittens"};
    g_plugin->SetPlayerPassword(Args(pw));
    REQUIRE(pw == Globals::AppManager()->m_pServerExoApp->GetNetLayer()->GetPlayerPassword().CStr());
}
```
Other than a good testing library for C++ already exists.  If anything it seems like people would be more generally likely to write tests and it'd be a huge boon in general to leave off nwscript tests altogether and test the public api on the C++ side, cause it'd just be recompile and go.  Things will already blow up terribly if the wrong amount/type of arguments are pushed from the nwscript side.  In addition some hooks which would be trivially callable on the C++ side to test them would potentially require adding events just to inspect their values.

I say all this with the huge caveat that I'm obviously not an active plugin writer and the chance of me porting my server from 1.69 is hovering around 0%.  So I don't mean to argue with whatever works best for real user workflows.

mtijanic:
Big difference between the two is that the nss one also tests:
- the arguments are pushed correctly on nss side
- argumemt passing works (very important when we change the abi)
- they are popped correctly on c++
- same for return value

A lot of nasty bugs can creep in that layer.

IMO testing framework on the nss side (with a few c++ utility functions) can be quite good, and the nss files could get embedded into nwnx_test.so,and automatically run it all

mtijanic:
Oh, one more benefit of moving testing up to nss: People who want to help but don't know C++ can write the tests and offload some effort from the plugin developers.

jd28:
Alright well I'll go ahead and close this one.

jd28:
How about something like this then?  It has the benefit of not needing a special module at least, just drop the scripts in <nwn>/development.  I can't reopen cause I force pushed.. oops.  But if this is more in line with what you were wanting, it still needs to be cleaned way up.

jd28:
Crap I'm going to reopen a new one...

PR #906, opened 2020-05-15
DotNET - Search pack folder for libnethost

Jorteck:
This should be a more permanent fix over adding each dotnet-sdk release to the paths lists (e.g. #905)

### DIFF
--- a/Plugins/DotNET/DotNET.cpp
+++ b/Plugins/DotNET/DotNET.cpp
@@ -85,25 +85,35 @@ bool DotNET::InitThunks()
         if (dir != nullptr)
         {
             dirent* directoryEntry = readdir(dir);
+            std::vector<std::string> paths;
 
             while (directoryEntry != nullptr)
             {
                 if (directoryEntry->d_type == DT_DIR)
                 {
                     const auto path = (std::string(hostBaseDir) + directoryEntry->d_name + hostLibSuffix);
-
-                    nethost = dlopen(path.c_str(), RTLD_LAZY);
-                    if (nethost)
-                    {
-                        LOG_INFO("Loaded libnethost.so from: %s (autodetected)", path);
-                        break;
-                    }
+                    paths.push_back(path);
                 }
 
                 directoryEntry = readdir(dir);
             }
 
             closedir(dir);
+
+            if (!paths.empty())
+            {
+                std::sort(paths.begin(), paths.end(), std::greater<std::string>());
+                for (std::string path : paths)
+                {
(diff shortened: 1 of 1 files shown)


Plugins/DotNET/DotNET.cpp
```diff
@@ -81,6 +75,38 @@ bool DotNET::InitThunks()
...
+        const auto hostBaseDir = "/usr/share/dotnet/packs/Microsoft.NETCore.App.Host.linux-x64/";
+        const auto hostLibSuffix = "/runtimes/linux-x64/native/libnethost.so";
+
+        DIR* dir = opendir(hostBaseDir);
+
+        if (dir != nullptr)
+        {
+            dirent* directoryEntry = readdir(dir);
+
+            while (directoryEntry != nullptr)
```
mtijanic, line 89: Do you always have exactly one path installed? If you have multiple versions, this will pick one semi-randomly.
I'm not sure this is even an issue - I half expect all the versions of libnethost.so to be byte-per-byte identical, but it's something we need to look into before merging.

Jorteck: I think this is currently an issue with docker as the image targets a major version number (dotnet-3.1 currently), and this tag is updated with a minor version release (3.1.3 -> 3.1.4) essentially breaking dotnet per release (or requiring a manual config update) if you are using docker.

urothis: Correct, the dotnet version needs to be updated in the nwnxee dockerfile as well. 
There should only be one libnethost.so per dotnet install (I would hope.)

With this fix in place we should be able to increment the dotnet version in the dockerfile and not have to worry adding paths anymore. 

I have not tested this pr, but if it works it would make our lives easier.

mtijanic: I have merged #905 for now until I (or someone else) verify that this will work as expected if you have multiple versions of dotnet installed. Otherwise might need to sort the dentrys and go from the top.

mtijanic:
Thanks. Did you test the latest version by any chance?

Jorteck:
With #905? I did, that was working. This is just a future proofing method :)

Jorteck:
I tested this in docker, and an older native copy which had 3.0.8 and 3.1.1

mtijanic:
Cool thanks. It looks correct, I was just wondering if I need to pull locally and give it a run. If nothing exploded I'll merge (as soon as CI is done), and if something does break we'll fix it later.

PR #908, opened 2020-05-16
plugins/test: Add test infrastructure

jd28:
Continued from https://github.com/nwnxee/unified/pull/904

Is this kind of more inline with what you were suggesting?  I like that it doesn't need a special module, DockerDemo works fine.  It needs to be cleaned way up, but want to make sure this is the right direction.

Edit: Or did you mean using like `xxd` and just including the file straight into a string and compile at run time?

### DIFF
--- a/Plugins/Test/Test.cpp
+++ b/Plugins/Test/Test.cpp
@@ -1,0 +1,299 @@
+#include "Test.hpp"
+
+#include "Utils/String.hpp"
+#include "External/rang/rang.hpp"
+#include "Services/Plugins/Plugins.hpp"
+#include "Services/Config/Config.hpp"
+
+#include "API/API/CExoFile.hpp"
+#include "API/API/CExoStringList.hpp"
+#include "API/API/CKeyTableEntry.hpp"
+#include "API/API/CScriptCompiler.hpp"
+#include "API/API/CExoResMan.hpp"
+#include "API/API/CRes.hpp"
+
+#include <dirent.h>
+#include <fstream>
+#include <ftw.h>
+#include <fnmatch.h>
+#include <linux/limits.h>
+#include <regex>
+
+Test::Test* g_plugin;
+
+using namespace NWNXLib;
+using namespace NWNXLib::API;
+
+NWNX_PLUGIN_ENTRY Plugin::Info* PluginInfo()
+{
+    return new Plugin::Info
+        {
+            "Test",
+            "Test some stuff",
+            "jmd",
+            "jmd2028@gmail.com",
+            1,
+            true,
+            0
+        };
+}
+
+NWNX_PLUGIN_ENTRY Plugin* PluginLoad(Plugin::CreateParams params)
+{
+    g_plugin = new Test::Test(params);
+    return g_plugin;
+}
+
+namespace Test
+{
+
+Test::Test(const Plugin::CreateParams& params)
+        : Plugin(params),
+            m_compiler{std::make_unique<CScriptCompiler>()}
+{
+    m_sourceDir = GetServices()->m_config->Get<std::string>("SOURCE_DIR", "");
+    m_waitForPC = GetServices()->m_config->Get<bool>("WAIT_FOR_PC", false);
+
+    if (m_sourceDir.empty())
+    {
+        // Check if the source path used to build exists, if so that's the source dir.
+        if (DIR* d = opendir(NWNX_TEST_SOURCE_PATH))
+        {
+            closedir(d);
+            m_sourceDir = NWNX_TEST_SOURCE_PATH;
+        }
+    }
+
+    if (!m_sourceDir.empty())
+        LOG_NOTICE("NWNX:EE Source Directory: %s", m_sourceDir);
+
+#define REGISTER(func) \
+    GetServices()->m_events->RegisterEvent(#func, \
+        [this](ArgumentStack&& args){ return func(std::move(args)); })
+
+    REGISTER(AddTestScript);
+    REGISTER(Context);
+    REGISTER(Flush);
+    REGISTER(ReportFailure);
+
+#undef REGISTER
+
+    if(!m_waitForPC)
+    {
+        GetServices()->m_hooks->RequestSharedHook<API::Functions::_ZN10CNWSModule16LoadModuleFinishEv, uint32_t>
+            (LoadModuleEndHook);
+    }
+    else
+    {
+        GetServices()->m_hooks->RequestSharedHook<Functions::_ZN21CServerExoAppInternal19LoadCharacterFinishEP10CNWSPlayerii, void, CServerExoAppInternal*, CNWSPlayer*, int32_t, int32_t>
+            (LoadCharacterFinishHook);
+    }
+}
+
+Test::~Test()
+{
+}
+
+ArgumentStack Test::AddTestScript(ArgumentStack&& args)
+{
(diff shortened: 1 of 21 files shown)


jd28:
So for the heck if I decided to try using `xxd` to bake scripts straight in at build time.  I didn't bother with all of them but the gist can be got from this.

mtijanic:
Yes, it is closer. Sorry, I don't have much time right now, can we keep this open for a day or two and I'll post more detailed thoughts then?

jd28:
Sure thing, there's no rush about it.  I did split the commits, redid all the cmake stuff, and added an optional  sPluginDepend to the AddTestScript event.

mtijanic:
A rough idea of the test framework on the NSS side that I mentioned:
```C
// Starts a new sub-test if called with a new name, and returns TRUE
// If called with the same name as currently active test, ends the test and returns FALSE.
// Example usage: for (;TEST("xxx");) // will only run one iteration
int TEST(string testname);

// If failure != "", fails the currently active test skipping other subtests and returns FALSE
// If failure == "", returns TRUE.
int ASSERT(string failure, string message="");

// Like ASSERT(), except future checks and subtests will not be skipped
int EXPECT(string failure, string message="");

// Returns "" if obj is a valid object
// Returns a failure message with context if not
string IS_VALID(object obj);

// Returns "" if a == b
// Returns a failure message with context if not
string IS_EQUAL_INT(int a, int b);

// Returns "" if a == FALSE
// Returns a failure message with context if not
string IS_FALSE(int a);

// Returns "" if a == TRUE
// Returns a failure message with context if not
string IS_TRUE(int a);
```

Example test set:

```C
void main()
{
    for (;TEST("Creature");)
    {
        object oCreature = CreateObject(OBJECT_TYPE_CREATURE, "nw_chicken", GetStartingLocation());
        ASSERT(IS_VALID(oCreature), "Test subject creature");

        for (;TEST("Feats");)
        {
            int nFeatCountLvl1 = NWNX_Creature_GetFeatCountByLevel(oCreature, 1);
            int nFeatCountTotal = NWNX_Creature_GetFeatCount(oCreature);

            EXPECT(IS_FALSE(NWNX_Creature_GetKnowsFeat(oCreature, FEAT_PLAYER_TOOL_01)), "GetKnowsFeat");

            NWNX_Creature_AddFeat(oCreature, FEAT_PLAYER_TOOL_01);
            EXPECT(IS_TRUE(NWNX_Creature_GetKnowsFeat(oCreature, FEAT_PLAYER_TOOL_01)), "AddFeat");

            EXPECT(IS_EQUAL_INT(NWNX_Creature_GetFeatCountByLevel(oCreature, 1), nFeatCountLvl1));
            EXPECT(IS_EQUAL_INT(NWNX_Creature_GetFeatCount(oCreature), (nFeatCountTotal+1)));

            EXPECT(IS_EQUAL_INT(NWNX_Creature_GetFeatByIndex(oCreature, nFeatCountTotal), FEAT_PLAYER_TOOL_01));
        }
    }
}
```

Example output if one call fails:
```
Running test case Creature
  Running sub-test case Creature::Feats
    FAIL: AddFeat - Expected TRUE, got FALSE
  Creature::Feats finished:
      4 checks passed, 1 check failed, 0 checks skipped
Creature finished:
    1 check passed, 0 checks failed, 0 checks skipped
    0 subtests passed, 1 subtest failed, 0 subtests skipped
===================================================================
All tests finished. FAILED TESTS:
    Creature
      Creature::Feats
===================================================================
```

There is a _ton_ of room for improvement here, but I think it shows roughly what I had in mind for the nss side.

mtijanic:
On the NWNX side and regarding actually running the tests, the logic should probably be:

1) Detect if you've build NWNX from source. If so, just run the latest `nwnx_xxx_t.nss` files from git directly
2) If not, try searching resman for `nwnx_xxx_t` (ncs and/or nss) and run that.
3) If neither finds it, maybe check binaries for embedded versions, or just ignore. I'm not sure.

The NSS can actually report the strings down to the C++ side and formatting/output be done there.

We can also parse the NSS and implement some kind of reflection so it can print more context.

Like I mentioned, would need _a lot_ more design. For now I'm just trying to convey the rough desired result.

jd28:
Alright, I have a much better idea the direction you want it to go now.  I'll try out some ideas.

jd28:
So I gave it another shot, just implementing the subset of tests you listed and did a very small subset of `nwnx_creature_t`.  I haven't really thought too much about the NWNX side, it's still just baking the scripts in the so in the most naive way.  I also haven't used NWScript for years, so I don't know what's good style

Output Example:
image

Still miles to go, but if this getting at least slightly closer, I think this something really worth doing.  Just using what I had helped me spot a couple bugs and make sure I hadn't messed anything up.

jd28:
I've updated this with most tests converted to a design based on @mtijanic 's.  No rush but I'd appreciate feedback if this off track from where you want to be.  I'm happy to do whatever.

Current sample ouput:
image

Plugins/Test/Test.cpp
```diff
@@ -0,0 +1,291 @@
...
+
+Test::Test(const Plugin::CreateParams& params)
+        : Plugin(params),
+            m_compiler{std::make_unique<CScriptCompiler>()}
+{
+    m_sourceDir = GetServices()->m_config->Get<std::string>("SOURCE_DIR", "");
+
+    if (m_sourceDir.empty())
+    {
+        // Check if we're symlinked in from build/source tree.  All this would be better with std::filesystem
```
mtijanic, line 58: A good way to do this is to try and open `__FILE__`.

mtijanic:
> I've updated this with most tests converted to a design based on @mtijanic 's. No rush but I'd appreciate feedback if this off track from where you want to be. I'm happy to do whatever.

Only did a quick skim, but this is looking great!

Plugins/Test/Test.cpp
jd28: That's a million times better idea,  thank you.

jd28: I opted to go with just adding a define in cmake.  When i tried `__FILE__` it was giving me relative pathnames into the source tree, I wouldn't have guessed that it would but.. at least the actually code here is a lot clearer and more robust.

jd28:
I'm taking off the WIP progress tag.  I think this is pretty good now. 

- [x] NWScript first testing framework that's easy for new contributors to add to and build on.  And general purpose enough for any NWNX:EE users.
- [x] Compile and go, fast iteration.  No need for special modules or manually copying test files.  
- [x] Code is pretty simple and clear for future enhancements. 
- [x] The test runner can run on module load or after the first PC enters, since some tests will require a PC. 
- [x] Converted all the test scripts that don't need direct interaction, like `nwnx_dialog_t*` and added annotations to some test scripts where they're failing for reasons other than bugs. 
- [ ] Compiling NSS scripts that are in resman but not in the NWNX:EE source tree and don't have NCS compatriots.  I forget the incantations to get a file out of resman.  
- [ ] Parsing/Preprocessing - I'm sure you can walk and modify the parse tree with `CScriptCompiler` but, I'm sorry, the interface to that reads like a PSA for why you should use Lua or .NET for your game instead.  Maybe someone familiar with the NWN:EE source could do it.  If I had to I'd opt to build `nwnsc` as a shared library and make `NWNX_Nwnsc`.  Or opt for some more naive awk-like approach.

PR #912, opened 2020-05-19
[Docker/Admin Plugin] Make ShutdownServer graceful

Jorteck:
Currently, the server process seems to treat SIGTERM like SIGKILL, and can sometimes prevent plugins from flushing data to disk. I'm not sure if this might make sense as a nwn issue instead as [SIGTERM should allow for a graceful shutdown.](https://www.gnu.org/software/libc/manual/html_node/Termination-Signals.html).

### DIFF
--- a/Plugins/Administration/Administration.cpp
+++ b/Plugins/Administration/Administration.cpp
@@ -128,9 +128,9 @@ Events::ArgumentStack Administration::SetDMPassword(Events::ArgumentStack&& args
 Events::ArgumentStack Administration::ShutdownServer(Events::ArgumentStack&&)
 {
     LOG_NOTICE("Shutting down the server!");
-    if (kill(getpid(), SIGTERM) != 0)
+    if (kill(getpid(), SIGINT) != 0)
     {
-      LOG_ERROR("Shutdown failed: SIGTERM signal not sent successfully");
+      LOG_ERROR("Shutdown failed: SIGINT signal not sent successfully");
     }
     return Events::Arguments();
 }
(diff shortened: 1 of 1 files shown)


Jorteck:
A quick heads up - this PR seems to break docker auto-restart, unless you add `stop_signal: SIGINT` in the docker-compose file.

mtijanic:
Did you actually have issues with it or are you going by best practices? NWN is special in that it treats SIGINT and SIGTERM exactly the same - it catches them and then just sets a global `bShutdown` flag, which causes it to shut down after the current MainLoop iteration ends. So I don't expect this change to have any effect at all (other than breaking docker as mentioned).

Jorteck:
I observed different behaviours using each signal. I've attached some logs to explain the different behaviour.

If I don't use SIGINT, the server seems to immediately halt, player connections timeout instead of getting a server shutdown message, and there are no log entries for unloading of plugins:

[sigterm_restart.txt](https://github.com/nwnxee/unified/files/4655510/sigterm_restart.txt)
[sigint_restart.txt](https://github.com/nwnxee/unified/files/4655511/sigint_restart.txt)

Daztek:
SIGTERM appears to work fine for me on WSL/Native

mtijanic:
This seems to be a case of docker catching SIGTERM and killing the process rather than passing the signal. Which is terrifying to me, but maybe that's how docker works.

For native, both of these ought to be equivalent, so I don't particularly care which version we keep, whatever works for docker.
in fact, probably best to just set the shutdown flag manually instead of sending a sognal at all. The flag is in the exposed globals.

Daztek:
Probably best to try setting the shutdown flag first, otherwise every admin needs to update their compose file.

jhett12321:
Agreed.

Jorteck:
The shutdown flag seems to work in docker nicely.

PR #915, opened 2020-05-21
object: Clarify note on NWNX_Object_SetMaxHitPoints

jd28:


Daztek:
`NWNX_Object_SetMaxHitPoints ` does work on NPCs tho?

jd28:
I would have thought so, but this test

https://github.com/nwnxee/unified/blob/266c7e099ea71856a36408fcf9f8a5ce563ae67b/Plugins/Object/NWScript/nwnx_object_t.nss#L32-L34

fails for me on `nw_chicken`, but works fine on a `nw_it_contain006`.  Does it work for you?

I just figured things changed.

Daztek:
It sets the base hitpoints, which doesn't take into account any modifiers from CON.

A chicken has a base of 5, subtract 1 hp due to its 8 CON, which gives it a max hp of 4.

You do MAX + 1, which sets its base to 5. The game subtracts 1 due to CON, so you're back at 4hp.

jd28:
Got it, thank you for taking the time to explain.  Everything works as you said.  I just assumed that test had returned `true` at some point.  Closing this.

PR #917, opened 2020-05-23
[WIP] Events & Inter Plugin Communcation Thoughts

jd28:
I started working on this before, but decided to try something more comprehensive.

* Add a cache to `GetEventData` similar in spirit to the DotNET plugin.
* Change the underling storage of Argument to std::variant
    * Bake in the concept to nullity/optionality so the std::optional wrappers aren't needed 
    * Add some type checking.  It's probably too cumbersome to force explicit type conversions/promotions at call sites, but it could be done.
* Make ArgumentStack its own thing.
    * It's opaque to parts of the code base but where it isn't std::stack isn't the funnest thing to work with.  Its logic is now all in one place.
    * It doesn't overtly break the no non-const ref guideline that the {Insert, Extract}Argument helper functions do.  I broke the naming conventions tho (PascalCase for member functions is just not right to me, I can change them tho) 
    * Can use brace-initialization

I added some basic usage examples as well as converting the helper functions.  I feel like it's a clear win for usage brevity, but at the core it is functionally the same..

---

I'm adding here some further proposals, because it's probably not too clear what the point is.

* Separate `Argument` and `ArgumentStack` from Events.  These would be the foundational types for communication either between NWN <-> Core and Plugin <-> Plugin. 
* Add `UserDefined*` to `Argument` for complex types that won't be going to NWScript for when basic NWScript types won't do.
* Add `ArgumentList` helper function as so (psuedo-code)
```c++
template <typename... Ts>
ArgumentStack ArgumentList(Ts&&... args)
{
    ArgumentStack stack;
    (..., stack.push(std::forward<Ts>(args)); // Reverse Parameter Order
    return stack;
}
```
* Change `Messaging::Message` (or get rid of that) and replace it with ArgumentStack.

Now what follows from that is `Messaging::BroadcastMessage` could be called as so:
```c++
    stuff_I_need_an_IDE_for->BroadcastMessage("NWNX_STDOUTER", AgrumentList("HelloWorld"));
```
and on the Handler side it's all the same ArgumentStack stuff that everyone is already familiar with.

```c++
    stuff_I_need_an_IDE_for->SubscribeMessage("NWNX_STDOUTER",
                                                           [](ArgumentStack stack) // This would mean copies, but this is all contrived examples. Could make it const ref and add `peek` to ArgumentStack
                                                           {
                                                               std::cout << stack.extract<std::string>() << std::endl;
                                                           });
```
This really isn't a good example but people seem to be mainly using subscribe message for RPC between plugins and one specific plugin.  So it might well be to add

```c++
    ArgumentStack Messaging::Call(const std::string& name, ArgumentStack&& stack);
    void Messaging::RegisterFunction(const std::string& name, std::function<ArgumentStack(ArgumentStack&&)> function);
```

From there one could see how the whole ball of wax could have been built on the same abstractions and there should be nothing really special or different from the NWScript <-> NWNX bridge and Plugin <-> Plugin.  I was even advocating for that at the dead end of NWNX2.

If one wonder to oneself "Why?", well there are plugins from the past viz. ResMan for which Messaging is wholly unsuitable -- it would still need work beyond this -- and I doubt anyone even likes it anyway.

No one should feel bad if they think this is all stupid or total crap.  My feelings won't be hurt.


NWNXLib/Services/Events/Events.cpp
```diff
@@ -149,23 +157,21 @@ void EventsProxy::ClearEvent(const std::string& eventName)
...
-    if (m_object) return Utils::ObjectIDToString(*m_object);
-    if (m_string) return *m_string;
-    if (m_effect) return *m_effect ? std::string("EffectID:") + std::to_string((*m_effect)->m_nID) : std::string("nullptr effect");
-
-    return std::string("");
+    return std::visit(detail::overloaded {
+        [](auto arg) { return std::to_string(arg); },
+        [](API::Types::ObjectID arg) { return Utils::ObjectIDToString(arg); },
+        [](const std::string& arg) { return arg; },
+        [](NullArgument) { return std::string("(null"); },
```
Mingun, line 172: You forgot to close the bracket

NWNXLib/Services/Events/Events.hpp
```diff
@@ -10,36 +10,97 @@
...
+            return real;
+        }
+
+        template <typename... Ts>
+        std::tuple<Ts...> extract_n() { return {extract<Ts>()...}; }
+
+        template <typename... Ts>
+        void push(Ts&&... arg) 
+        {
+            static_assert(sizeof...(Ts) > 0, "You must insert at least one argument.");
```
Mingun, line 94: Maybe encode this requirement in function signature?
```cpp
template <typename T1, typename... Ts>
void push(T1&& arg1, Ts&&... arg)
```

NWNXLib/Services/Events/Events.cpp
```diff
@@ -149,23 +157,21 @@ void EventsProxy::ClearEvent(const std::string& eventName)
...
-    if (m_effect) return *m_effect ? std::string("EffectID:") + std::to_string((*m_effect)->m_nID) : std::string("nullptr effect");
-
-    return std::string("");
+    return std::visit(detail::overloaded {
+        [](auto arg) { return std::to_string(arg); },
+        [](API::Types::ObjectID arg) { return Utils::ObjectIDToString(arg); },
+        [](const std::string& arg) { return arg; },
+        [](NullArgument) { return std::string("(null)"); },
+        [](CGameEffect* arg) { return arg ? std::string("EffectID:") + std::to_string(arg->m_nID) : std::string("nullptr effect");}
+    }, m_data);
```
mtijanic, line 174: This seems like a step backwords to me. It's about the same verbosity, but much less clearer. It raises the C++ knowledge bar to follow this code quite a lot (I personally know dozens of people who would understand the old version but not the new). And I can't even imagine what sort of eldritch errors it will spew out if something goes wrong.

jd28:
I agree 100%, I realized it was a terrible mistake shortly after I made the
PR.

jd28:
I've updated the original post and name of this, so it's more clear that I'm not just suggesting changing stuff for change's sake.

mtijanic:
Hey, sorry, I still don't have the time to properly think about all the implications here. But some things are obvious improvements with no downsides - like the caching. I think this should be three separate PRs in the end. But feel free to shelve for now until once had time to look into it.

jd28:
I'm gonna  close this and re-open a new one, the commit history is already messed up anyway.  I just don't like submitting code with no plan or submitting a plan with no code.

PR #919, opened 2020-05-24
messaging: Change Handler type to void(const Message&)

jd28:
It's not used very consistently, so I'm not totally sure this is the  right choice.

https://github.com/nwnxee/unified/blob/6dfc95e23e885df343b556278ed2af40d794c0dd/NWNXLib/Services/Messaging/Messaging.cpp#L45-L58

As it is, it looks like it's making a copy per subscriber?

### DIFF
--- a/Plugins/Profiler/Profiler.cpp
+++ b/Plugins/Profiler/Profiler.cpp
@@ -148,14 +148,14 @@ Profiler::Profiler(const Plugin::CreateParams& params)
 
     {
         GetServices()->m_messaging->SubscribeMessage("NWNX_PROFILER_SET_PERF_SCOPE_RESAMPLER",
-        [this, sum](std::vector<std::string> message)
+        [this, sum](const std::vector<std::string>& message)
         {
             ASSERT(message.size() == 1);
-            SetPerfScopeResampler(std::move(message[0]));
+            SetPerfScopeResampler(std::string(message[0]));
         });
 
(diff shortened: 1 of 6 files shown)


jd28:
Did a small test for anyone else that might be curious about what happens: https://gist.github.com/jd28/3655cb34be3786ee8345a277e5e6c1d2

Plugins/Profiler/Profiler.cpp
```diff
@@ -148,14 +148,14 @@ Profiler::Profiler(const Plugin::CreateParams& params)
 
     {
         GetServices()->m_messaging->SubscribeMessage("NWNX_PROFILER_SET_PERF_SCOPE_RESAMPLER",
-        [this, sum](std::vector<std::string> message)
+        [this, sum](const std::vector<std::string>& message)
         {
             ASSERT(message.size() == 1);
-            SetPerfScopeResampler(std::move(message[0]));
+            SetPerfScopeResampler(std::string(message[0]));
```
mtijanic, line 154: Should this not be just `message[0]`? Why force a copy?

jd28:
Underlying function call has a std::string&&, I'll change that.

jd28:
Another idea as it relates to this and some of the thoughts on events would be to move Argument out of events (possibly add a userdefined `void*`), change `Message` to `std::vector<Argument>`.  Then there would be one simple concrete type for NWN <-> Core and plugin <-> plugin communication without all the string conversions.

PR #924, opened 2020-05-28
Argument{Stack} Breakout

jd28:
Follows on https://github.com/nwnxee/unified/pull/917 I will reference this in an issue so as to tie some ideas together. **No** plugin level code is harmed or needs to be changed by this PR.


NWNXLib/CMakeLists.txt
```diff
@@ -16,7 +16,7 @@ macro(nwnxlib_add)
     endif()
 endmacro()
 
-nwnxlib_add("Assert.cpp" "Log.cpp" "Plugin.cpp" "Serialize.cpp" "Utils.cpp" "Encoding.cpp")
+nwnxlib_add("Assert.cpp" "Log.cpp" "Plugin.cpp" "Serialize.cpp" "Utils.cpp" "Encoding.cpp" NWVariant.cpp)
```
mtijanic, line 19: Bikeshed: I find the name confusing. I'd like to treat anything `C[A-Z]\w+` or `NW\w+` as base game reserved. I understand this is in NWNXLib namespace, but most code is `using` that already.

How about just `Variant`? Or `GameType`, `ScriptType`, since it holds any of the times visible to nwscript? `ScriptVariant` as a combo? etc..

NWNXLib/NWVariant.cpp
```diff
@@ -0,0 +1,28 @@
+#include "NWVariant.hpp"
+#include "Utils.hpp"
+
+namespace NWNXLib
+{
+
+std::string NWVariant::toString() const
```
mtijanic, line 7: The header file is already 4x the cpp file, and this is almost a trivial function. I'd just collapse it into the header and get rid of the cpp entirely - we don't gain much from this, it just splits code further.

NWNXLib/CMakeLists.txt
jd28: I didn't really like the name either.  I like `ScriptVariant`, I think that's the clearest of all the options.

NWNXLib/NWVariant.cpp
jd28: Will do.

PR #927, opened 2020-05-29
[nwnxlib/plugin, services/plugins] Add check for build revision

jd28:
* Change m_targetVersion to s_targetBuild, make it a static member (I can't see anyone wanting to pass in that value)
* Add s_targetBuildReisvion cause the addresses aren't stable among build revisions and if they don't match in a hook bad times will be had.
* Check when the plugin is loaded that both build and revision match.
* Get rid of m_flags, it's never used, not obvious how it would be used.  Easier to add it back if a need arose than having it and trying to figure out what it's supposed to be used for.
* Remove some pointless public access specifiers on a struct
* Remove the `Version` alias.  Maybe there was going to be a bigger idea here, but 90%+ of the plugins are still version 1 anyway.

I imagine all plugins are in tree or downstream in tree so these changes really wouldn't make a ton of difference.

### DIFF
--- a/NWNXLib/Plugin.cpp
+++ b/NWNXLib/Plugin.cpp
@@ -15,18 +15,14 @@ Plugin::Info::Info(std::string name,
     std::string description,
     std::string author,
     std::string contact,
-    const Version version,
-    const bool hotswap,
-    const uint32_t flags,
-    const Version targetVersion)
+    const uint32_t version,
+    const bool hotswap)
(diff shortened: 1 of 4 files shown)


Daztek:
I think @mtijanic is planning on removing the whole PluginInfo struct in his spring cleaning.

jd28:
That sounds like an even better idea!  Thank you for the heads up.  I'll go ahead and close this.

PR #929, opened 2020-05-30
scripts: Add bootstrap_plugin.sh

jd28:
I can't imagine anyone not just copying and pasting from an existing plugin, might as well formalize it.  Feedback welcome.

I didn't put any effort into Plugin::Info since Daz mentioned in https://github.com/nwnxee/unified/pull/927 that it's going to be spring cleaned.


mtijanic:
Cool! Will need some tweaks as we simplify plugins, but it's a good idea!

PR #940, opened 2020-06-15
Player: restore associate type after possession

weavejester:
If a summon or other non-familiar associate is possessed, the associate type should be reset back to its previous value when unpossessed. This prevents the PC from losing their associate from their party.

This commit will store the possessed creature's associate type, and if it has a type (i.e. is not zero), it will re-add the creature as that associate type to the possessor.

### DIFF
--- a/Plugins/Player/Player.cpp
+++ b/Plugins/Player/Player.cpp
@@ -1249,10 +1249,20 @@ ArgumentStack Player::PossessCreature(ArgumentStack&& args)
                             pPossessor->RemoveAssociate(possessedOidPOS);
                             pPOS->Remove(pPossessor->m_idSelf, "possessedOid");
                             pPOS->Remove(possessedOidPOS, "possessorOid");
+
+                            auto possessedAssociateType = *pPOS->Get<int>(pPossessor->m_idSelf, "possessedAssociateType");
+                            if (possessedAssociateType)
+                            {
+                                pPossessor->AddAssociate(possessedOidPOS, possessedAssociateType);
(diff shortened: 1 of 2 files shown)


Plugins/Player/Player.cpp
```diff
@@ -1249,11 +1249,21 @@ ArgumentStack Player::PossessCreature(ArgumentStack&& args)
                             pPossessor->RemoveAssociate(possessedOidPOS);
                             pPOS->Remove(pPossessor->m_idSelf, "possessedOid");
                             pPOS->Remove(possessedOidPOS, "possessorOid");
+
+                            auto possessedAssociateType = *pPOS->Get<int>(pPossessor->m_idSelf, "possessedAssociateType");
```
mtijanic, line 1253: From https://en.cppreference.com/w/cpp/utility/optional/operator* :
> The behavior is undefined if `*this` does not contain a value. 

I guess it just default-constructed it, but better to clear it up. I'd write this as:

```C
auto possessedAssociateType = pPOS->Get<int>(...);
if (possessedAssociateType && *possessedAssociateType != 0)
{
    pPossessor->AddAssociate(possessedOidPOS, *possessedAssociateType);
    ...
}
```

mtijanic: You can also do:
```C
auto possessedAssociateType = pPOS->Get<int>(...).value_or(0);
```

weavejester: Thanks for the review. I'll update the PR tomorrow.

PR #946, opened 2020-06-22
Creature: add NWNX_Creature_SetMovementRateFactorCap

weavejester:
This PR adds functionality for upping the movement rate cap on creatures, which is usually capped at 1.5 base speed.

I've tested this in a sandbox environment, but I'm going to put it on my test server and see how it shakes out in order before I'm completely confident that there isn't any weirdness around it. To this end, I've added a WIP, though as far as I know the code is done.

### DIFF
--- a/Plugins/Creature/Creature.cpp
+++ b/Plugins/Creature/Creature.cpp
@@ -107,6 +107,7 @@ Creature::Creature(const Plugin::CreateParams& params)
     REGISTER(SetMovementRate);
     REGISTER(GetMovementRateFactor);
     REGISTER(SetMovementRateFactor);
+    REGISTER(SetMovementRateFactorCap);
     REGISTER(SetAlignmentGoodEvil);
     REGISTER(SetAlignmentLawChaos);
     REGISTER(SetDomain);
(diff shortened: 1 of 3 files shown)


weavejester:
Having tested this without issue for a week, I'm removing the WIP tag.

mtijanic:
What's the actual need behind this function? Do you actually want a different cap, or do you want to be able to set a movement rate factor that exceeds the cap?

weavejester:
This allows movement speed increase effects from spells, equipment and other sources to stack over 150%. 

So for example, say on one server barbarians have a natural 20% movement speed bonus, so their base movement speed is 120%. If they were to use expeditious retreat, which gives a 50% movement bonus, then you'd expect their speed to be 120+50=170% of base. However, all classes aside from monks are capped at 150%, so a barbarian using expeditious retreat would be the same 150% movement speed as anyone else.

To give another example, on our server we've given all characters a "Dash" instant use feat that increases movement speed by 50% + their athletics skill for one round, and we want this to stack with spells and other sources of movement increases. If a figher with 10 athletics uses dash *and* expeditious retreat, then they can close gaps at 210% movement speed.

I've tried other ways of doing this, including intercepting events and trying to manually calculate the movement speed, and it's a huge hassle to account for stealth/detect/slow/overburdening etc. that can affect movement speed. This is a far simpler solution, and so far has worked fine. All character on our server now have a 300% movement cap.

mtijanic:
Would it be better if SetMovementRateFactor would just ignore the 150% cap and work with any value? Maybe an additional arg to the function (`bIgnoreCap`)?

mtijanic:
It might be as simple as: 
```diff
--- /home/mtijanic/nwn/nwnx/Plugins/Creature/Creature.cpp   Wed Jul  1 10:17:09 2020
+++ /home/mtijanic/nwn/nwnx/Plugins/Creature/Creature.cpp   Sun Jul  5 02:36:31 2020
@@ -1069,7 +1069,7 @@
     if (auto *pCreature = creature(args))
     {
         const float factor = Services::Events::ExtractArgument<float>(args);
-        pCreature->SetMovementRateFactor(factor);
+        pCreature->m_fMovementRateFactor = factor;
     }
     return Services::Events::Arguments();
 }
```

mtijanic:
Or maybe it should be a tweak? `NWNX_TWEAKS_IGNORE_MOVEMENT_RATE_FACTOR_CAPS=y`?

weavejester:
So I think there should still be a cap, since there may be some combination of spells or abilities that allow movement speed to be increased indefinitely. I'd rather not see how NWN copes with a 10000000% increase in speed, for instance. In our case, we've chosen 300% as that's the cap on base-NWN monks, so NWN should be able to handle that no problem.

mtijanic:
Fair enough.

PR #955, opened 2020-06-29
Added Player: SetCreatureNameOverride()

julien-lecomte:
Following short discussion on discord.

Adds a Player function that does what it's name says.


mtijanic:
Wouldn't it be better to collapse this into a single SetObjectNameOverride that works on all object types?

julien-lecomte:
Probably. And deprecate the others.

mtijanic:
Let's merge this now and then combine afterwards..

PR #966, opened 2020-07-23
Fix to prevent unlimited uses of potions/scrolls

GoLoT:
Disclaimer: Use at your own risk.

Demo: https://www.youtube.com/watch?v=NeDrCFpZ090

Tested in 8193.12, it should work in newer versions.

It removes the item instantly right before the effect is applied, preventing users from triggering the bug by un/stacking items. I don't think it should cause any issues because when the item is removed the info needed by the spell script is stored already and the item/oid isn't used anywhere as far as I can tell, other than to signal the source of the spell.

I thought about moving the item somewhere so the player can't interact with it and destroying it after the spell is cast, but I don't think it's needed. If the fix causes any issues this alternative approach might work.

### DIFF
--- a/Plugins/Tweaks/Tweaks/FixUnlimitedPotionsBug.cpp
+++ b/Plugins/Tweaks/Tweaks/FixUnlimitedPotionsBug.cpp
@@ -1,0 +1,52 @@
+#include "Tweaks/FixUnlimitedPotionsBug.hpp"
+
+#include "Services/Hooks/Hooks.hpp"
+#include "Utils.hpp"
+
+#include "API/C2DA.hpp"
+#include "API/CAppManager.hpp"
+#include "API/CExoString.hpp"
+#include "API/CNWSCombatRound.hpp"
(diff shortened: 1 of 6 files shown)


Plugins/Tweaks/Tweaks/FixUnlimitedPotionsBug.cpp
```diff
@@ -0,0 +1,36 @@
...
+#include "Utils.hpp"
+
+#include "API/Functions.hpp"
+
+namespace Tweaks {
+
+using namespace NWNXLib;
+using namespace NWNXLib::API;
+
+bool s_bUsableItemRemoval = false;
```
mtijanic, line 13: If it is `s_`, it should be `static`.

GoLoT: Done. Is a static global OK or should I make it a static member instead?

PR #969, opened 2020-07-25
Unhardcode shield AC using baseitems.2da

GoLoT:
Small change to use values from baseitems.2da to determine shield AC instead of the hardcoded values.

Alternatively the whole switch can be replaced by `nAC = pBaseItem->m_nBaseAC;` since all non-shield items have a BaseAC value of 0 and even if there is an item with an armor value it should be ignored by the engine, I'll leave the baseitem check just to be safe though.

P.S.: Ninja edited changelog.md to add `NWNX_TWEAKS_FIX_UNLIMITED_POTIONS_BUG` because I forgot to add it in its PR.


mtijanic:
I feel like it should be  `nAC = pBaseItem->m_nBaseAC` (with the armor exception). I think that might allow you to have shield-like items

Arvirago:
Thanks GoLoT!!

GoLoT:
> I feel like it should be `nAC = pBaseItem->m_nBaseAC` (with the armor exception). I think that might allow you to have shield-like items

There are a couple more places where shield checks happen. There is one in `CNWSCreature::ComputeArmourClass()` that stores/removes the shield armor value for the creature (m_nACShieldBase) when a shield is un/equipped. Then there is a check in `CNWSCreature::CanEquipItem()` to make sure that you can equip a shield. It will run `CNWSItem::CanEquipWeapon()` instead but I don't think a change is needed, we can ignore it. It will equip the item as an offhand weapon with shield AC and should work fine, otherwise the Events plugin can be used to script a custom check. And there may be some other checks somewhere that I missed but those won't affect the AC calculations so those can probably be ignored too.

I think changing `CNWSCreature::ComputeArmourClass()` should be enough to have shield-like items, otherwise the armor value is completely ignored by the engine and never applied. I'll give it a shot.

GoLoT:
Done. The new changes should allow creating new shield-like items by setting a BaseAC value for them on baseitems.2da.
Alternatively a check for `WeaponWield == 7` can be added so it only works with items explicitly defined as shields, but I don't see the need since all non-shield items will have a BaseAC of 0 and this will allow using off-hand weapons as blocking weapons (i.e. a dagger with +2 AC when equipped off-hand).

Demo: https://www.youtube.com/watch?v=t5iH8yZlyJ0
(Tower shield BaseAC: 4; Torch BaseAC: 5)

Issue #957, opened 2020-07-08
[Docs] Incorrect docs in comment of NWNX_ON_VALIDATE_ITEM_EQUIP_*


weavejester:
The tag for the variable should be `ITEM_OBJECT_ID` instead of `ITEM`:

https://github.com/nwnxee/unified/blob/87f1663d110c013b6e34b19cae2ef15922b709ac/Plugins/Events/NWScript/nwnx_events.nss#L127

mtijanic:
Fixed in 0d65c36befa04c0cc320de16a1b1d0d340e796ee

(was it really easier to open a bug then just fix through github web editor? :stuck_out_tongue: )

weavejester:
I completely forgot GitHub has a web editor 😃

PR #974, opened 2020-07-28
documentation correction

hendrikgit:
@mtijanic in https://github.com/nwnxee/unified/issues/957 you said this was fixed by a PR I made but it wasn't. Now it is though, so your plan worked ;)


mtijanic:
I definitely remember looking at it and seeing it already fixed. :shrug:

PR #977, opened 2020-07-30
Update Awesome-NWNX.md

shadguy:
Add link to nwscript basics doc under nwnsc


mtijanic:
Thanks! I've updated it to be a more descriptive link and moved it to a top level section under 'other'

PR #979, opened 2020-07-31
Add spawn point encounter functions

julien-lecomte:
I had a toolset crash everytime I wanted to do this:

```
/// @brief A spawn point list entry for an encounter.
struct NWNX_Encounter_SpawnPoint
{
    vector vec; ///< The vector.
    float orientation; ///< The orientation.
};
```

So I didn't take the high road and used x,y,z directly.

### DIFF
--- a/Plugins/Encounter/Encounter.cpp
+++ b/Plugins/Encounter/Encounter.cpp
@@ -3,6 +3,7 @@
 #include "API/CAppManager.hpp"
 #include "API/CServerExoApp.hpp"
 #include "API/CNWSEncounter.hpp"
+#include "API/CEncounterSpawnPoint.hpp"
 #include "API/CEncounterListEntry.hpp"
 #include "API/Constants.hpp"
 #include "API/Globals.hpp"
@@ -39,6 +40,11 @@ Encounter::Encounter(Services::ProxyServiceList* services)
     REGISTER(SetPlayerTriggeredOnly);
     REGISTER(GetResetTime);
     REGISTER(SetResetTime);
+    REGISTER(GetNumberOfSpawnPoints);
+    REGISTER(GetSpawnPointByIndex);
+    REGISTER(GetMinNumSpawned);
+    REGISTER(GetMaxNumSpawned);
+    REGISTER(GetCurrentNumSpawned);
 
 #undef REGISTER
 }
@@ -88,6 +94,7 @@ ArgumentStack Encounter::GetEncounterCreatureByIndex(ArgumentStack&& args)
     if (auto *pEncounter = encounter(args))
     {
         const auto index = Services::Events::ExtractArgument<int32_t>(args);
+        ASSERT_OR_THROW(index >= 0.0);
 
         if (index < pEncounter->m_nNumEncounterListEntries)
         {
@@ -107,7 +114,7 @@ ArgumentStack Encounter::SetEncounterCreatureByIndex(ArgumentStack&& args)
         const auto index = Services::Events::ExtractArgument<int32_t>(args);
         const auto resRef = Services::Events::ExtractArgument<std::string>(args);
         auto cr = Services::Events::ExtractArgument<float>(args);
-          ASSERT_OR_THROW(cr >= 0.0);
+        ASSERT_OR_THROW(cr >= 0.0);
         auto unique = Services::Events::ExtractArgument<int32_t>(args);
         unique = !!unique;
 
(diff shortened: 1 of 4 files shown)


julien-lecomte:
Maybe rename `NWNX_Encounter_SpawnPoint` to `NWNX_Encounter_SpawnPointEntry` so that it's like creature list.

julien-lecomte:
Fixed the previous comment by returning a proper `location`

Plugins/Encounter/Encounter.cpp
```diff
@@ -200,4 +203,36 @@ ArgumentStack Encounter::SetResetTime(ArgumentStack&& args)
...
+    return Services::Events::Arguments(retVal);
+}
+
+ArgumentStack Encounter::GetSpawnPointByIndex(ArgumentStack&& args)
+{
+    float x = 0.0, y = 0.0, z = 0.0, o = 0.0;
+
+    if (auto *pEncounter = encounter(args))
+    {
+        const auto index = Services::Events::ExtractArgument<int32_t>(args);
```
Daztek, line 224: Should add an `ASSERT_OR_THROW(index >= 0);` here.

If you're feeling fancy, add one to `GetEncounterCreatureByIndex` too, which *someone* seems to have forgotten.

Daztek:
Oh also, don't forget to update the changelog!

Plugins/Encounter/NWScript/nwnx_encounter.nss
```diff
@@ -162,3 +173,29 @@ void NWNX_Encounter_SetResetTime(object encounter, int resetTime)
...
+  NWNX_PushArgumentInt(NWNX_Encounter, sFunc, index);
+  NWNX_PushArgumentObject(NWNX_Encounter, sFunc, encounter);
+  NWNX_CallFunction(NWNX_Encounter, sFunc);
+
+  float o = NWNX_GetReturnValueFloat(NWNX_Encounter, sFunc);
+  float z = NWNX_GetReturnValueFloat(NWNX_Encounter, sFunc);
+  float y = NWNX_GetReturnValueFloat(NWNX_Encounter, sFunc);
+  float x = NWNX_GetReturnValueFloat(NWNX_Encounter, sFunc);
+
+  return Location(GetArea(encounter), Vector(x, y, z), o);
```
Daztek, line 200: 4 spaces of indentation here, please!

Plugins/Encounter/Encounter.cpp
```diff
@@ -192,12 +199,78 @@ ArgumentStack Encounter::SetResetTime(ArgumentStack&& args)
...
 
         pEncounter->m_nResetTime = resetTime;
     }
 
     return Services::Events::Arguments();
 }
 
+ArgumentStack Encounter::GetNumberOfSpawnPoints(ArgumentStack&& args)
+{
+    int32_t retVal = 0;
```
mtijanic, line 212: Wouldn't `-1` be a better error return value, since 0 is a valid value?

Plugins/Encounter/Encounter.cpp
```diff
@@ -192,12 +199,78 @@ ArgumentStack Encounter::SetResetTime(ArgumentStack&& args)
...
+    {
+        const auto index = Services::Events::ExtractArgument<int32_t>(args);
+        ASSERT_OR_THROW(index >= 0);
+
+        if (index < pEncounter->m_nNumSpawnPoints)
+        {
+            x = pEncounter->m_pSpawnPointList[index].m_vPosition.x;
+            y = pEncounter->m_pSpawnPointList[index].m_vPosition.y;
+            z = pEncounter->m_pSpawnPointList[index].m_vPosition.z;
+            o = pEncounter->m_pSpawnPointList[index].m_fOrientation;
```
mtijanic, line 236: I wonder if orientation actually does anything here. And if not, you could just return a `vector` instead of `location`. No big deal either way..

Plugins/Encounter/Encounter.cpp
```diff
@@ -192,12 +199,78 @@ ArgumentStack Encounter::SetResetTime(ArgumentStack&& args)
...
+    {
+        retVal = pEncounter->m_nMaxSpawnedCreatures;
+    }
+
+    return Services::Events::Arguments(retVal);
+}
+
+ArgumentStack Encounter::GetCurrentNumSpawned(ArgumentStack&& args)
+{
+    int32_t retVal = 0;
```
mtijanic, line 267: Above 3 as well, -1 seems like a better error value.

Plugins/Encounter/Encounter.cpp
julien-lecomte: Yes it would, but it's contrary to 99% of the NWN stock functions that don't let you differenciate 0 from ERROR.

Plugins/Encounter/Encounter.cpp
julien-lecomte: I have no clue if orientation is used by the NWN engine, but a user using this API can use the orientation if he wishes.

Plugins/Encounter/Encounter.cpp
julien-lecomte: For example: Encounter::GetResetTime returns 0 on error. It could return -1 to indicate error.

PR #981, opened 2020-08-01
Player: add FloatingTextStringOnCreature()

Daztek:


Plugins/Player/Player.cpp
```diff
@@ -1428,4 +1430,30 @@ ArgumentStack Player::SetCreatureNameOverride(ArgumentStack&& args)
...
+        if (auto *pCreature = Utils::AsNWSCreature(Utils::GetGameObject(oidCreature)))
+        {
+            if (auto *pMessage = static_cast<CNWSMessage*>(Globals::AppManager()->m_pServerExoApp->GetNWSMessage()))
+            {
+                auto *pMessageData = new CNWCCMessageData();
+                pMessageData->SetObjectID(0, pCreature->m_idSelf);
+                pMessageData->SetInteger(9, 94);
+                pMessageData->SetString(0, text);
+
+                pMessage->SendServerToPlayerCCMessage(pPlayer->m_nPlayerID, 11, pMessageData, nullptr);
```
Daztek, line 1451: Not sure if `pMessageData` needs a `delete` after sending it.

Daztek: Digging around some more, it appears it does need one.

GoLoT: Read in clippy's voice: If it's deleted when it goes out of this scope, you can stack allocate with `CNWCCMessageData messageData;` instead of using a raw pointer.

Daztek: Oh, that's even better, thanks!

Plugins/Player/Player.cpp
```diff
@@ -1428,4 +1430,30 @@ ArgumentStack Player::SetCreatureNameOverride(ArgumentStack&& args)
     return Services::Events::Arguments();
 }
 
+ArgumentStack Player::FloatingTextStringOnCreature(ArgumentStack&& args)
+{
+    if (auto *pPlayer = player(args))
+    {
+        auto oidCreature = Services::Events::ExtractArgument<ObjectID>(args);
+          ASSERT_OR_THROW(oidCreature != Constants::OBJECT_INVALID);
```
julien-lecomte, line 1438: Is there a reason this is indented by 2 extra? I un-indented this in my MR.

mtijanic: This half-indent was used for some of these ASSERT_OR_THROW calls to make it more obvious it is relevant to the line above. Originally, the asserts were in the same line as extract, after the `;`, but that was unwieldy so this was done.

I don't really feel strongly either way. It _does_ look slightly nicer to me, but it is also another thing to remember and it's just not worth the cognitive load to enforce.

PR #982, opened 2020-08-06
Events: allow any integer result in equip validation

### LINKED ISSUE #958: [Bug] Equipping armor after validation overridden.
With the `NWNX_ON_VALIDATE_ITEM_EQUIP_AFTER` event, we can override the default item equip validation, allowing items to be equipped that could not normally be. For example, a character without heavy armor proficiency can be allowed equip a full plate item.

```
#include "nwnx_events"
#include "nwnx_object"

void main()
{
    object oItem = NWNX_Object_StringToObject(NWNX_Events_GetEventData("ITEM_OBJECT_ID"));
    int    nSlot = StringToInt(NWNX_Events_GetEventData("SLOT"));

    if (nSlot == INVENTORY_SLOT_CHEST && GetTag(oItem)== "SPECIAL_FULL_PLATE")
    {
        NWNX_Events_SetEventResult("1");
    }
}
```

However, this results in strange behavior when equipping the special full plate, and there is already another armor item in the chest slot. The existing armor won't unequip, but will be erased or overwritten. My guess is that this is either because the equip check is side-effectful and also unequips the armor if necessary, or the unequipping of existing armor is determined by another function that isn't overridden.

I've tried a couple of workarounds. The first workaround was to copy the object that was replaced, but this continually increases the weight carried by the creature, since the weight of the overwritten armor seems to still apply, at least until the player relogs.

```
object oExistingItem = GetItemInSlot(INVENTORY_SLOT_CHEST);

if (oExistingItem != OBJECT_INVALID)
{
    CopyItem(oExistingItem, OBJECT_SELF, TRUE);
    DestroyObject(oExistingItem);
}
```

The second workaround was just to reject the item and then force an unequip. This works, but isn't instantaneous.

```
if (oExistingItem != OBJECT_INVALID)
{
    ActionUnequipItem(oExistingItem);
    ActionEquipItem(oItem, INVENTORY_SLOT_CHEST);
    return;
}
```

weavejester:
This fixes #958 by allowing the `NWNX_ON_VALIDATE_ITEM_EQUIP_AFTER` event to return any integer, rather than just `0` or `1`.

Ideally we could be a little smarter and automate the process, since it appears that a value of `2` means "replace the current item in the slot" and a value of `3` means "replace the current main and offhand weapon in the slot". More experimentation of whether this is the right interpretation is probably needed.

In my case, I manually check in NWScript whether the chest slot is occupied, and return `1` or `2` depending on that.

However, while this PR could be improved, this at least makes it possible to use the validate equip event on occupied slots, and may do as an interim solution.

### DIFF
--- a/Plugins/Events/Events/ItemEvents.cpp
+++ b/Plugins/Events/Events/ItemEvents.cpp
@@ -329,7 +329,7 @@ int32_t ItemEvents::CanEquipItemHook(CNWSCreature* thisPtr, CNWSItem* pItem, uin
     Events::PushEventData("BEFORE_RESULT", std::to_string(retVal));
     Events::SignalEvent("NWNX_ON_VALIDATE_ITEM_EQUIP_AFTER", thisPtr->m_idSelf, &sAfterEventResult);
 
-    retVal = sAfterEventResult.empty() ? retVal : sAfterEventResult == "1";
+    retVal = sAfterEventResult.empty() ? retVal : std::stoi(sAfterEventResult);
 
     return retVal;
 }


mtijanic:
You're right about the meaning of 0-3. There's no other values. It would be nice to document this in the nwnx_events.nss too (and changelog.md)

PR #983, opened 2020-08-08
Build8193.14 support

mtijanic:
..Still need to update readme and changelog..

..and a ton of testing..


mtijanic:
I _think_ this is ready now, and dedicated server package is also up (albeit with libsndio.so dependency).

I'll leave the PR up for 12 more hours in case someone decides to test it further and then merge and tag an official build.

PR #992, opened 2020-08-12
Add generic Push/Pop/Free variants for game defined structs (sqlquery type support)

Jorteck:
There doesn't seem to be a need to have these split up, and it would make future engine type additions have less breaking effects for DotNet projects.


Plugins/DotNET/DotNETExports.cpp
```diff
@@ -176,27 +137,48 @@ void DotNET::StackPushVector(Vector value)
 
 void DotNET::StackPushEffect(CGameEffect* value)
 {
-    PushedCount += StackPushGameDefinedStructure(0, value);
+    LOG_WARNING("StackPushEffect is deprecated. Use StackPushGameDefinedStructure instead.");
+    StackPushGameDefinedStructure(0, value);
```
Daztek, line 141: There's constants for VM Structures here: https://github.com/nwnxee/unified/blob/master/NWNXLib/API/Constants/VirtualMachine.hpp#L32

You'll have to add sqlquery tho.

Plugins/DotNET/DotNETExports.cpp
```diff
@@ -176,27 +137,48 @@ void DotNET::StackPushVector(Vector value)
 
 void DotNET::StackPushEffect(CGameEffect* value)
 {
-    PushedCount += StackPushGameDefinedStructure(0, value);
+    LOG_WARNING("StackPushEffect is deprecated. Use StackPushGameDefinedStructure instead.");
```
mtijanic, line 140: If we're deprecating it, we should update the managed code to match. I think we can make it a breaking change though and remove all these.

PR #1000, opened 2020-08-16
Util: add GetWorldTime()

Daztek:


Daztek:
Renamed to GetWorldTime() since you can use it for other time related stuff too, such as timestamps that adhere pause/timestop

PR #1002, opened 2020-08-17
Util: update GetWorldTime() to allow negative adjustments

Daztek:


ELadner:
How is the time represented as a single value?

Daztek:
It returns two, a struct with a calendar day value and a time of day value.

PR #1004, opened 2020-08-17
Events: add Request{Buy|Sell} events to StoreEvents

### LINKED ISSUE #991: [Feature Request] Add support for blocking store transactions from taking place (buying, selling items to stores)
It would be nice to be able to stop a store transaction from happening. Right now it's kind of hard to limit different kinds of items with same base item type from being sold to stores and dealing with items after the fact is kind of messy. Also there's the issue of selling unidentified items to stores making them identified.

Daztek:
Resolves #991


mtijanic:
Might be useful to also expose `pStore->CalculateItemSellPrice(pItem, pCreature->m_idSelf)` somehow. Either as an extra event data, or even a standalone funciton.

PR #1015, opened 2020-08-22
Encounters: Allow to tag a creature as already used

julien-lecomte:


mtijanic:
This will require a recompile + nss update for everyone using it, but the alternative would be even more hassle, so.. good enough.

julien-lecomte:
I wouldn't have minded if I were to keep this on a branch of my own on top of master until you decide it's time for a new release.

Also, edited (and fixed) my sample nss code.

PR #1024, opened 2020-09-05
Weaponpluginmisc

Morderon:
1) Good AIIM SLING SWITCH NWNX_WEAPON_GOOD_AIM_SLING if true allows non-halflings to gain +1 AB with slings if they have good AIM. This is a correction for the base game which includes throwing weapons but tied slings specifically to the halfling race.

Developer Note: I wanted to do the above as a tweak but fitting function had an exclusive hook in nwnx weapons

2) New function (SetOneHalfStrength) gives a melee weapon an addition Strength/2 damage. For use with One and a half weapons to signify holding them two handed.

Modifications:

1) Weapon feats (focus, spec, etc.) now support having multiple feats for each weapon

2) The weapon feats defined in baseitems.2da serves as a fallback instead of being overriden

### DIFF
--- a/Plugins/Weapon/Weapon.cpp
+++ b/Plugins/Weapon/Weapon.cpp
@@ -53,6 +53,8 @@ Weapon::Weapon(Services::ProxyServiceList* services)
     REGISTER(GetEventData);
     REGISTER(SetEventData);
     REGISTER(SetOneHalfStrength);
+    REGISTER(GetOneHalfStrength);
+
 #undef REGISTER
 
     m_GetWeaponFocusHook = GetServices()->m_hooks->RequestExclusiveHook<Functions::_ZN17CNWSCreatureStats14GetWeaponFocusEP8CNWSItem>(&Weapon::GetWeaponFocus);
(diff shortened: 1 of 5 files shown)


Plugins/Weapon/Weapon.cpp
```diff
@@ -980,4 +1193,22 @@ int Weapon::GetLevelByClass(CNWSCreatureStats *pStats, uint32_t nClassType)
...
+
+    if(objectId == Constants::OBJECT_INVALID)
+    {
+        LOG_INFO("Invalid Object Passed into SetOneHalfStrength");
+        return Services::Events::Arguments();
+    }
+
+    auto bMulti = Services::Events::ExtractArgument<int32_t>(args);
+    if(bMulti)
+        g_plugin->GetServices()->m_perObjectStorage->Set(objectId, "ONE_HALF_STRENGTH", 1);
```
Daztek, line 1208: If you want, you can change the above to `g_plugin->GetServices()->m_perObjectStorage->Set(objectId, "ONE_HALF_STRENGTH", 1, true);` and it'll persist the value when the item gets saved, so you only have to set it once.

PR #1027, opened 2020-09-11
Damage: allow placeables to trigger damage script

weavejester:
This is useful for making placeables that react to certain damage types. For example, barrels that explode when hit with fire or electricity.


Daztek:
Should probably make a note about it in the changelog :D

weavejester:
Agh, I always forget that. Hang on...

weavejester:
Updated the changelog

PR #1028, opened 2020-09-11
Effect - New functions

Morderon:
UnpackEvent - now displays ID, useful to tell linked applied effects. Additional notes: Linked effects that have been applied to a PC no longer report their left/right links, making their id the only way to tell that they are linked.

New functions:

GetTrueEffect(Count) - Use to loop through the effects (including those normally hidden, such like those from the race plugin or item properties)

RemoveEffectById - In case you want to remove the effect but all you have to work with is it's Id

SetEffectImmunityBypass - This bypasses effect level immunity, such as when you use an ApplyEffect Paralyze on a PC and immunity to mind spells blocks it.

ReplaceEffectByElement - A more specific ReplaceEffect as you can single out which effect to change.


Plugins/Effect/NWScript/nwnx_effect.nss
```diff
@@ -90,16 +91,40 @@ object NWNX_Effect_GetEffectExpiredCreator();
 /// @return Number of internal effects updated.
 int NWNX_Effect_ReplaceEffect(object obj, effect eOld, effect eNew);
 
-/// @}
+/// @brief Gets the true effect count
+/// @param oObjece The object to get the count of.
+/// @return the number of effects (item propertiies and other non-exposed effects included)
```
plenarius, line 96: Just some minor typos on this line: `properties` and previous line `oObject`

Plugins/Effect/Effect.cpp
```diff
@@ -285,4 +306,142 @@ ArgumentStack Effect::ReplaceEffect(ArgumentStack&& args)
...
+        if(auto *pObject = Utils::AsNWSObject(Globals::AppManager()->m_pServerExoApp->GetGameObject(objectId)))
+        {
+            auto element = Services::Events::ExtractArgument<int32_t>(args);
+              ASSERT_OR_THROW(element >= 0);
+              ASSERT_OR_THROW(element < pObject->m_appliedEffects.num);
+            auto eff = pObject->m_appliedEffects.element[element];
+
+            eff->m_sCustomTag = Services::Events::ExtractArgument<std::string>(args).c_str();
+
+            auto vector1z = Services::Events::ExtractArgument<float>(args);
```
mtijanic, line 356: Shouldn't this lot be in a ResolvePack() function or equivalent?

Plugins/Effect/Effect.cpp
```diff
@@ -285,4 +306,142 @@ ArgumentStack Effect::ReplaceEffect(ArgumentStack&& args)
     return found;
 }
 
+ArgumentStack Effect::GetTrueEffectCount(ArgumentStack&& args)
+{
+    int32_t retVal = 0;
+    auto objectId = Services::Events::ExtractArgument<ObjectID>(args);
+
+    if(objectId != Constants::OBJECT_INVALID)
```
mtijanic, line 314: Nit: no need for this check, `GetGameObject(OBJECT_INVALID)` will return nullptr, and AsNWSObject() of that will also be null. Same in other instances of this pattern.

Plugins/Effect/Effect.cpp
```diff
@@ -285,4 +306,142 @@ ArgumentStack Effect::ReplaceEffect(ArgumentStack&& args)
...
+    int32_t ret = 0;
+    auto objectId = Services::Events::ExtractArgument<ObjectID>(args);
+    if(objectId != Constants::OBJECT_INVALID)
+    {
+        if (auto *pObject = Utils::AsNWSObject(Globals::AppManager()->m_pServerExoApp->GetGameObject(objectId)))
+        {
+            auto id = Services::Events::ExtractArgument<std::string>(args);
+            auto it = std::find_if(id.begin(), id.end(), [](unsigned char c) { return !std::isdigit(c); });
+            if(!id.empty() && it == id.end())
+                ret = pObject->RemoveEffectById(std::stoi(id));
```
mtijanic, line 418: `std::stoi` converts it to a 32bit integer, but the effect ID is 64bit. You want `std::stol()`.
Also, there is no need to sanity check it with `std::isdigit` because the conversion functions will throw an exception if the conversion doesn't go smoothly, which NWNX will already catch and print.

Plugins/Effect/NWScript/nwnx_effect.nss
```diff
@@ -90,16 +91,40 @@ object NWNX_Effect_GetEffectExpiredCreator();
...
+
+/// @brief Removes effect by ID
+/// @param oObject The object to remove the effect from
+/// @param sID The id of the effect, can be retrieved by unpacking effects.
+/// @return FALSE/0 on failure TRUE/1 on success.
+int NWNX_Effect_RemoveEffectById(object oObject,  string sID);
+
+/// @brief Disables effect-level immunities.
+/// @param nEnable set to true to turn on.
+void NWNX_Effect_SetEffectImmunityBypass(int nEnable);
```
mtijanic, line 120: This seems better suited for a Tweak? Is this actually something you expect people to toggle on/off constantly at runtime?

Plugins/Effect/NWScript/nwnx_effect.nss
```diff
@@ -9,6 +9,7 @@ const string NWNX_Effect = "NWNX_Effect"; ///< @private
 /// An unpacked effect
 struct NWNX_EffectUnpacked
 {
+    string sID; ///< @todo Describe
```
mtijanic, line 12: This is an ABI breaking change, so it should be clearly noted in the changelog.

Plugins/Effect/NWScript/nwnx_effect.nss
```diff
@@ -134,10 +159,13 @@ struct NWNX_EffectUnpacked NWNX_Effect_UnpackEffect(effect e)
...
-    n.eLinkRight = NWNX_GetReturnValueEffect(NWNX_Effect, sFunc);
-    n.bLinkLeftValid = NWNX_GetReturnValueInt(NWNX_Effect, sFunc);
-    n.eLinkLeft = NWNX_GetReturnValueEffect(NWNX_Effect, sFunc);
+    if(bLink)
+    {
+        n.bLinkRightValid = NWNX_GetReturnValueInt(NWNX_Effect, sFunc);
+        n.eLinkRight = NWNX_GetReturnValueEffect(NWNX_Effect, sFunc);
+        n.bLinkLeftValid = NWNX_GetReturnValueInt(NWNX_Effect, sFunc);
+        n.eLinkLeft = NWNX_GetReturnValueEffect(NWNX_Effect, sFunc);
+    }
```
mtijanic, line 168: `else { n.bLinkRightValid = FALSE; n.bLinkLeftValid = FALSE; }` ?

Plugins/Effect/NWScript/nwnx_effect.nss
```diff
@@ -90,16 +91,40 @@ object NWNX_Effect_GetEffectExpiredCreator();
...
+struct NWNX_EffectUnpacked NWNX_Effect_GetTrueEffect(object oObject, int nElement);
 
-    NWNX_PushArgumentEffect(NWNX_Effect, sFunc, e);
-    NWNX_CallFunction(NWNX_Effect, sFunc);
+/// @brief Replaces an already applied effect with another.
+/// @param oObject The object with the effect to replace
+/// @param nElement The array element to be replaced
+/// @param e The unpacked effect to replace it with.
+/// @note Cannot replace an effect with a different type.
+void NWNX_Effect_ReplaceEffectByElement(object oObject, int nElement, struct  NWNX_EffectUnpacked e);
```
mtijanic, line 110: Naming nit: `ReplaceEffectByIndex`

mtijanic (commented):
Sorry it took so long to get to this. Few minor issues inside, but overall it looks really great.

The only important issue to work out before merging it in is whether the immunity bypass belongs here. For the rest, so long as the API is good (and it mostly is), further improvements can come later.

Plugins/Effect/NWScript/nwnx_effect.nss
Morderon: Answering this one new given it's more philosophical than coding, get to the coding in a couple days hopefully:

I entertained making this one a tweak when it was initially coded but decided against for 2 reasons:

1) Most (assumption there :P) nwnx users are working with established modules and likely have allowed immunities to levels they find satisfactory. But  may wish to allow only a few effects to bypass immunity. Instead of allowing all effects to bypass immunities and then having to go to each nwscript individually to add a GetHasEffectImmunity call.

2) Some immunities don't have clear events available on the nwscript side, some examples being sneak attack, crippling strike, critical hits, death attack - paralyze. Meaning if you disabled all immunities these would always proc; as long as their individual conditions are met.

2b) Some of these may be workable with nwnx events; but i'm more comfortable with saying "if you want crippling strike to be able to bypass immunity you can use xyz event with this function" instead of "if you want to return to original functionality without disabling the tweak for crippling strike, you can enable the events plugin and use xyz event and skip it. when these conditions are met"

That said i do think there's a flaw in it's current implementation, I've had it running in onEquip before and nwscript onequip (i didn't have a chance to try the after yet) and immunities weren't toggleing back on fast enough; some sneak attacks got through to creatures that should be immune and the like. So I'm thinking about allowing people to specify the object it's working for, or the effect, if not both (and likely it'd be both) so it wouldn't interfere with other players on the server.

mtijanic: That would be a lot more useful. Perhaps something like
```C
// override: 0 - not immune, 1 - immune, -1 use efault 
NWNX_Creature_SetImmunityOverride(object creature, int immunity_type, int override);
```

Morderon:
The immunity bypass isn't include here as WilliamDraco helped me out there, just in case anyone is looking for it.

mtijanic:
This seems to be failing our doxygen build because the nss docs don't match the function. Do you want to handle that or should we when merging?

Morderon:
Added the ability to get the item property source id, as might as well get all ABI breaking changes out of the way.

This update 1) Allows easier comparison between effects and their matching item property. and 2) Allows one to create new item properties for effects that don't have one. Note: 2 could always be fudged but this allows the game to handle the effects removal so that it doesn't have to be handled through additional nwscript.

Also fixed the doxygen error.

PR #1031, opened 2020-09-12
Administration: DeletePlayerCharacter now removes TURD as well

### LINKED ISSUE #1023: NWNX_Admin: DeletePlayerCharacter should DeleteTURD
Could maybe make it an argument on whether to keep the TURD defaulting to FALSE. Anyone think of an edge case where the TURD is still wished to be kept?

plenarius:
Resolves #1023 - an additional argument was added to the `NWNX_Administration_DeletePlayerCharacter` function should the admin wish to retain the TURD for some reason.

### DIFF
--- a/Plugins/Administration/Administration.cpp
+++ b/Plugins/Administration/Administration.cpp
@@ -4,6 +4,8 @@
 #include "API/CNetLayer.hpp"
 #include "API/CNetLayerPlayerInfo.hpp"
 #include "API/CNWSPlayer.hpp"
+#include "API/CNWSCreature.hpp"
+#include "API/CNWSCreatureStats.hpp"
 #include "API/CServerExoApp.hpp"
 #include "API/CServerExoAppInternal.hpp"
 #include "API/CExoBase.hpp"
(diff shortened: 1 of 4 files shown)


Plugins/Administration/Administration.cpp
```diff
@@ -122,6 +154,7 @@ Events::ArgumentStack Administration::DeletePlayerCharacter(Events::ArgumentStac
 {
     const auto objectId = Events::ExtractArgument<ObjectID>(args);
     const auto bPreserveBackup = static_cast<bool>(Events::ExtractArgument<int32_t>(args));
+    const auto retainTurd = Events::ExtractArgument<int32_t>(args);
```
Daztek, line 157: Should probably wrap this in a try/catch for a little while:

```c
int32_t retainTurd;
// TODO: Remove this try/catch at some later point
try
{
    retainTurd  = Services::Events::ExtractArgument<int32_t>(args);
}
catch (const std::runtime_error& e)
{
    retainTurd  = false;
}
```

plenarius: Decided I was being too fancy allowing for TURD preservation and just nixed that logic.

PR #1033, opened 2020-09-12
Send dm all creator lists

Morderon:
Resends dm all palettes.

Use case: Updating palettes through the developer folder and wanting to send them to all DM's without a relog.


Plugins/Player/Player.cpp
```diff
@@ -1652,4 +1653,26 @@ ArgumentStack Player::SetSpawnLocation(ArgumentStack&& args)
     return Services::Events::Arguments();
 }
 
+ArgumentStack Player::SendDMAllCreatorLists(ArgumentStack&& args)
+{
+    if(auto *pPlayer = player(args))
+    {
+        auto *pCreature = Globals::AppManager()->m_pServerExoApp->GetCreatureByGameObjectID(pPlayer->m_oidNWSObject);
+
+        if(pCreature->m_pStats->GetIsDM())
```
Daztek, line 1662: `if (pCreature && pCreature->m_pStats->GetIsDM())` here, please!

PR #1045, opened 2020-09-21
Tweak: Change HideDMsOnCharList to HidePlayersOnCharList 

plenarius:
Allow being able to hide both DMs and PCs or either

### DIFF
--- a/Plugins/Tweaks/Tweaks.cpp
+++ b/Plugins/Tweaks/Tweaks.cpp
@@ -7,7 +7,7 @@
 #include "Tweaks/ParryAllAttacks.hpp"
 #include "Tweaks/SneakAttackCritImmunity.hpp"
 #include "Tweaks/PreserveDepletedItems.hpp"
-#include "Tweaks/HideDMsOnCharList.hpp"
+#include "Tweaks/HidePlayersOnCharList.hpp"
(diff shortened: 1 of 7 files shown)


Plugins/Tweaks/Tweaks.cpp
```diff
@@ -89,8 +89,19 @@ Tweaks::Tweaks(Services::ProxyServiceList* services)
 
     if (GetServices()->m_config->Get<bool>("HIDE_DMS_ON_CHAR_LIST", false))
     {
-        LOG_INFO("DMs will not be visible on character list");
-        m_HideDMsOnCharList = std::make_unique<HideDMsOnCharList>(GetServices()->m_hooks.get());
+        LOG_INFO("NWNX_TWEAKS_HIDE_DMS_ON_CHAR_LIST has been deprecated, please use NWNX_TWEAKS_HIDE_PLAYERS_ON_CHAR_LIST = 1");
+        m_HidePlayersOnCharList = std::make_unique<HidePlayersOnCharList>(GetServices()->m_hooks.get(), 1);
+    }
+
+    if (auto mode = GetServices()->m_config->Get<int>("HIDE_PLAYERS_ON_CHAR_LIST", 0))
```
Daztek, line 96: I dunno what happens if someone sets both env vars, but maybe change this to an `else if`

PR #1078, opened 2020-10-19
Events: Spell Interruption Events

### LINKED ISSUE #1077: NWNX_Events: on cast interrupt (request)
I've implemented a mana-based spellcasting system where spells are restored after each casting as long as there is enough mana remaining. My one problem is that I have no way of detecting when a spell gets interrupted.  Any chance something like that could be added to the events plugin?

plenarius:
Resolves #1077 

Rather than try to dive deep into the mess of `CNWSEffectListHandler::OnApplyDamage` we hook the specific visual effect event for spell interruption. This is not skippable of course.

Note: It's not possible to send the `LAST_DAMAGER_ID` with the event because that value is not yet set before the spell is interrupted so the only data sent is the `SPELL_ID`


burmashave:
Thank you!

Issue #791, opened 2020-02-25
nwnx_chat - Enable NWNX_Chat_SendMessage() to do per target message sending on channel NWNX_CHAT_CHANNEL_PLAYER_DM


Chimerik:
NWNX_Chat_SendMessage() is able to send message exclusively to a target on most channels, but not on NWNX_CHAT_CHANNEL_PLAYER_DM.

Would it be possible to enable this method to send message per target on this channel?

Thanks!

WilliamDraco:
I anecdotally understand this also applies to NWNX_CHAT_CHANNEL_PLAYER_PARTY in the same way. Attempting to send a different message over party to two players results in both players receiving both messages.

Issue #746, opened 2020-01-20
Use the Events plugin infrastructure for Chat and Damage


mtijanic:
Chat and Damage (and maybe other?) plugins implement their own ad-hoc version to register and signal events, and to pass data. They should either use the message bus to signal them via nwnx_events, or just move the functionality out of these plugins into the events directly.

Should leave a deprecation compat shim for a while

mtijanic:
Also Weapon's devastating critical. I think Weapon and Damage can just be collapsed entirely into other plugins.

mtijanic:
...and devastating critical from Weapon...

PR #1090, opened 2020-10-27
Events: Add Chat

plenarius:
As requested in #746 

I'm not really sure how to best create a deprecation shim so for now it's just a bunch of notices

Note: I also added/(fixed?) targeted messages now being able to be sent on party channel.


plenarius:
Now closes #791 too

plenarius:
Tabling this as I reflect on some hurdles with module chat script and core nwn chat functionality. This may just have to be a breaking change but I'd like to come up with a document for migration if that's the case.

PR #1108, opened 2020-11-02
Object: Don't SetNumIntegers to less than 8 in AddIconEffect

plenarius:
It looks like setting the number of integers to a value less than 8 can cause some buffer overruns later, something to remember.

Copy of backtrace: https://gist.github.com/plenarius/9d2e41da411efefc239426fe83d01071


Daztek:
Yea, don't think the game ever sets it to <8 anywhere, only to like 20 for EffectDamage and some other stuff.

Issue #817, opened 2020-03-10
Feature: Detect mode events


weavejester:
It's sometimes useful to know whether a player has activated detect mode, so I'm going to propose adding four more events to the events plugin:

- `NWNX_ON_ENTER_DETECT_BEFORE`
- `NWNX_ON_ENTER_DETECT_AFTER`
- `NWNX_ON_EXIT_DETECT_BEFORE`
- `NWNX_ON_EXIT_DETECT_AFTER`

These would behave the same way as the current stealth mode events.

Baaleos:
Can I suggest we standardise the naming convention with regard to ENTER/EXIT mode events?

NWNX_ON_ENTER_DETECT_BEFORE
NWNX_ON_ENTER_DETECT_AFTER
NWNX_ON_EXIT_DETECT_BEFORE
NWNX_ON_EXIT_DETECT_AFTER

to become:

NWNX_ON_DETECT_ENTER_BEFORE
NWNX_ON_DETECT_ENTER_AFTER
NWNX_ON_DETECT_EXIT_BEFORE
NWNX_ON_DETECT_EXIT_AFTER

If we were to enumerate / index these event names - it keeps events related to the same functionality side by side - where as putting the EXIT/Enter earlier in the string can end up with events being spread across a wider list.

This could also be applied to the existing Stealth event on enter/exit.

Arvirago:
I would suggest yet another event to detect when a PC enters/exits combat:
NWN_ON_COMBAT_ENTER_BEFORE
NWN_ON_COMBAT_ENTER_AFTER
NWN_ON_COMBAT_EXIT_BEFORE
NWN_ON_COMBAT_EXIT_AFTER

PR #1128, opened 2020-11-07
Events: Adds Combat Enter/Exit events

plenarius:
As requested in #817 - only works for PCs. Hooks into the message from the server to the client to start playing battle music.


Arvirago:
Thanks!

PR #1131, opened 2020-11-08
Player: Added an optional duration parameter to ApplyLoopingVisualEffectToObject()

### LINKED ISSUE #847: Adding duration variable to NWNX_Player_ApplyLoopingVisualEffectToObject
As the title implies, I think, if feasible, it would be a good idea to add a duration to this NwNx function since otherwise it needs to be removed manually.

void NWNX_Player_ApplyLoopingVisualEffectToObject(object player, object target, int visualeffect)

Thanks!

plenarius:
Resolves #847 - This just checks for an expiry when doing the `ComputeGameObjectUpdateForObject` which is about 5-6 times a second when the object is in view.

About halfway through I realized you could just make an nss wrapper like below to do the same thing but I guess this is more convenient.
```c
void ApplyLoopingVisualEffectToObjectDuration(object oPC, object oObject, int vfx, float fDuration)
{
    NWNX_Player_ApplyLoopingVisualEffectToObject(oPC, oObject, vfx);
    DelayCommand(fDuration, NWNX_Player_ApplyLoopingVisualEffectToObject(oPC, oObject, vfx));
}
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -45,6 +45,7 @@ The following plugins were added:
 - Events: `NWNX_ON_DM_SPAWN_OBJECT_*` now provides the resref as event data.
 - Events: `NWNX_ON_STORE_REQUEST_*_AFTER` now provides the result as event data.
 - Events: ResourceEvents now support valid custom resource directory aliases.
+- Player: added an optional duration parameter to ApplyLoopingVisualEffectToObject()
 - Util: added an optional appearance type parameter to CreateDoor()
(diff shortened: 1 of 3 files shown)


Daztek:
I feel the user just using a DelayCommand wrapper like you said would be fine.

mtijanic:
Yes please, avoid adding complexity in the lower levels when it can be done simpler in nwscript. Also will be more optimal with a delaycommand.

PR #1134, opened 2020-11-10
dotnet 5.0 dockerfile changes

urothis:
This will update both dockerfiles to dotnet 5.0.
Only affect should be supporting more projects.


plenarius:
If you're just getting the deb directly through wget then there's no need for `gpg`, and `apt-transport-https` any more. With that said if you remove those you'll have to restore `libssl1.1` to the run-deps. I think it's probably better that we revert [this commit ](https://discordapp.com/channels/@me/534906990305083407/775871049807560724 ) any way, it's not saving much time when apt tells us it's already installed and we don't want to later break things if we change our install dependencies. 

Final thought, install-deps should only be `wget` and `ca-certificates`, add `libssl1.1` and `libsqlite3-0` to run-deps and purge `wget` and `ca-certificates` when completed.

PR #1144, opened 2020-11-15
very complex and important commit)

stenly87:
fix for calculating the skill bonus from an ability with a negative modifier


plenarius:
Nice catch, thank you!

Can you update the CHANGELOG.md under the ### Fixed section, add:
`SkillRanks: fixed skill bonus for abilities not properly calculating when ability scores were negative`
(after the "Rename: properly updates....")

stenly87:
I tried to do something) hope its correct

PR #1146, opened 2020-11-16
Core: Log reports NWNX version/sha on startup and on crashes

plenarius:


mtijanic:
This is the sha1 of nwnx commit, but the constant name and the output makes it seem like it's the nwn sha1 commit (which we also have and print earlier). Would be nice to rename to something more obvious

PR #1148, opened 2020-11-16
CI: Update Dockerfiles to use new base image

plenarius:
Also includes:
- label our builds build8193.16-HEAD for example
- building against libmariadb and not libmariadbclient and caching all deps properly
- using ADD instead of COPY so docker only needs one new layer as it auto-extracts the tar
- OCI labels

Note: This Pull Request build will break due to needing the changes from PR #1147


plenarius:
Oh right, the PR didn't break because we don't do anything with docker until we push!

PR #1195, opened 2021-01-02
Extend Crit range/modifiers with BaseItem parameters

WilliamDraco:
Updates and extends the functions Get/Set pairs for CriticalMultiplier and CriticalRange, both modifiers and overrides, as follows:

* Now allows optional declaration of a BASE_TYPE_*. If specified, modifier/override applies only to attacks made with the specified weapon. Can apply to Unarmed (`nBaseItem == -1`). Weapon declaration takes priority in Overrides. That is, the order goes:
  1. Baseitem and specified hand
  2. Baseitem and non-specified hand
  3. No Baseitem and specified hand
  4. No Baseitem and non-specified hand

* Order of parameters maintained for compatibility, although I would've liked it earlier.

* Modifiers now apply in addition to Overrides, and multiple modifiers take effect if relevant. I.E Overriding Main-hand Longsword with multiplier of 3, and Modifier of any Off-Hand of -1, and a Modifier of any Longsword of +2 will result in (3-1+2 =) 4, rather than previously where it would simply stop at the Override and return 3.

### DIFF
--- a/Plugins/Creature/Creature.cpp
+++ b/Plugins/Creature/Creature.cpp
@@ -5,6 +5,8 @@
 #include "API/CServerInfo.hpp"
 #include "API/CNWSCreature.hpp"
 #include "API/CNWSCreatureStats.hpp"
+#include "API/CNWSItem.hpp"
+#include "API/CNWSInventory.hpp"
 #include "API/CNWLevelStats.hpp"
 #include "API/CNWSFaction.hpp"
 #include "API/CNWSStats_Spell.hpp"
@@ -373,7 +375,6 @@ ArgumentStack Creature::GetMeetsFeatRequirements(ArgumentStack&& args)
     }
     return Services::Events::Arguments(retVal);
 }
-
 
 ArgumentStack Creature::GetSpecialAbility(ArgumentStack&& args)
 {
@@ -1978,7 +1979,6 @@ ArgumentStack Creature::SetFaction(ArgumentStack&& args)
     return Services::Events::Arguments();
 }
 
-
 ArgumentStack Creature::GetFaction(ArgumentStack&& args)
 {
     int32_t retVal = -1;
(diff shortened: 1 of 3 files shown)


Plugins/Creature/NWScript/nwnx_creature.nss
```diff
@@ -685,61 +685,69 @@ int NWNX_Creature_GetCasterLevelOverride(object oCreature, int nClass);
...
 
-/// @brief Sets the critical hit multiplier modifier for the creature
+/// @brief Sets the critical hit multiplier modifier for the Creature
 /// @param oCreature The target creature
 /// @param nModifier The modifier to apply
 /// @param nHand 0 for all attacks, 1 for Mainhand, 2 for Offhand
 /// @param bPersist Whether the modifier should persist to .bic file if applicable
-/// @note Persistence is activated each server reset by first use of either 'SetCriticalMultiplier*' functions. Recommended to trigger on a dummy target OnModuleLoad to enable persistence.
-void NWNX_Creature_SetCriticalMultiplierModifier(object oCreature, int nModifier, int nHand = 0, int bPersist = FALSE);
+/// @param nBaseItem Applies the.modifier only when the attack used this baseitem. -1 for 'Unarmed', '-2' for 'all'
```
mtijanic, line 693: The game and nwnx generally use BASE_ITEM_GLOVES for unarmed.

Plugins/Creature/Creature.cpp
```diff
@@ -2397,53 +2438,66 @@ void Creature::InitCriticalRangeHook()
                 int32_t retVal;
                 if (!bOffHand) //mainhand
                 {
-                    if (auto critRngOvr = g_plugin->GetServices()->m_perObjectStorage->Get<int32_t>(pThis->m_pBaseCreature->m_idSelf, "CRITICAL_RANGE_OVERRIDE!1"))
-                    {
+                    auto pItem = pThis->m_pBaseCreature->m_pInventory->GetItemInSlot(Constants::EquipmentSlot::RightHand);
+                    std::string nBaseItemID;
```
mtijanic, line 2442: Ideally don't use hungarian notation prefixes at all, but please don't use `n` for strings.

mtijanic (commented):
Sorry, just did a quick skim. It's a big diff so it'll take a while to review properly.

Plugins/Creature/NWScript/nwnx_creature.nss
WilliamDraco: Swapped -1 out for Gloves constants, and changed the "all" from -2 to -1 accordingly.

Plugins/Creature/Creature.cpp
WilliamDraco: Removed all Hungarian (Including some that was from the previous incarnation)

PR #1197, opened 2021-01-03
Add docker env variable support for NWSync hosted modules.

Jorteck:


Daztek:
You'll also need to update the `run-server.patch` file. (`diff run-server.sh.orig run-server.sh.new > run-server.patch`)

And it appears the [upstream](https://github.com/niv/dockerhub-beamdog-nwserver/blob/master/run-server.sh) run-server.sh has changed too, so run-server.sh.new needs those changes too.

Jorteck:
Updated :)

Daztek:
This looks ok, I don't use docker so I can't test. #yolo

PR #1200, opened 2021-01-08
SWIG: Add DotNET Native Interop support

Jorteck:
This PR implements the generation of bindings that are needed to support Managed <-> Native interop in DotNET. The Managed implementation can be found here: https://github.com/nwn-dotnet/NWN.Native/tree/main/src/main/API

I'd like some feedback on this and possibly some help migrating the "generate.sh" script to a SWIG CMake task: https://cmake.org/cmake/help/latest/module/UseSWIG.html


NWNXLib/API/API/CNWSDialog.hpp
```diff
@@ -61,15 +62,15 @@ struct CNWSDialog
...
+    BOOL GetStartEntryOneLiner(CNWSObject * pNWSObjectOwner, CExoLocString & sOneLiner, CResRef & sSound, CResRef & sScript, CExoArrayList<ScriptParam>& scriptParams);
     float SetDialogDelay(CNWSObject * pNWSObjectOwner, CExoLocString sLocText, uint32_t nMinSeconds, BOOL bEnableNPCTiming);
     void AddJournalEntry(const CExoString & szPlotID, uint32_t nState, uint32_t nPlayerId);
     BOOL SendDialogEntry(CNWSObject * pNWSObjectOwner, uint32_t nPlayerIdGUIOnly, uint32_t iEntry, BOOL bPlayHelloSound);
     BOOL SendDialogReplies(CNWSObject * pNWSObjectOwner, uint32_t nPlayerIdGUIOnly);
     BOOL HandleReply(uint32_t nPlayerID, CNWSObject * pNWSObjectOwner, uint32_t nReplyIndex, BOOL bEscapeDialog, uint32_t currentEntryIndex);
     BOOL RemovePlayer(OBJECT_ID oidRemove);
     BOOL IsPlayerInDialog(OBJECT_ID oidPlayer);
-    void LoadScriptParams(CResGFF * pRes, ScriptParamsList & paramsList, CResStruct * pStructure, char * szFieldID);
+    void LoadScriptParams(CResGFF * pRes, CExoArrayList<ScriptParam>& paramsList, CResStruct * pStructure, char * szFieldID);
```
mtijanic, line 73: +1, the old way didn't match the mangled name.

NWNXLib/API/API/CNWSScriptVar.hpp
```diff
@@ -23,7 +23,6 @@ struct CNWSScriptVar
     CScriptLocation m_location;
 
     CNWSScriptVar();
-    CNWSScriptVar(const CNWSScriptVar &);
```
mtijanic, line 26: Ugh, this whole class is missing from FunctionsLinux.hpp. The API _is_ correct. I'll add the offsets.

NWNXLib/API/API/CTlkTable.hpp
```diff
@@ -47,7 +47,7 @@ struct CTlkTable
     virtual void GetTokenValue(const CTlkTableToken & cTlkTableToken, CExoString & sToken);
     BOOL FetchInternal(uint32_t strId, STR_RES & strRes, BOOL bParse);
     void CloseFile(CTlkFile * pTlkFile);
-    BOOL OpenFile(const CExoString & sFileName, CTlkFile * pTlkFile);
+    BOOL OpenFile(const CExoString & sFileName, CTlkFile ** pTlkFile);
```
mtijanic, line 50: Technically, it's `CTlkFile *pTlkFile[2]`, and the CloseFile() should be updated as well.

Plugins/DotNET/CMakeLists.txt
```diff
@@ -1,3 +1,14 @@
...
-    "DotNET.cpp"
-    "DotNETExports.cpp")
+add_custom_target(
+    DotNETGenerate
+    COMMAND ${CMAKE_CURRENT_SOURCE_DIR}/generate.sh
+    WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}
+)
+
+add_library(DotNET SHARED "DotNET.cpp" "DotNETExports.cpp" "DotNETInterop.cpp")
+set_property(SOURCE ${CMAKE_CURRENT_SOURCE_DIR}/DotNETInterop.cpp PROPERTY GENERATED 1)
```
mtijanic, line 8: Can you post what the generated .cpp file looks like?

mtijanic:
I merged the API fixup separately, so we can keep this more ontopic.

Plugins/DotNET/CMakeLists.txt
```diff
@@ -1,3 +1,3 @@
 add_plugin(DotNET
     "DotNET.cpp"
-    "DotNETExports.cpp")
+    "DotNETExports.cpp")
```
mtijanic, line 3: Should probably revert all changes to DotNET/ - ere, and DotNET/DotNET.i

Plugins/SWIG/DotNET.i
```diff
@@ -0,0 +1,156 @@
...
+%template(CResHelperIFO) CResHelper<CResIFO,2014>;
+%template(CResHelperSET) CResHelper<CResSET,2013>;
+%template(CResHelperWOK) CResHelper<CResWOK,2016>;
+%template(CResHelperNSS) CResHelper<CResNSS,2009>;
+%template(CResHelperTLK) CResHelper<CResTLK,2018>;
+%template(CResHelperNDB) CResHelper<CResNDB,2064>;
+%template(CResHelperNCS) CResHelper<CResNCS,2010>;
+
+// Ignore APIs with no symbols.
+%ignore CNWSObject::BindEventToHandler;
```
mtijanic, line 71: All of these should no longer be necessary.

Plugins/SWIG/NWNXLib.i
```diff
@@ -0,0 +1,671 @@
+%{
+#include "API/ALL_CLASSES.hpp"
+#include "API/AurList.hpp"
```
mtijanic, line 3: You'll need to regenerate this list, I deleted quite a few of the headers..

mtijanic (commented):
I'm not sure if we should generate the .cpp file every build, or simply keep it checked in along with generate.sh and run it manually. API updates are really rare, so it's no issue to do it manually, and saves everyone a dependency during build time.. and makes cmake much simpler. What do you think?

Jorteck:
> I'm not sure if we should generate the .cpp file every build, or simply keep it checked in along with generate.sh and run it manually. API updates are really rare, so it's no issue to do it manually, and saves everyone a dependency during build time.. and makes cmake much simpler. What do you think?

I wouldn't mind keeping it for validation of the API. Is the extra build time significant?

.github/workflows/build.yml
```diff
@@ -24,7 +24,7 @@ env:
 
 jobs:
   build:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-20.04
```
mtijanic, line 27: Can you explain why this is needed? It sounds like a good idea to use a fixed version regardless, but it also sounds like it should be a separate PR altogether.

Plugins/DotNET/CMakeLists.txt
mtijanic: I'm still seeing DotNET/DotNET.i file and a change (removed newline) in DotNET/CMakeLists.txt.

Plugins/SWIG/generate.sh
```diff
@@ -0,0 +1,22 @@
...
+echo "%}" >> $NWNXLIB_SWG
+
+for include in "${INCLUDES[@]}"
+do
+    echo "%include \"$include\"" >> $NWNXLIB_SWG
+done
+
+find ../../NWNXLib/API -type f -name '*.hpp' "${FIND_ARGS[@]}" -printf '%%include "%P"\n' >> $NWNXLIB_SWG
+
+swig ${@:1}
```
mtijanic, line 22: Please configure your editor to produce a newline at the end of files. There are plenty of tools that get confused by this, and also most languages' standards explicitly require it (even if the actual implementation can deal with it missing)

Plugins/SWIG/CMakeLists.txt
```diff
@@ -0,0 +1,18 @@
...
+    set_property(SOURCE ${CMAKE_CURRENT_SOURCE_DIR}/${interopfile} PROPERTY GENERATED 1)
+    add_sanitizers(${target})
+    target_link_libraries(${target} Core "-Wl,--no-undefined")
+    set_target_properties(${target} PROPERTIES PREFIX "${PLUGIN_PREFIX}")
+    target_include_directories(${target} PRIVATE "${CMAKE_CURRENT_SOURCE_DIR}")
+    target_compile_definitions(${target} PRIVATE "-DPLUGIN_NAME=\"${PLUGIN_PREFIX}${target}\"")
+    add_dependencies(${target} Generate_${target})
+endfunction()
+
+add_swig_plugin(SWIG_DotNET "DotNETInterop.cpp" -csharp -dllimport "${PLUGIN_PREFIX}SWIG_DotNET" -namespace "NWN.Native.API" "DotNET.i")
```
mtijanic, line 18: If we are going to keep generating it every build, rather than do manually, we need two more things:
1. Skip the plugin altogether if people don't have the swig dependency installed. Adding a new plugin should never break people's builds.
2. Set up make dependencies on the API, so that it is only rebuilds the generated files if the source API files changed. If I'm reading this right, it will generate DotNETInterop.cpp every time I edit an unrelated file in a different plugin and try to build it.

mtijanic (commented):
I think this also needs the actual barebones plugin body that exposes `PluginLoad`, otherwise the core will throw an error when loading it. See Scripts/bootstrap_plugin.sh for the minimal(ish) body needed.

.github/workflows/build.yml
Jorteck: ubuntu-latest (18.04) only has SWIG 3.0, and has trouble parsing static_asserts: https://github.com/nwnxee/unified/runs/1677788148#step:10:83.

20.04 fixes this as it has access to SWIG 4.

PR #1210, opened 2021-01-12
Docker: Pin "ubuntu-latest" to 18.04

Jorteck:
~Required for PR: https://github.com/nwnxee/unified/pull/1200~
Required to stop everything breaking.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -24,7 +24,7 @@ env:
 
 jobs:
   build:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-18.04
     steps:
     - uses: actions/checkout@v2
 
@@ -180,7 +180,7 @@ jobs:
            ./NWScript.zip
 
   docker:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-18.04
     if: github.event_name == 'push'
     needs: build
     steps:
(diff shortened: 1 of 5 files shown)


Daztek:
What's ubuntu-latest at? 18.04? If so this would up the requirement for our downloadable binaries to 20.04 too I think

Jorteck:
It's currently being rolled out to 20.04: https://github.com/actions/virtual-environments/issues/1816

Daztek:
Do you know if stuff compiled with 20.04 still works in beamdog/nwserver's debian buster?

Jorteck:
Doesn't look like it - `libstdc++6` is one version off (Ubuntu: 3.4.26/6.0.26, Debian: 3.4.25/6.0.25):

> ./nwserver-linux: /usr/lib/x86_64-linux-gnu/libstdc++.so.6: version 'GLIBCXX_3.4.26' not found (required by /home/debian/nwn-server/bin/linux-x86/NWNX_Core.so)

https://packages.debian.org/buster/amd64/libstdc++6/filelist

Daztek:
Don't think we can switch to 20.04 in that case and should probably pin it to 18.04 before everything blows up.

.github/workflows/docs.yml
```diff
@@ -18,7 +18,7 @@ on:
 
 jobs:
   build:
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-18.04
```
Daztek, line 21: I think docs has to be on 20.04 for doxygen

plenarius: Yes, docs needs to be on 20.04 please, we need doxygen >= 1.8.15

Daztek:
If this breaks everything, it wasn't me

PR #1221, opened 2021-01-17
Time sqlite

### LINKED ISSUE #1218: Deprecate Time Plugin in favour of sqlite functions
Can be easily managed with sqlite now. See https://sqlite.org/lang_datefunc.html

* Send deprecation warning on Time plugin loaded
* Change nwnx_time to include helper stub inc_sqlite_time and call the functions that exist in that new library
* Default return value of GetSystemDate to `%m/%d/%Y` but provide an optional format string argument
* Default return value of GetSystemTime to `%H:%M:%S` but provide an optional format string argument
* include new GetSystemDateTime for both that defaults to `'%Y-%m-%d %H:%M:%S'` but provide an optional format string argument
* GetHighResTimeStamp will now only report in milliseconds instead of microseconds as that's all sqlite provides, should be okay

hendrikgit:
Giving this a try so there is at least some code to talk about. Thanks for the comments to get started @plenarius. I have a few questions.

I did not add an optional format string argument to GetSystemDate and GetSystemDate. Nor did I add those functions specifically in the nwscript include. I figure when you can pass a format string you could make either of those functions return any format and that seemed weird.
So instead in the new include I added SQLite_GetFormattedSystemTime(string format) where you pass in a format string. That is also the function that is now used in nwnx_time.nss in NWNX_Time_GetSystemDate and NWNX_Time_GetSystemTime. Let me know what you think, please.

Should I add GetSystemDate, GetSystemTime and a new GetSystemDateTime to the include for convenience or are we going to stop providing this include file at some point and people are supposed to write their own anyway?

I'm not familiar with the DoxyGen comments. In the include should this `/// @file inc_sqlite_time.nss` match the file name or point to nwnx_time.nss (I saw that in inc_array)? Generally I'm not sure I used all the doc comments correctly.

Is it too spammy to write to the log on every call of a NWNX_Time_* function?

Is my NWNX_Time_GetHighResTimeStamp implementation correct, the bit where I just multiply the milliseconds by 1000?

Once this is on the right track I'll try to move NWNX_Time_GetHighResTimeStamp to NWNX_Util_GetHighResTimeStamp with the original implementation so there is still microsecond resolution available. I don't know any C++ so I did not start on that yet.

(Resolves #1218)

### DIFF
--- a/Plugins/Time/NWScript/nwnx_time.nss
+++ b/Plugins/Time/NWScript/nwnx_time.nss
@@ -3,17 +3,22 @@
 /// @{
 /// @file nwnx_time.nss
 #include "nwnx"
+#include "nwnx_util"
+#include "inc_sqlite_time"
 
 const string NWNX_Time = "NWNX_Time"; ///< @private
 
 /// @brief Returns the current date.
+/// @deprecated Use SQLite functions (see inc_sqlite_time). This will be removed in future NWNX releases.
 /// @return The date in the format (mm/dd/yyyy).
 string NWNX_Time_GetSystemDate();
 
 /// @brief Returns current time.
+/// @deprecated Use SQLite functions (see inc_sqlite_time). This will be removed in future NWNX releases.
 /// @return The current time in the format (24:mm:ss).
 string NWNX_Time_GetSystemTime();
 
+/// @deprecated Use SQLite functions (see inc_sqlite_time). This will be removed in future NWNX releases.
 /// @return Returns the number of seconds since midnight on January 1, 1970.
 int NWNX_Time_GetTimeStamp();
 
@@ -24,6 +29,7 @@ struct NWNX_Time_HighResTimestamp
     int microseconds; ///< Microseconds
 };
 
+/// @deprecated Use NWNX_Util_GetHighResTimeStamp(). This will be removed in future NWNX releases.
 /// @return Returns the number of microseconds since midnight on January 1, 1970.
 struct NWNX_Time_HighResTimestamp NWNX_Time_GetHighResTimeStamp();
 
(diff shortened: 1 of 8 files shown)


plenarius:
Thanks for working on this!

> Giving this a try so there is at least some code to talk about. Thanks for the comments to get started @plenarius. I have a few questions.
> 
> I did not add an optional format string argument to GetSystemDate and GetSystemDate. Nor did I add those functions specifically in the nwscript include. I figure when you can pass a format string you could make either of those functions return any format and that seemed weird.
> So instead in the new include I added SQLite_GetFormattedSystemTime(string format) where you pass in a format string. That is also the function that is now used in nwnx_time.nss in NWNX_Time_GetSystemDate and NWNX_Time_GetSystemTime. Let me know what you think, please.
> 
> Should I add GetSystemDate, GetSystemTime and a new GetSystemDateTime to the include for convenience or are we going to stop providing this include file at some point and people are supposed to write their own anyway?


I think we should provide those first two functions in the new include even though they basically just call `SQLite_GetFormattedSystemTime`  as a convenience so existing users of Time can just do a quick search and replace of `NWNX_Time_` to `SQLite_`  and `#include "nwnx_time"` with `#include "inc_sqlite_time"` and then  drop this new include in their module and be done with any deprecation warnings and not have to go and manually edit each call they used to do. While true we won't support this include in the nwnx namespace forever, making the transition include as drop-in easy as possible I think is nice for the users.

As for the other function `GetSystemDateTime` , no then, that's not necessary to provide then with what you're doing.

> 
> I'm not familiar with the DoxyGen comments. In the include should this `/// @file inc_sqlite_time.nss` match the file name or point to nwnx_time.nss (I saw that in inc_array)? Generally I'm not sure I used all the doc comments correctly.

Let me take a look at this, inc_array may be wrong.

> 
> Is it too spammy to write to the log on every call of a NWNX_Time_* function?

Nope, deprecation warnings are typically spammy.

> Is my NWNX_Time_GetHighResTimeStamp implementation correct, the bit where I just multiply the milliseconds by 1000?
> 
> Once this is on the right track I'll try to move NWNX_Time_GetHighResTimeStamp to NWNX_Util_GetHighResTimeStamp with the original implementation so there is still microsecond resolution available. I don't know any C++ so I did not start on that yet.

I don't think we can do what you've done with `NWNX_Time_GetHighResTimeStamp` during deprecation. It should work as it did before, using NWNX_Time  but now with a deprecation warning to change to NWNX_Util.

plenarius:
Minor nit, can you capitalize `select` and `substr` in your SQL queries.
Also please include an update to the CHANGELOG.md

Otherwise I  think it all looks good :+1:

hendrikgit:
Should be good to go now. Unless my changelog additions are too verbose.

Plugins/Time/NWScript/nwnx_time.nss
```diff
@@ -24,35 +28,33 @@ struct NWNX_Time_HighResTimestamp
...
-    return NWNX_GetReturnValueInt(NWNX_Time, sFunc);
+    WriteTimestampedLogEntry("WARNING:  NWNX_Time is deprecated.  You should migrate to SQLite based functions (see inc_sqlite_time).");
+    return SQLite_GetTimeStamp();
 }
 
 struct NWNX_Time_HighResTimestamp NWNX_Time_GetHighResTimeStamp()
 {
+    WriteTimestampedLogEntry("WARNING:  NWNX_Time is deprecated.  NWNX_Time_GetHighResTimeStamp is moving to NWNX_Utility.");
     struct NWNX_Time_HighResTimestamp t;
     string sFunc = "GetHighResTimeStamp";
```
mtijanic, line 60: Don't we want to move the call to NWNX_Util_GetHighResTimestamp() here?

hendrikgit: I noticed I wrote NWNX_Utility instead of NWNX_Util. I'll correct that.

Changing to calling the function in NWNX_Util there would mean it would stop working for someone who does not have the util plugin enabled I think?

mtijanic: Yeah, but only if they update the nss. If they're already updating the nss, they can just enable Util.

Also, I have a hard time imagining someone using this time function but not having Util enabled.

hendrikgit: I made the change (and rebased on current master and cleaned up commits a bit).

Issue #960, opened 2020-07-10
request: GetObjectAILevel() SetObjectAILevel()


shadguy:
Per subject line; specifically interested in placeables.  Background info from Discord chatter is pasted below.
---

From Discord:

Shadooow: Placeables tend to fire the spell with a delay in large modules. But maybe I can just time the delay and hope it will be consistent.

clippy: Oh, I think that's just their VERY_LOW AI, try SetObjectAILevel() a higher level on it temporarily and then fire the spell

shadguy:  I see a SetAILevel() command with comments about creatures, but not SetObjectAILevel();  it sounds worth trying the former on placeables regardless, from your comment.

clippy: That's the one I meant. I think it works on placeables.  If not, something nwnx can easily add

shadguy: Thanks - I've got some respawning traps that didnt get deployed much because they basically stop performing after respawn - will try this weekend.  Sounds strongly related to Shadooow's comment, thanks for the proposed solution :slight_smile:

Daz: SetAILevel() only works on non player creatures

Daztek:
Resolved by #1224

PR #1224, opened 2021-01-20
{Get/Set}AILevel for objects

hendrikgit:
Functions as requested in #960.

In my tests it "worked fine" insofar as levels between -1 and 4 could be read and set for all types of objects. Items, placeables, creatures, doors and even a PC.

Should this be limited to certain object types, like placeable, anyway?

Is setting of AI level 4 ok? It seemed to work but in the nwnlexicon it says that 4 (AI_LEVEL_VERY_HIGH) can not be set manually on creatures with the base game function and that level is reserved to creatures in combat.

### DIFF
--- a/Plugins/Object/Object.cpp
+++ b/Plugins/Object/Object.cpp
@@ -1,6 +1,7 @@
 #include "Object.hpp"
 
 #include "API/CAppManager.hpp"
+#include "API/CServerAIMaster.hpp"
 #include "API/CServerExoApp.hpp"
 #include "API/CNWSObject.hpp"
 #include "API/CNWSScriptVar.hpp"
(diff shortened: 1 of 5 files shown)


Plugins/Object/Object.cpp
```diff
@@ -1003,4 +1005,29 @@ ArgumentStack Object::GetCurrentAnimation(ArgumentStack&& args)
...
+}
+
+ArgumentStack Object::SetAILevel(ArgumentStack&& args)
+{
+    if (auto *pObject = object(args))
+    {
+        const auto nLevel = Services::Events::ExtractArgument<int32_t>(args);
+        ASSERT_OR_THROW(nLevel >= -1);
+        ASSERT_OR_THROW(nLevel <= 4);
+        pObject->m_nAILevel = nLevel;
```
mtijanic, line 1027: You (also?) want to call the method in the AIMaster, otherwise this won't actually do much:
```c

    CServerAIMaster* ai = Globals::AppManager()->m_pServerExoApp->GetServerAIMaster();
    ai->SetAILevel(pObject, nLevel);
```

hendrikgit: Thank you. Updated. The setting of `pObject->m_nAILevel` directly does not seem necessary when calling SetAILevel.

PR #1225, opened 2021-01-20
SWIG: Parse FunctionsLinux.hpp, Add API extensions, Cleanup Warnings

Jorteck:


mtijanic:
I'll be merging swig changes from you as soon as they pass the build and they are not markef WIP, with just a cursory glance at the diff.
This one is failing to build, but if you're going to be adding to it, please mark as WIP so I don't merge something half-done as soon as the build part is done.

Jorteck:
Should be good to go, was some platform specific issues where find was returning a different include order locally compared to GHA.

These should now be sorted to be more consistent.

PR #1228, opened 2021-01-21
Get/Set perspace and related fields

### LINKED ISSUE #1194: Get/Set perspace and related fields
`CNWSCreature::m_pcPathfindInformation` has:

```C
    float m_fPersonalSpace;
    float m_fCreaturePersonalSpace;
    float m_fHeight;
    float m_fHitDistance;
```

That should be exposed, with getters and setters. Should call `ComputeStepTolerance()` after modifying `m_fPersonalSpace`.

hendrikgit:
Resolves #1194.

I don't know if ComputeStepTolerance() does what it is supposed to do and how to test that.

I think these functions can be used well together with SetObjectVisualTransform for OBJECT_VISUAL_TRANSFORM_SCALE.
Considering that, would it make sense to add getter and setter for `m_fPreferredAttackDistance`?

I scaled a dragon down with visual transform and reduced the perspace fields by the same factor.
It works for collision detection but the walk animation is off. Does this have to do with the fields `WALKDIST` and `RUNDIST` from appearance.2da? Can it be changed via the creature object?
I saw `float ComputeTotalWalkDistance()` but did not get further than that.

Edit: Daz told me walk animation can be changed with SetObjectVisualTransform() for OBJECT_VISUAL_TRANSFORM_ANIMATION_SPEED. So that seems to be sufficient.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@ https://github.com/nwnxee/unified/compare/build8193.20...HEAD
 - Creature: DoPerceptionUpdateOnCreature()
 - Creature: {Get/Set}PersonalSpace()
 - Creature: {Get/Set}CreaturePersonalSpace()
+- Creature: {Get/Set}Height()
 - Encounter: GetGeometry()
 - Encounter: SetGeometry()
 - Util: GetInstructionLimit()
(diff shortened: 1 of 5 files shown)


mtijanic:
Is it still wip? Looks ready to merge to me.

hendrikgit:
> Is it still wip? Looks ready to merge to me.

The functions that are already there are ready to be merged.

However, can you help me with `WALKDIST` and `RUNDIST` @mtijanic? Would changing those help with the walk animations when using visual transform of scale? If so, where/how can those be set? And finally, if I can add set/get for them, should that be in this PR or would you prefer a new one?

mtijanic:
Can you forget about WALKDIST for a bit? There's a fix for the VT scaling coming in the base game. If, after that lands, there's still problems, we can revisit.

PR #1232, opened 2021-01-23
Fix some hooks that might cause trouble due to hook chaining

Daztek:
Fixes ValidateCharacter hooks for ELC/Race and CanLevelUp for NWNX_CreatureLevelUp/NWNX_MaxLevel


Daztek:
Also namespaces the hook order constants

PR #1239, opened 2021-01-26
Encounter: {Get/Set}Reset

hendrikgit:


Plugins/Encounter/Encounter.cpp
```diff
@@ -190,6 +190,29 @@ ArgumentStack Encounter::SetPlayerTriggeredOnly(ArgumentStack&& args)
     return Services::Events::Arguments();
 }
 
+ArgumentStack Encounter::GetReset(ArgumentStack&& args)
```
mtijanic, line 193: These two also need to be added to the `REGISTER()` list at the top of the file

Plugins/Encounter/NWScript/nwnx_encounter.nss
```diff
@@ -53,6 +53,16 @@ int NWNX_Encounter_GetPlayerTriggeredOnly(object encounter);
 /// @param playerTriggeredOnly TRUE/FALSE
 void NWNX_Encounter_SetPlayerTriggeredOnly(object encounter, int playerTriggeredOnly);
 
+/// @brief Get if the encounter respawns or not.
+/// @param encounter The encounter object.
+/// @return TRUE if the encounter does respawn, FALSE otherwise.
+int NWNX_Encounter_GetReset(object encounter);
```
mtijanic, line 59: The function name is not very descriptive.. maybe `{G,S}etRespawning

Plugins/Encounter/Encounter.cpp
hendrikgit: Oops. Thanks. Added now.

Plugins/Encounter/NWScript/nwnx_encounter.nss
hendrikgit: I thought about that too but at the same time that just seems to be it's name.
The member variable is called reset and so is the field in the gff. There are also the two related functions GetResetTime, SetResetTime.
People who are familiar with the gff might look/search for something with reset.

Should I change it anyway?

mtijanic: For what it's worth, the internal name is actually `{G,S}etEncounterCanReset()`, it's just the functions are not exposed in the API, only the field. I think something GetReset() is too confusing. GetCanReset() would be much better.

hendrikgit: Changed it to GetCanReset, SetCanReset.

PR #1248, opened 2021-02-05
Tweaks: use new plugin format

Daztek:
It's beautiful.


Plugins/Tweaks/UnhardcodeShields.cpp
```diff
@@ -10,25 +9,37 @@
...
-    s_ComputeArmourClass = Hooks::HookFunction(Functions::_ZN12CNWSCreature18ComputeArmourClassEP8CNWSItemii,
-                                        (void*)&CNWSCreature__ComputeArmourClass, Hooks::Order::Late);
+    if (!Config::Get<bool>("UNHARDCODE_SHIELDS", false))
+        return;
+
+    LOG_INFO("Using baseitems.2da to define shield AC and create shield-like items");
+
+    s_ComputeArmorClassHook = Hooks::HookFunction(Functions::_ZN8CNWSItem17ComputeArmorClassEv,
+                                                  (void*)&ComputeArmorClassHook, Hooks::Order::Final);
+    s_ComputeArmourClassHook = Hooks::HookFunction(Functions::_ZN12CNWSCreature18ComputeArmourClassEP8CNWSItemii,
```
mtijanic, line 38: Could we use more descriptive names for these two hooks? :D

Daztek: But one has a `u` and one doesn't, isn't that descriptive enough?!

PR #1274, opened 2021-03-02
New function: CreateSoundObject()

BhaalM:


CHANGELOG.md
```diff
@@ -28,6 +28,7 @@ https://github.com/nwnxee/unified/compare/build8193.20...HEAD
 - Admin: {Get/Set}MaxLevel()
 - Area: NWNX_Area_GetAmbientSound{Day/Night}()
 - Area: NWNX_Area_GetAmbientSound{Day/Night}Volume
+- Area: NWNX_Area_CreateSoundObject()
```
mtijanic, line 31: Pre-existing, but we don't list `NWNX_Area_` prefix in the changelog. See `Creature:` below. Doesn't matter either way, I'll clean it up when making the release.

Plugins/Area/NWScript/nwnx_area.nss
```diff
@@ -288,6 +288,12 @@ int NWNX_Area_GetAmbientSoundDayVolume(object oArea);
 /// @return The volume.
 int NWNX_Area_GetAmbientSoundNightVolume(object oArea);
 
+/// @brief Create a sound object.
+/// @param lLoc The location where the sound object will be created
+/// @param sResRef The ResRef of the sound object
+/// @return The sound object.
+object NWNX_Area_CreateSoundObject(location lLoc, string sResRef);
```
mtijanic, line 295: The pattern is that all NWNX_Area functions take `object oArea` as the first argument, so could you please expand this into `object NWNX_Area_CreateSoundObject(object oArea, vector vPosition, string sResRef);`?

PR #1280, opened 2021-03-09
GetSet Effect Immunity Bypass

WilliamDraco:
Implements a Get/Set combo for bypassing/overriding effect immunity (I.E.  any of the IMMUNITY_TYPE_* group)

This is set on the target oCreature, and can be either:
A) For all effects that oCreature attempts to place on others (I.E. outgoing effects), or
B) For all effects that target oCreature (I.E. incoming effects)

If an incoming and outgoing bypass applies in opposite directions, the default effect happens (as though no bypass was used)

Additionally, the bypass can be given an nChance (%) of applying, where:
1) Positive chances 1:100 result in the effect bypassing immunity and forcing NOT IMMUNE
2) Negative chances -1:-100 result in the effect bypassing immunity (or lack of) and forcing IMMUNE)

This implements one part of #1028. I've reached out to Mord re: this PR but it's late my time. If there's an issue, happy to work that out.

### DIFF
--- a/Plugins/Creature/Creature.cpp
+++ b/Plugins/Creature/Creature.cpp
@@ -2882,6 +2882,77 @@ NWNX_EXPORT ArgumentStack GetShieldCheckPenalty(ArgumentStack&& args)
     {
         return pCreature->m_pStats->m_nShieldCheckPenalty;
     }
+    return 0;
+}
 
+NWNX_EXPORT ArgumentStack SetBypassEffectImmunity(ArgumentStack&& args)
+{
+    static Hooks::Hook pGetEffectImmunity_hook =
+        Hooks::HookFunction(Functions::_ZN17CNWSCreatureStats17GetEffectImmunityEhP12CNWSCreaturei,
+        (void*)+[](CNWSCreatureStats *pThis, uint8_t nType, CNWSCreature * pVersus, int32_t bConsiderFeats) -> int32_t
+        {
+            int32_t BypassCounter = 0;
+
+            auto ReturnImmBypass = [&](CNWSCreature *pBypassCreature, std::string Bypass)
+            {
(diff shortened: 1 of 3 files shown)


Plugins/Creature/Creature.cpp
```diff
@@ -2865,3 +2866,84 @@ NWNX_EXPORT ArgumentStack SetPreferredAttackDistance(ArgumentStack&& args)
...
+                }
+            }
+
+            if (BypassCounter > 0)
+                return true;
+            else if (BypassCounter < 0)
+                return false;
+            else
+                return pGetEffectImmunity_hook->CallOriginal<int32_t>(pThis, nType, pVersus, bConsiderFeats);
+        }, Hooks::Order::Late);
```
Daztek, line 2904: You can just put this bit inside `NWNX_EXPORT ArgumentStack SetBypassEffectImmunity(ArgumentStack&& args)` at the top, no need for the bool/function.

Plugins/Creature/Creature.cpp
```diff
@@ -2865,3 +2866,84 @@ NWNX_EXPORT ArgumentStack SetPreferredAttackDistance(ArgumentStack&& args)
 
     return {};
 }
+
+static void InitEffectImmunityBypassHook()
+{
+    static Hooks::Hook pGetEffectImmunity_hook =
+        Hooks::HookFunction(Functions::_ZN17CNWSCreatureStats17GetEffectImmunityEhP12CNWSCreaturei,
+        (void*)+[](CNWSCreatureStats *pThis, uint8_t nType, CNWSCreature * pVersus, int32_t bConsiderFeats = true) -> int32_t
```
Daztek, line 2874: Can leave out the `= true` here, don't think it will do anything.

WilliamDraco:
On request of Mord, added type 255 as an "all Immunity" type. Also put the check into Lambda, hopefully have done that right.

PR #1285, opened 2021-03-14
Player: add SetTlkOverride()

Daztek:
```c
/// @brief Override nStrRef from the TlkTable with sOverride for oPlayer only.
/// @param oPlayer The player.
/// @param nStrRef The StrRef.
/// @param sOverride The new value for nStrRef or "" to remove the override.
/// @param bRestoreGlobal If TRUE, when removing a personal override it will attempt to restore the global override if it exists.
/// @note Overrides will not persist through relogging.
void NWNX_Player_SetTlkOverride(object oPlayer, int nStrRef, string sOverride, int bRestoreGlobal = TRUE);
```

### DIFF
--- a/Plugins/Player/Player.cpp
+++ b/Plugins/Player/Player.cpp
@@ -35,6 +35,7 @@
 #include "API/CExoLocString.hpp"
 #include "API/CNWSPlayerStoreGUI.hpp"
 #include "API/CExoResMan.hpp"
+#include "API/CTlkTable.hpp"
 #include "API/Constants.hpp"
 #include "API/Globals.hpp"
 #include "API/Functions.hpp"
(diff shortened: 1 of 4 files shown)


Plugins/Player/Player.cpp
```diff
@@ -5,14 +5,10 @@
 #include "API/CServerInfo.hpp"
 #include "API/CNWSArea.hpp"
 #include "API/CNWSPlayer.hpp"
-#include "API/CNWSMessage.hpp"
```
mtijanic, line 8: I assume these lot were included transitively? I'd rather we include explicitly what we need than rely on transitive API includes that might change.

Plugins/Player/Player.cpp
```diff
@@ -1620,3 +1608,90 @@ NWNX_EXPORT ArgumentStack CloseStore(ArgumentStack&& args)
 
     return {};
 }
+
+NWNX_EXPORT ArgumentStack SetTlkOverride(ArgumentStack&& args)
+{
+    static Hooks::Hook s_SendServerToPlayerSetTlkOverrideListHook =
```
mtijanic, line 1614: I think it might be better to just do it like SetCustomToken - i.e. override for the current session only, and the upper layer can keep a list of overrides in sqlite and apply them OnClientEnter.

Plugins/Player/Player.cpp
Daztek: Yeah, I'll restore em

Plugins/Player/Player.cpp
Daztek: Works for me

mtijanic (approved):

PR #1287, opened 2021-03-14
Cleanup deprecated stuff

mtijanic:
* Remove Data and Time plugins, leave the script wrappers.
* Remove try/catch blocks to guard against stale nss files
* API BREAKING: Rename nwnx_damage fields

also, migrate sql to new events API. Can't migrate the whole plugin yet because it uses the metrics service.


Daztek (approved):
Looks good to me

PR #1301, opened 2021-03-19
Add isatty

julien-lecomte:
In order to be able to turn off ANSI color codes in log writing

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@ https://github.com/nwnxee/unified/compare/build8193.21...HEAD
 - Effect: RemoveEffectById()
 - Effect: ReplaceEffectByIndex()
 - Player: SetTlkOverride()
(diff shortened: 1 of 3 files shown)


Scripts/buildnwnx.sh
```diff
@@ -43,7 +43,9 @@ pushd ./build-nwnx
 
 cmake -D CMAKE_BUILD_TYPE=$BUILD_TYPE $SANITIZE ..
 
-make ${JOBS} all
+if ! make ${JOBS} all; then 
+  exit 1
+fi
```
mtijanic, line 48: Doesn't this clobber the value and return 1 always? Not what the commit message says?
Also, not related to this PR, would be best as separate.

julien-lecomte: Yes, it does clobber. I can do it in an other MR

PR #1308, opened 2021-03-25
nwscript: don't pass Plugin/Function name for PUSH/POP

Daztek:
#yolo


Daztek:
I think I caught everything that broke. Will merge this tomorrowish.

PR #1314, opened 2021-04-01
Racefixes - Effects vs child race.

Morderon:
Removed the race switch in gettotaleffects, it was unncessary due to all effects that are applied against a parent race apply to each child race.  -> This fixes effects applied vs child race, handled by gettotaleffectbonus, not applying. 
The game handling of stacking is unmodified.

AB added via plugin (NWNX_Race_SetRacialModifier) now increases AB cap by adding AB vs child race AND abvs parent race as that matched how it was stacking above. (Ie a +2 Vs Elf (parent race of drow) and a +3 Vs Drow will increase AB cap vs Drow by 5);

new CGameEfffect to false as no need to generate an id if one is being copied over.


Morderon:
Later added a hook for GetAttackModifierVersus and GetArmorClassVersus, hopefully i got the ordering right. But AttackModifierVersus did work with the weapons plugin

There's no additional conditions for the AC increase as even though the game claims it's a dodge bonus it doesn't function as one and is a plus all the time AC.

PR #1316, opened 2021-04-02
SetCasterLevel* fixes

Morderon:
Fix: Only changes class level when a verified caster level (resistance/effect creation events included) event fires

Fix 2:  SetCasterLevelModification no longer sets caster level to a 255 or higher

Other additions: Changed it so the caster level override fires after modification, so that it's actually an override.

Changed hook order so that the Creatures hook happens after the tweak hook. Again so that a set override can override the tweak. (bit strange that was by changing it to earliest.. but okay)

### DIFF
--- a/Plugins/Creature/Creature.cpp
+++ b/Plugins/Creature/Creature.cpp
@@ -1799,18 +1799,14 @@ static uint8_t CNWSCreatureStats__GetClassLevel(CNWSCreatureStats* pThis, uint8_
 
         if (nClass != Constants::ClassType::Invalid)
         {
+            auto nLevelOverride = pThis->m_pBaseCreature->nwnxGet<int>("CASTERLEVEL_OVERRIDE" + std::to_string(nClass));
+            if (nLevelOverride)
+                return std::clamp(nLevelOverride.value(), 0, 255);
+
             int32_t nModifier = 0;
             auto nLevelModifier = pThis->m_pBaseCreature->nwnxGet<int>("CASTERLEVEL_MODIFIER" + std::to_string(nClass));
             if (nLevelModifier)
                 nModifier = nLevelModifier.value();
-
-            auto nLevelOverride = pThis->m_pBaseCreature->nwnxGet<int>("CASTERLEVEL_OVERRIDE" + std::to_string(nClass));
-            if (nLevelOverride)
-            {
-                auto nLevel = nLevelOverride.value();
-                nModifier = nLevel - retVal;
-            }
-
 
             //Make sure m_nLevel doesn't over/underflow
             nModifier = std::min(nModifier, 255 - pThis->m_ClassInfo[nMultiClass].m_nLevel);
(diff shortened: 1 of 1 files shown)


Plugins/Creature/Creature.cpp
```diff
@@ -1793,23 +1793,24 @@ static uint8_t CNWSCreatureStats__GetClassLevel(CNWSCreatureStats* pThis, uint8_
...
 
-    if (s_bAdjustCasterLevel || nMultiClass < pThis->m_nNumMultiClasses)
+    if (s_bAdjustCasterLevel && nMultiClass < pThis->m_nNumMultiClasses)
     {
         auto nClass = pThis->m_ClassInfo[nMultiClass].m_nClass;
 
         if (nClass != Constants::ClassType::Invalid)
         {
             int32_t nModifier = 0;
+            auto nLevelModifier = pThis->m_pBaseCreature->nwnxGet<int>("CASTERLEVEL_MODIFIER" + std::to_string(nClass));
```
GoLoT, line 1803: This should be after CASTERLEVEL_OVERRIDE so the override can exit early and skip reading the modifier.

Plugins/Creature/Creature.cpp
```diff
@@ -1793,23 +1793,24 @@ static uint8_t CNWSCreatureStats__GetClassLevel(CNWSCreatureStats* pThis, uint8_
...
+            if (nLevelModifier)
+                nModifier = nLevelModifier.value();
+
             auto nLevelOverride = pThis->m_pBaseCreature->nwnxGet<int>("CASTERLEVEL_OVERRIDE" + std::to_string(nClass));
             if (nLevelOverride)
             {
-                auto nLevel = std::max(nLevelOverride.value(), 255);
-                nModifier = nLevel - pThis->m_ClassInfo[nMultiClass].m_nLevel;
+                auto nLevel = nLevelOverride.value();
```
GoLoT, line 1810: These 2 lines can be replaced by `return std::clamp(nLevelOverride.value(), 0, 255);`. No need to calculate the modifier since the hook changes let us set a return value directly instead of modifying the class level in `CNWSCreatureStats`.

GoLoT (commented):
```(bit strange that was by changing it to earliest.. but okay)```

This is because hooks act as a chain now. First (Earliest) hook uses CallOriginal to call the next hook until the last registered hook doesn't CallOriginal (should be a Final hook) or calls the vanilla function. So Earliest will be the first hook to register which will call any other earliest hooks registered after it, then the early hooks and so on.

Morderon:
Also need to figure out the actual maximum is. 255 is the maximum for the datatype but when it's set to 255 nwscript getcasterlevel returns 1.

GoLoT:
I think 40, 60 or w/e value is set by the MaxLevel plugin is a sensible maximum. I don't see a need for anything that goes over the limits set by the engine.

Morderon:
Okay 255 does work as a maximum which I'm happy to let players have even if i'd personally never use it.

I was going over the constraint due to having the tweak enabled. Fixed.

PR #1321, opened 2021-04-07
Fix: changed code example to use correct struct member iFocusFeat

Kaikas:
The example in the Readme referenced a value focusFeat in the struct NWNX_SkillRanks_SkillFeat which should be iFocusFeat. See: https://github.com/nwnxee/unified/blob/19cdd48ff1bb7cbcf1e15dbdd66dcb165c99983a/Plugins/SkillRanks/NWScript/nwnx_skillranks.nss#L46


Daztek:
Thanks!

PR #1323, opened 2021-04-08
Compiler Plugin

Jorteck:
Proof of concept for a plugin that compiles a specified directory of nss files, and spits out the compiled result into another, before immediately closing the server with the appropriate exit code.

I used std::filesystem, thinking that it was in C++17, and in GCC 7, but it looks like it's GCC 8+. The other APIs looked painful to use, and I might need some help swapping this stuff out.

### DIFF
--- a/Plugins/Compiler/Compiler.cpp
+++ b/Plugins/Compiler/Compiler.cpp
@@ -144,7 +144,7 @@ static void CreateResourceDirectory(const CExoString& alias, const std::string& 
 
     Globals::ExoBase()->m_pcExoAliasList->Add(qualifiedAlias, path.c_str());
     Globals::ExoResMan()->CreateDirectory(qualifiedAlias);
-    Globals::ExoResMan()->AddResourceDirectory(qualifiedAlias, priority, true);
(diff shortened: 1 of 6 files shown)


Daztek:
Alas, `filesystem` is still an experimental header in gcc7. `NWNXCore::InitialSetupPlugins()` has some stuff wrt opening a folder and reading stuff using POSIX, if you want to give it a try.

Plugins/Compiler/Compiler.cpp
```diff
@@ -0,0 +1,220 @@
...
+        LOG_INFO("Skipping compilation. NWNX_COMPILER_SRC_DIR or NWNX_COMPILER_OUT_DIR was not specified.");
+        return;
+    }
+
+    const auto& sourcePath = source.value();
+    const auto& outputPath = output.value();
+
+    if (sourcePath == outputPath)
+    {
+        LOG_INFO("Skipping compilation. NWNX_COMPILER_SRC_DIR must not be the same path as NWNX_COMPILER_OUT_DIR.");
```
mtijanic, line 46: This should probably be a supported usecase, but can go in later

PR #1339, opened 2021-05-12
[WIP] Attack Animation Override

Aschent89:
WIP~

This function will override Attack Animations for the player. Care should be taken when using animations that are much longer in duration than a standard attack. This **does** allow Whirlwind animation to be fired at will/OnAttack, however duration is a factor. Looping & Fire/Forget Animations have been proven to work, and this function follows the same requirements as SetRestAnimation.


Aschent89:
I'll be swapping this to experimental. Consider as a WIP for now.

PR #1345, opened 2021-05-22
Core: fix Get/SetLocalCassowary()

Daztek:


Core/NWNXCoreVM.cpp
```diff
@@ -93,6 +93,18 @@ extern NWNXCore* g_core;
...
 int32_t NWNXCore::GetVarHandler(CNWVirtualMachineCommands* thisPtr, int32_t nCommandId, int32_t nParameters)
 {
+    switch (nCommandId)
+    {
+        case VMCommand::GetLocalInt:
+        case VMCommand::GetLocalFloat:
+        case VMCommand::GetLocalString:
+        case VMCommand::GetLocalObject:
+        case VMCommand::GetLocalLocation:
+            break;
```
mtijanic, line 103: Could the location be in the default as well?

Daztek: Sure!

PR #1363, opened 2021-06-06
Events: use the new plugin format

Daztek:
Deletes a ton of unneeded code, now it takes a lot less effort to add some new events


mtijanic (approved):
Quick skim says yay!

PR #1378, opened 2021-06-30
tweaks: add a tweak to change how material names are passed from nwscript

mtijanic:
Similar to the Reset function, if you pass `sMaterial=""`, it will first clear all params with the same name on the object, regardless of the material, and then set it globally for all materials.


Daztek:
Changelog :)

PR #1386, opened 2021-07-16
SetClassByPosition: Also replace old class references in CNWLevelStats

Kalopsia-dev:
While NWNX_Creature_SetClassByPosition correctly replaced the Class ID in the ClassList node, it did not change references to the old class in LvlStatList. This means that when trying to level up after having used the function, the game got confused, would throw a "[Server] Invalid number of multiclasses" error message and revert the levelup.

This PR modifies the function's default behavior to also update these old class references and thus fixes the levelup issue. Should this not be desired, the old functionality remains accessible via the bUpdateLevels argument.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,7 +17,7 @@ https://github.com/nwnxee/unified/compare/build8193.29...HEAD
 - Object: SetLastTriggered()
 
 ### Changed
-- N/A
+- ***API BREAKING*** Creature: SetClassByPosition by default replaces all occurrences of the old class in CNWLevelStats. This can be disabled with the argument 'bUpdateLevels'.
 
 ### Deprecated
 - Weapon: SetWeaponIsMonkWeapon()
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,7 +17,7 @@ https://github.com/nwnxee/unified/compare/build8193.29...HEAD
 - Object: SetLastTriggered()
 
 ### Changed
-- ***API BREAKING*** Creature: SetClassByPosition by default replaces all occurrences of the old class in CNWLevelStats. This can be disabled with the argument 'bUpdateLevels'.
+- ***ABI BREAKING*** Creature: SetClassByPosition by default replaces all occurrences of the old class in CNWLevelStats. This can be disabled with the argument 'bUpdateLevels'.
 
 ### Deprecated
 - Weapon: SetWeaponIsMonkWeapon()
--- a/Plugins/Creature/Creature.cpp
+++ b/Plugins/Creature/Creature.cpp
@@ -1048,18 +1048,19 @@ NWNX_EXPORT ArgumentStack SetClassByPosition(ArgumentStack&& args)
         const auto position = args.extract<int32_t>();
         const auto classID = args.extract<int32_t>();
         const auto bUpdateLevels = args.extract<int32_t>();
-        const auto classIDold = pCreature->m_pStats->GetClass(position);
           ASSERT_OR_THROW(position >= 0);
           ASSERT_OR_THROW(position <= 2);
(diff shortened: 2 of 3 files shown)


CHANGELOG.md
```diff
@@ -17,7 +17,7 @@ https://github.com/nwnxee/unified/compare/build8193.29...HEAD
 - Object: SetLastTriggered()
 
 ### Changed
-- N/A
+- ***API BREAKING*** Creature: SetClassByPosition by default replaces all occurrences of the old class in CNWLevelStats. This can be disabled with the argument 'bUpdateLevels'.
```
mtijanic, line 20: This is ABI breaking (i.e. you need to recompile your scripts), but I wouldn't say API breaking (i.e. you need to edit your scripts)

Plugins/Creature/Creature.cpp
```diff
@@ -1047,12 +1047,26 @@ NWNX_EXPORT ArgumentStack SetClassByPosition(ArgumentStack&& args)
     {
         const auto position = args.extract<int32_t>();
         const auto classID = args.extract<int32_t>();
+        const auto bUpdateLevels = args.extract<int32_t>();
+        const auto classIDold = pCreature->m_pStats->GetClass(position);
```
mtijanic, line 1051: This line should move below the ASSERT_OR_THROWs, so that if `position` is not valid, we get a throw and not a potential crash.

Plugins/Creature/Creature.cpp
```diff
@@ -1047,12 +1047,26 @@ NWNX_EXPORT ArgumentStack SetClassByPosition(ArgumentStack&& args)
...
           ASSERT_OR_THROW(position >= 0);
           ASSERT_OR_THROW(position <= 2);
           ASSERT_OR_THROW(classID >= Constants::ClassType::MIN);
           ASSERT_OR_THROW(classID <= Constants::ClassType::MAX);
 
         pCreature->m_pStats->SetClass(position, classID);
+
+        if (bUpdateLevels)
+        {
+            auto pLevelStats = pCreature->m_pStats->m_lstLevelStats;
```
mtijanic, line 1061: This should be `auto&`, you don't want to make a copy of `m_lstLevelStats`. `auto x = y;` makes a copy of `y` and puts it in `x`. `auto& x = y` makes `x` an alias of `y` (simplified, but close enough).

mtijanic: Also, nit, don't use the `p` prefix unless it's actually a pointer. just `levelStats` or even `levels` would do here as a name.

mtijanic (commented):
Hi, thanks so much for the PR. Overall this looks very good! I have left a few comments inline, and we can merge as soon as those are resolved.

Kalopsia-dev:
Thanks for pointing these things out and explaining why they're necessary. I've edited the code accordingly, so this can now be merged.

Plugins/Creature/Creature.cpp
```diff
@@ -1047,12 +1047,27 @@ NWNX_EXPORT ArgumentStack SetClassByPosition(ArgumentStack&& args)
     {
         const auto position = args.extract<int32_t>();
         const auto classID = args.extract<int32_t>();
+        const auto bUpdateLevels = args.extract<int32_t>();
```
Daztek, line 1050: If you want to be extra fancy you can replace the above with: 

```cpp
int32_t bUpdateLevels = true;
try
{
    bUpdateLevels = args.extract<int32_t>()
}
{
    LOG_WARNING("NWNX_Creature_SetClassByPosition() called from NWScript without bUpdateLevels parameter. Please update nwnx_creature.nss");
}
```

So it doesn't throw when someone calls the function without recompiling their script.

mtijanic: Let them ~~eat cake~~ read changelogs.

PR #1388, opened 2021-07-24
API: Fix cswy engine structure size

### LINKED ISSUE #1387: Possible Invalid Structure mapping - CassowarySolverEngineStructure
When iterating over `std::unordered_map<CExoString, CNWSScriptVar>` and the `CNWSScriptVar` value is resolved, the iterator becomes invalid and crashes the server sometime after.

It looks like the `CassowarySolverEngineStructure` structure is invalid, as appending 4 extra bytes prevents the server crash:

```c++
struct CassowarySolverEngineStructure
{
    std::shared_ptr<void*> m_shared;
    char extra_data[4];
};
```

mtijanic:
Needs to have a correct vtable. This should suffice, hopefully.

Fixes #1387 . @Jorteck could you please verify? Thanks.

### DIFF
--- a/NWNXLib/API/API/CassowarySolverEngineStructure.hpp
+++ b/NWNXLib/API/API/CassowarySolverEngineStructure.hpp
@@ -8,14 +8,18 @@
 NWN_API_PROLOGUE(CassowarySolverEngineStructure)
 #endif
 
-
-
-
-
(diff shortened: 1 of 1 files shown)


Jorteck (approved):
Yep, this resolves the issue. Thanks!

PR #1404, opened 2021-08-21
Use stdout for all error reporting, require it on bug reports

mtijanic:
Policy change: All bug reports now need to have stdout attached.

To avoid confusion, disable `nwnx.txt` and the `nwserver-crash-xxx.log` by default unless overridden via config.
Also allow nwnx.txt to be any file path.


Daztek:
Please update `Core/README.md` too!

PR #1407, opened 2021-08-23
Fixes NPC conversation, adds ON_ITEM_MERGE event.

drake127:
This PR consists of two changes:

1) Fixes recently introduced NPC targeted chat messages. I didn't understand original behavior well and OnConversation event should trigger only in case there's matched pattern.

2) I added ON_ITEM_MERGE events to Events plugin.

### DIFF
--- a/Plugins/Events/Events/ItemEvents.cpp
+++ b/Plugins/Events/Events/ItemEvents.cpp
@@ -24,6 +24,7 @@ static Hooks::Hook s_ItemEventHandlerHook;
 static Hooks::Hook s_UseLoreOnItemHook;
 static Hooks::Hook s_PayToIdenfifyItemHook;
 static Hooks::Hook s_SplitItemHook;
+static Hooks::Hook s_MergeItemHook;
 static Hooks::Hook s_CanUseItemHook;
 static Hooks::Hook s_AcquireItemHook;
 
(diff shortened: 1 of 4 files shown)


Plugins/Events/Events/ItemEvents.cpp
```diff
@@ -468,6 +475,21 @@ void SplitItemHook(CNWSCreature *thisPtr, CNWSItem *pItemToSplit, int32_t nNumbe
 
     PushAndSignal("NWNX_ON_ITEM_SPLIT_AFTER");
 }
+void MergeItemHook(CNWSCreature *thisPtr, CNWSItem *pItemToMergeInto, CNWSItem *pItemToMerge)
+{
+    auto PushAndSignal = [&](const std::string& ev) -> bool {
+        PushEventData("ITEM_TO_MERGE_INTO", Utils::ObjectIDToString(pItemToMergeInto->m_idSelf));
+        PushEventData("ITEM_TO_MERGE", Utils::ObjectIDToString(pItemToMerge->m_idSelf));
```
Daztek, line 482: Won't `pItemToMerge` be a nullptr after `->CallOriginal()` ?

drake127: Hah! Let me check, I think I saw in other events that those could be invalid, but both nullptr or use after free would be bad. I tested only _BEFORE a skip (my bad).

drake127: It leads me to thinking that we have a pointer, possibly invalid, because we don't know if original function destroyed it. And no way how to know it? nullptr check isn't enough as we are likely to use leftover memory. Should I better remove _AFTER event or is there a way?

Daztek: You can just do `auto oidItemToMerge = pItemToMerge->m_idSelf;` before the lambda and push that. It'll be up to the user to deal with an invalid item in the _AFTER, maybe make a note in .nss

drake127: Good idea, I tried to be more safe and return OBJECT_INVALID if the item was merged.

PR #1409, opened 2021-08-24
ELC: add config flag to non-basegame caster primary stat check

### LINKED ISSUE #1408: Racialtypes.2da custom races ability bonus mess with NWNX_ELC checks when picking a class with a primary cast ability that corresponds to a modified ability from the race picked
It seems like that even if the primary cast ability of a class is unchanged (left at 11 during character creation for classes with spells), the ELC from `nwnx_elc` might fail during character creation IF:

1. the class taken is a class for which the primary cast ability corresponds to one of the race buffed abilities (i.e., Paladin has WIS as primary ability and the race also buffs WIS by +2 during character creation) 
2. after the character has taken some RDD levels (which modifies abilities, but i couldn't reproduce easily).

Down here is an example of a custom race with custom stats messing with the ELC check during character creation, even if all the buff values from the race are not higher than 2 (to avoid "other" base game issues), and even if the primary ability is left untouched during character creation:

- _Race:_ `Human` with modified entries on `racialtypes.2da`
- _Bonus:_ `STR +2, DEX +2, WIS +2, CHA +2`
- _Class taken:_ `Paladin`

Here are the stats assigned during character creation:
image

So, `STR` is taken to the the max of 20, all other points are given to `DEX`. `WIS` is left at the default value of 11.
If you try to enter, **ELC FAILS!**

If you try to do the same as above _BUT_ you take `Fighter` instead of `Paladin` (and so, no primary cast ability), even assigning the stats as in the screen above, the character can properly login without issues of any sort!

**Recap:**
It seems like there is a problem with entries in `racialtypes.2da` that alter the value of a primary ability of a class, and the ELC check from `nwnx_elc` that the game makes to check everything is fine. This problem occurs ONLY when you pick a class with a primary cast ability for which the race has a modified value (as for the example above).

Daztek:
Should fix #1408

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ https://github.com/nwnxee/unified/compare/build8193.30...HEAD
 
 ### Added
 - Events: Added NWNX_ON_ITEM_MERGE_BEFORE and NWNX_ON_ITEM_MERGE_AFTER.
+- ELC: added `NWNX_ELC_ENFORCE_CASTER_PRIMARY_STAT_IS_11`, if enabled, ELC will check when a character's first level class is a spellcaster, if their primary casting stat is >= 11.
 
 ##### New Plugins
 - N/A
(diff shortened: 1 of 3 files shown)


TomRotfl:
To be extremely precise, should it checks against `CHARGEN_BASE_ABILITY_MIN_PRIMARY` from `rulesets.2da`, instead of a constant `11`?

PR #1411, opened 2021-08-28
Race: Add support for specifying armor class types in the AC Modifier

minikanyas:
This PR consists of one change:

1. Add support for specifying armor class types in the Race plugin AC modifier.


Daztek:
Thank you for your PR! Just a couple of things:

1) Please update the changelog too!
2) I assume this doesn't break people's stuff if they don't change their 2das/scripts?
3) You should probably note that non-dodge AC effects won't stack. So if your race gives you 2 shield ac and you equip a tower shield, you'll only have +3 shield ac.

minikanyas:
> Thank you for your PR! Just a couple of things:
> 
> 1. Please update the changelog too!
> 2. I assume this doesn't break people's stuff if they don't change their 2das/scripts?
> 3. You should probably note that non-dodge AC effects won't stack. So if your race gives you 2 shield ac and you equip a tower shield, you'll only have +3 shield ac.

1. Done.
2. I've tested this and it shouldn't break anything, they will get a warning message indicating that the specified AC type is invalid and that it will be falling back to Dodge.
3. Shouldn't this stack? If I'm not wrong, the shield base AC stacks with effects that use shield AC bonus (or at least with item properties)

stenly87:
I think, in non-dodge AC no sense. Even dodge AC has limit (+20). If you want to add more AC (and flat-footed state too), you can add AC with GetACNaturalBase func.

Daztek:
>     3. Shouldn't this stack? If I'm not wrong, the shield base AC stacks with effects that use shield AC bonus (or at least with item properties). But still, I'm aware that some AC types prevent stacking and this was one of the reasons I made this PR.

Yeah, bad example from my part, that case should work.

Plugins/Race/README.md
```diff
@@ -125,18 +125,19 @@ NWNX_Race_SetRacialModifier(150, NWNX_RACE_MODIFIER_ABVSRACE, RACIAL_TYPE_DRAGON
 The `AC` entry is used to modify the base AC for a race.
```
Daztek, line 126: A note here that non-dodge AC types won't stack with other non-dodge AC from spells effects, would be good

minikanyas: Done

Issue #899, opened 2020-05-13
[api/cresref] [GCC 9.3] -Wdeprecated-copy warnings for implicit copy constructor


jd28:
Warning's here: https://github.com/nwnxee/unified/blob/6d60d138a8a5122edbc5c63bb674cb398c16d8dc/Plugins/MaxLevel/MaxLevel.cpp#L202

Not sure there's much NWNX can do about it, but fyi.

jd28:
There's also an issue with this `typedef uint8_t RESREF[16];` in CResRef.hpp in some cases.  I don't get why this wouldn't be `typedef char RESREF[16];`  Character arrays/strings in C & C++ aren't unsigned.

mtijanic:
>  Character arrays/strings in C & C++ aren't unsigned.

It's surprisingly more complicated than that. `char` in C can have either the range of `unsigned char` or `signed char`, it's implementation defined. GCC offers a switch to choose. However, it's a different type from either of those. So, whether it's signed or unsigned, `char`, `signed char` and `unsigned char` are three different types.

jd28:
Never knew that.  Thank you for taking time explain!

jd28:
Maybe try ‘CResRef(const CResRef& other) = default’?  Only replying since I
got emails about this.  I'm wishing you all pushing this game forward the
best.

I'll add an old idea, since i'm probably banned most places, that maybe you
can convince someone at beamdog to add "extern funcdecl(parameter_list)" to
nwscript, function name at the top of the stack, op code CALL_EXTERN..
CNWSWhatever::RegisterExternalCommand, you all will know what i mean.  Then
the "final refactor".

As above, all the best.

PR #1415, opened 2021-09-12
Build8193.31

mtijanic:
Much WIP, but we can start testing early this way.


Daztek:
Fix for #899 is deferred

PR #1419, opened 2021-09-21
add NWNX_Util_GetModuleMtime()

hendrikgit:
(Edited. Previously this was about getting a hash of the module file.)

The idea is to compare the last modified timestamp (mtime) of the currently running module with a stored one and only run certain actions if the module changed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,7 @@ https://github.com/nwnxee/unified/compare/build8193.32...HEAD
 - N/A
 
 ##### New NWScript Functions
-- N/A
+- Util: GetModuleMtime()
 
 ### Changed
 - N/A
--- a/Plugins/Util/Util.cpp
+++ b/Plugins/Util/Util.cpp
@@ -14,6 +14,7 @@
 #include "API/CWorldTimer.hpp"
 #include "API/CGameObjectArray.hpp"
 #include "API/CScriptCompiler.hpp"
+#include "API/CExoAliasList.hpp"
 #include "API/CExoFile.hpp"
 #include "API/CNWSDoor.hpp"
 #include "API/CResStruct.hpp"
@@ -27,6 +28,7 @@
 #include <cmath>
 #include <chrono>
 #include <unistd.h>
+#include <sys/stat.h>
 
 using namespace NWNXLib;
 using namespace NWNXLib::API;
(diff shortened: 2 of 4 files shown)


Plugins/Util/Util.cpp
```diff
@@ -126,6 +128,24 @@ NWNX_EXPORT ArgumentStack Hash(ArgumentStack&& args)
 }
 }
 
+NWNX_EXPORT ArgumentStack GetModuleMtime(ArgumentStack&&)
+{
+    CNWSModule *pMod = Utils::GetModule();
+    if (pMod->m_bIsSaveGame) {
```
Daztek, line 134: Brackets on their own line, please!

Plugins/Util/Util.cpp
```diff
@@ -126,6 +128,24 @@ NWNX_EXPORT ArgumentStack Hash(ArgumentStack&& args)
...
+        LOG_DEBUG("NWNX_GetModuleMtime() module is a save game, returning 0");
+        return 0;
+    }
+    std::string modName = pMod->m_sModuleResourceName.SubString(12); // discard "CURRENTGAME:"
+    std::string modPath = Globals::ExoBase()->m_pcExoAliasList->ResolveFileName("MODULES:" + modName, Constants::ResRefType::MOD);
+    struct stat stat;
+    if (lstat(modPath.c_str(), &stat) == 0) {
```
Daztek, line 141: I think we'd want `stat` instead of `lstat`

From the [stat manpage](https://man7.org/linux/man-pages/man2/lstat.2.html):

```
       stat() and fstatat() retrieve information about the file pointed
       to by pathname; the differences for fstatat() are described
       below.

       lstat() is identical to stat(), except that if pathname is a
       symbolic link, then it returns information about the link itself,
       not the file that the link refers to.
```

Also bracket again!

CHANGELOG.md
```diff
@@ -14,7 +14,7 @@ https://github.com/nwnxee/unified/compare/build8193.31...HEAD
 - N/A
 
 ##### New NWScript Functions
-- N/A
+- Util: NWNX_Util_GetModuleMtime()
```
Daztek, line 17: Super minor: 

`- Util: GetModuleMtime()`

hendrikgit:
Thanks Daz. I made the changes and gave it a quick test. Should be good to go now.

PR #1442, opened 2021-11-17
Ensure builds run sequentially

Jorteck:
If a build running on a newer commit finishes before an older commit, the older commit will be incorrectly tagged as the latest build.

This PR should prevent newer builds from running until the previous one finishes:
https://docs.github.com/en/actions/learn-github-actions/workflow-syntax-for-github-actions#concurrency


Daztek:
If this breaks stuff, you get to fix it!

PR #1447, opened 2021-11-29
Add ability to change alias folder in NWNX_Object_Export

stenly87:


Plugins/Object/NWScript/nwnx_object.nss
```diff
@@ -206,7 +206,8 @@ void NWNX_Object_RemoveIconEffect(object obj, int nIcon);
 /// @brief Export an object to the UserDirectory/nwnx folder.
 /// @param sFileName The filename without extension, 16 or less characters.
 /// @param oObject The object to export. Valid object types: Creature, Item, Placeable, Waypoint, Door, Store, Trigger
-void NWNX_Object_Export(object oObject, string sFileName);
+/// @param sAlias The alias of the resource directory to add the .git file to. Default: UserDirectory/nwnx
+void NWNX_Object_Export(object oObject, string sFileName, string sAlias = "NWNX");
```
Mingun, line 210: Linux has case-sensitive filesystems by default. Are you sure that the default directory will be `nwnx`, not `NWNX`?

Plugins/Object/NWScript/nwnx_object.nss
```diff
@@ -641,10 +642,11 @@ void NWNX_Object_RemoveIconEffect(object obj, int nIcon)
     NWNX_CallFunction(NWNX_Object, sFunc);
 }
 
-void NWNX_Object_Export(object oObject, string sFileName)
+void NWNX_Object_Export(object oObject, string sFileName, string sAlias = "NWNX")
 {
     string sFunc = "Export";
-
+	
+	NWNX_PushArgumentString(sAlias);
```
Mingun, line 649: Here is tab instead of spaces

Plugins/Object/Object.cpp
```diff
@@ -526,13 +526,22 @@ NWNX_EXPORT ArgumentStack Export(ArgumentStack&& args)
         const auto fileName = args.extract<std::string>();
           ASSERT_OR_THROW(!fileName.empty());
           ASSERT_OR_THROW(fileName.size() <= 16);
+		auto alias = args.extract<std::string>();
```
Mingun, line 529: Here tab too

Plugins/Object/NWScript/nwnx_object.nss
stenly87: yeap. all folders in nwn are always upcased

Plugins/Object/Object.cpp
```diff
@@ -526,13 +526,22 @@ NWNX_EXPORT ArgumentStack Export(ArgumentStack&& args)
         const auto fileName = args.extract<std::string>();
           ASSERT_OR_THROW(!fileName.empty());
           ASSERT_OR_THROW(fileName.size() <= 16);
+        auto alias = args.extract<std::string>();
```
Daztek, line 529: Needs to be the following so it doesn't break prior compiled scripts:

```c++
std::string alias;
try
{
    alias = args.extract<std::string>();
}
{
    LOG_WARNING("NWNX_Object_Export() called from NWScript without sAlias parameter. Please update nwnx_object.nss");
    alias = "NWNX";
}
```

stenly87: Thanks

Daztek:
Thank you for your contribution!

PR #1455, opened 2021-12-19
Update base.Dockerfile to dotnet 6 packages

urothis:
No regressions noted in testing. 
Backwards compatible with old projects and no breakages.

### DIFF
--- a/base.Dockerfile
+++ b/base.Dockerfile
@@ -17,8 +17,8 @@ RUN runDeps="hunspell \
     inotify-tools \
     patch \
     unzip \
-    dotnet-runtime-5.0 \
-    dotnet-apphost-pack-5.0" \
+    dotnet-runtime-6.0 \
+    dotnet-apphost-pack-6.0" \
     installDeps="ca-certificates wget" \
     && apt-get update \
     && apt-get install -y --no-install-recommends $installDeps \


Daztek:
You'll be responsible if it breaks!

zunath:
FYI - this did cause a break. You can't simply remove an older version from the image like this. Luckily in this case we're going to upgrade, but please be careful in the future.

```
swlor-server_1   | The framework 'Microsoft.NETCore.App', version '5.0.0' (x64) was not found.
swlor-server_1   |   - The following frameworks were found:
swlor-server_1   |       6.0.1 at [/usr/share/dotnet/shared/Microsoft.NETCore.App]
swlor-server_1   |
swlor-server_1   | You can resolve the problem by installing the specified framework and/or SDK.
swlor-server_1   |
swlor-server_1   | The specified framework can be found at:
swlor-server_1   |   - https://aka.ms/dotnet-core-applaunch?framework=Microsoft.NETCore.App&framework_version=5.0.0&arch=x64&rid=debian.10-x64
```

urothis:
@zunath Sorry about that, noted for next time though.

We could include multiple dotnet versions in the image, but that will increase image size a smidge.

urothis:
I did some prelim research on what it would look like to add multiple dotnet versions

```
nwnxee-dotnet6 : 690MB
nwnxee-dotnet3.1-6 : 845MB
```

Adding all supported versions would add ~150MB to the image size.

zunath:
We chatted on Discord briefly about this, but I'm not too bothered with forcing an upgrade. I'd just prefer a grace period that it's coming. We've got a few people running local servers at this point and trying to avoid issues if there's suddenly a .NET upgrade. 

I don't think it's necessary to include the other version in the image at this point - seems like a waste.

PR #1460, opened 2021-12-29
Events: allow script chunks to subscribe to events

Daztek:
Really niche, but script chunks can now subscribe to events.


Daztek:
Also adds `NWNX_Events_GetNumSubscribers()`

PR #1466, opened 2022-01-16
DotNET: Cleanup and Refactoring

Jorteck:

PR #1469, opened 2022-01-18
DotNET: Add raw string stack methods

Jorteck:
Merge first: https://github.com/nwnxee/unified/pull/1466

### DIFF
--- a/Plugins/DotNET/DotNETExports.cpp
+++ b/Plugins/DotNET/DotNETExports.cpp
@@ -229,6 +229,24 @@ static void StackPushString(const char* value)
     }
 }
 
+static void StackPushRawString(const char* value)
+{
+    auto vm = Globals::VirtualMachine();
+    ASSERT(vm->m_nRecursionLevel >= 0);
+
+    LOG_DEBUG("Pushing string '%s'.", value);
+    CExoString str(value);
+
+    if (vm->StackPushString(str))
+    {
+      ++s_pushedCount;
+    }
+    else
+    {
+      LOG_WARNING("Failed to push string '%s' - recursion level %i.", str.m_sString, vm->m_nRecursionLevel);
+    }
+}
+
 static void StackPushObject(uint32_t value)
 {
     auto vm = Globals::VirtualMachine();
@@ -327,6 +345,24 @@ static const char* StackPopString()
 
     // TODO: Less copies
     return strdup(String::ToUTF8(value.CStr()).c_str());
+}
+
+static const char* StackPopRawString()
+{
+    auto vm = Globals::VirtualMachine();
+    ASSERT(vm->m_nRecursionLevel >= 0);
+
+    CExoString value;
+    if (!vm->StackPopString(&value))
+    {
+      LOG_WARNING("Failed to pop string - recursion level %i.", vm->m_nRecursionLevel);
+      return nullptr;
+    }
+
+    static std::string retVal = value.CStr();
+    LOG_DEBUG("Popped string '%s'.", value.m_sString);
+
+    return retVal.c_str();
 }
 
 static uint32_t StackPopObject()
@@ -452,6 +488,11 @@ static void NWNXPushString(const char* s)
     Events::Push(String::FromUTF8(s));
 }
 
+static void NWNXPushRawString(const char* s)
+{
+    Events::Push<std::string>(s);
+}
+
 static void NWNXPushEffect(CGameEffect* e)
 {
     Events::Push(e);
@@ -481,6 +522,20 @@ static const char* NWNXPopString()
 {
     auto str = Events::Pop<std::string>().value_or(std::string{""});
     return strdup(String::ToUTF8(str).c_str());
+}
+
+static const char* NWNXPopRawString()
+{
+    auto value = Events::Pop<std::string>();
+    if (value.has_value())
+    {
+      static auto retVal = value.value();
+      return retVal.c_str();
+    }
+    else
+    {
+      return nullptr;
+    }
 }
 
 static CGameEffect* NWNXPopEffect()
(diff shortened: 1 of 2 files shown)


Plugins/DotNET/DotNETExports.cpp
```diff
@@ -331,6 +346,24 @@ static const char* StackPopString()
...
+    if (!vm->StackPopString(&value))
+    {
+      LOG_WARNING("Failed to pop string - recursion level %i.", vm->m_nRecursionLevel);
+      return nullptr;
+    }
+
+    static std::string retVal = value.CStr();
+    LOG_DEBUG("Popped string '%s'.", value.m_sString);
+
+    return retVal.c_str();
```
mtijanic, line 364: Either this one or `StackPopString()` above is wrong. The previous one assumes dotnet takes ownership of the pointer to heap memory and will free it. This one will still return a pointer to heap memory, but it does not relinquish the ownership. So second time it's called, when `= value.CStr()` is run, it might try to use or free the old memory, causing a boom.

Alternatively, the previous assumption from `StackPopString()` is wrong and it actually leaks memory, in which case it should use a `static` variable like here instead of `strdup()`. Could you verify which one it is?

Plugins/DotNET/DotNETExports.cpp
```diff
@@ -331,6 +346,24 @@ static const char* StackPopString()
...
+static const char* StackPopRawString()
+{
+    auto vm = Globals::VirtualMachine();
+    ASSERT(vm->m_nRecursionLevel >= 0);
+
+    CExoString value;
+    if (!vm->StackPopString(&value))
+    {
+      LOG_WARNING("Failed to pop string - recursion level %i.", vm->m_nRecursionLevel);
+      return nullptr;
```
mtijanic, line 358: Use `""` instead of `nullptr`. Although I now see above that it might need to be `strdup("")` instead.

Plugins/DotNET/DotNETExports.cpp
```diff
@@ -428,61 +461,93 @@ static int32_t ClosureActionDoCommand(uint32_t oid, uint64_t eventId)
...
+{
+    auto value = Events::Pop<std::string>();
+    if (value.has_value())
+    {
+      static auto retVal = value.value();
+      return retVal.c_str();
+    }
+    else
+    {
+      return nullptr;
```
mtijanic, line 536: As above, `""` instead of `nullptr`. But this branch goes away if you use `value_or()`

Plugins/DotNET/DotNETExports.cpp
```diff
@@ -428,61 +461,93 @@ static int32_t ClosureActionDoCommand(uint32_t oid, uint64_t eventId)
...
+static const char* NWNXPopString()
 {
     auto str = Events::Pop<std::string>().value_or(std::string{""});
     return strdup(String::ToUTF8(str).c_str());
 }
-static CGameEffect* nwnxPopEffect()
+
+static const char* NWNXPopRawString()
+{
+    auto value = Events::Pop<std::string>();
```
mtijanic, line 528: Go like above, with `.value_or("")`

Plugins/DotNET/DotNETExports.cpp
```diff
@@ -538,27 +603,31 @@ std::vector<void*> GetExports()
...
-    exports.push_back((void*)&nwnxPopString); // reserved utf8
-    exports.push_back((void*)&nwnxPopEffect);
-    exports.push_back((void*)&nwnxPopItemProperty);
-    exports.push_back((void*)&nwnxCallFunction);
+    exports.push_back((void*)&NWNXSetFunction);
+    exports.push_back((void*)&NWNXPushInt);
+    exports.push_back((void*)&NWNXPushFloat);
+    exports.push_back((void*)&NWNXPushObject);
+    exports.push_back((void*)&NWNXPushString);
+    exports.push_back((void*)&NWNXPushString); // reserved utf8
```
mtijanic, line 611: Rather than adding new ones, you should re-use these four reserved slots. This is exactly what they were for.

Plugins/DotNET/DotNETExports.cpp
```diff
@@ -226,8 +224,25 @@ static void StackPushString(const char* value)
...
+{
+    auto vm = Globals::VirtualMachine();
+    ASSERT(vm->m_nRecursionLevel >= 0);
+
+    LOG_DEBUG("Pushing string '%s'.", value);
+    CExoString str(value);
+
+    if (vm->StackPushString(str))
+    {
+      ++s_pushedCount;
```
mtijanic, line 241: nit, but this change has inconsistent indentation. Would be a lot less of an issue if it wasn't bundled with a massive reformat :P

Plugins/DotNET/DotNETExports.cpp
```diff
@@ -329,6 +347,24 @@ static const char* StackPopString()
...
+    ASSERT(vm->m_nRecursionLevel >= 0);
+
+    CExoString value;
+    if (!vm->StackPopString(&value))
+    {
+      LOG_WARNING("Failed to pop string - recursion level %i.", vm->m_nRecursionLevel);
+      return nullptr;
+    }
+
+    static std::string retVal = value.CStr();
```
mtijanic, line 362: By the way, if the answer to the other question is "strdup() leaks memory, need to use static retVal", then just make the `CExoString` the static variable and avoid this extra copy here.

Plugins/DotNET/DotNETExports.cpp
Jorteck: See my above comment. The intention is that null pointers are handled by the managed code.

Plugins/DotNET/DotNETExports.cpp
Jorteck: As above, the intention is that null pointers are handled by the managed code.

Plugins/DotNET/DotNETExports.cpp
Jorteck: As above, the intention is that null pointers are handled by the managed code.

Plugins/DotNET/DotNETExports.cpp
mtijanic: Ah, I missed that part. Alright then, sounds good. But then it does need four more entries and can't use the reserved UTF8 ones.

Just keep in mind that you have to copy the string before anyone tries to pop a second one, because the buffer is shared.

PR #1476, opened 2022-01-28
Add FIX_TRIGGER_ENTER_DETECTION tweak

Jorteck:
I initially tried using `CNWSTrigger::InTrigger` for this check and it fixed most cases where this happens, but there were still some cases where the object position was way outside the trigger bounds, and `CNWSTrigger::InTrigger` still returned true.

The custom `VectorInTriggerBounds` implementation seemed to work for all the cases I tested, but I wonder if there's any other caveats I am missing :)

### DIFF
--- a/Plugins/Tweaks/FixTriggerEnterDetection.cpp
+++ b/Plugins/Tweaks/FixTriggerEnterDetection.cpp
@@ -1,0 +1,63 @@
+#include "nwnx.hpp"
+
+#include "API/CNWSTrigger.hpp"
+#include "API/CScriptEvent.hpp"
+
+namespace Tweaks {
+
+using namespace NWNXLib;
+using namespace NWNXLib::API;
+using namespace NWNXLib::API::Constants;
+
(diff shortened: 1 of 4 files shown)


Plugins/Tweaks/FixTriggerEnterDetection.cpp
```diff
@@ -0,0 +1,63 @@
...
+    int i, j;
+    bool inside = false;
+
+    for (i = 0, j = numVertices - 1; i < numVertices; j = i++)
+    {
+        if ((vertices[i].y >= point.y ) != (vertices[j].y >= point.y))
+        {
+            if ((point.x <= (vertices[j].x - vertices[i].x) * (point.y - vertices[i].y) / (vertices[j].y - vertices[i].y) + vertices[i].x))
+            {
+                inside = !inside;
```
mtijanic, line 26: I can't begin to wrap my head around this, so I'll take your word for it.

PR #1481, opened 2022-02-13
DotNET: Add null checks for script/loop hooking. Add ON_NWNX_LOADED signal

Jorteck:
Adds some additional null checks for the bootstrapped `AllHandlers` structure.

For null function pointers, the related function will not be hooked. This adds support for optionally picking which functions should be hooked (e.g. hook the VM run script callback, but not the server loop).

This also adds a new core signal `ON_NWNX_LOADED` which is invoked immediately after NWNX and all plugins have loaded. This is to resolve a dependency issue when trying to use SWIG in the initial plugin load, and felt like a nicer change compared to renaming the SWIG plugin :)

### DIFF
--- a/Plugins/DotNET/DotNETExports.cpp
+++ b/Plugins/DotNET/DotNETExports.cpp
@@ -135,17 +135,16 @@ static void RegisterHandlers(AllHandlers* handlers, unsigned size)
     MessageBus::Subscribe("NWNX_CORE_SIGNAL",
         [](const std::vector<std::string>& message)
         {
-            // We will crash the server if we try to create a script context after the server is destroyed.
-            if (message[0] == "ON_DESTROY_SERVER_AFTER")
-            {
-                s_handlers.SignalHandler(message[0].c_str());
-            }
-            else
+            if (API::Globals::VirtualMachine())
             {
                 int spBefore = Utils::PushScriptContext(Constants::OBJECT_INVALID, 0, false);
                 s_handlers.SignalHandler(message[0].c_str());
                 int spAfter = Utils::PopScriptContext();
                 ASSERT_MSG(spBefore == spAfter, "spBefore=%x, spAfter=%x", spBefore, spAfter);
+            }
+            else
+            {
+                s_handlers.SignalHandler(message[0].c_str());
             }
         });
 }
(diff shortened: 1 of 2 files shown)


Plugins/DotNET/DotNETExports.cpp
```diff
@@ -150,6 +146,42 @@ static void RegisterHandlers(AllHandlers* handlers, unsigned size)
...
+        LOG_WARNING("RegisterHandlers argument missing some entries - Managed/Unmanaged mismatch, update managed code");
+    }
+
+    LOG_INFO("Registering managed code handlers.");
+    s_handlers = *handlers;
+
+    if (s_handlers.MainLoop)
+    {
+        RegisterMainLoopHandler();
+    }
```
mtijanic, line 167: I don't really feel too strongly about styles here, but:
* Please keep nonfunctional refactors to individual commits. Whether these are inlined or kept in separate functions is a pure preference, and we can discuss which is better, but it makes the functional part harder to review.
* If you are insisting on keeping it as a separate function, then the null checks should be in those functions, not here. And the functions should be passed as arguments, not globals.

mtijanic:
Sorry the review took so long, but I don't really have a lot of time during the week for this, so any time I see a non-trivial change I defer to the weekend. Hence my insistence on keeping things as separate commits (ideally as separate PRs!).

Plugins/DotNET/DotNETExports.cpp
```diff
@@ -16,17 +16,17 @@ using namespace NWNXLib::API;
...
-using MainLoopHandlerType  = void (*)(uint64_t);
+using MainLoopHandlerType = void (*)(uint64_t);
 using RunScriptHandlerType = int (*)(const char*, uint32_t);
 using ClosureHandlerType = void (*)(uint64_t, uint32_t);
 using SignalHandlerType = void (*)(const char*);
 
 struct AllHandlers
 {
-    MainLoopHandlerType  MainLoop;
+    MainLoopHandlerType MainLoop;
```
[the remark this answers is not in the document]
Jorteck: No preference, but I couldn't see a clear style/convention when looking at other plugins. Some align, others declare them without extra padding.

I'll remove the change if the preference is to align them :)

Plugins/DotNET/DotNETExports.cpp
Jorteck: I'll break out the changes, and hopefully find a middle ground. I was called out on early PRs for making too many commits :P

PR #1489, opened 2022-03-06
Add 'Decrement Spell Count' event

WilliamDraco:
As it says on the tin. Non-skippable event for when the spell-count decreases after casting a spell, or having it interrupted.


WilliamDraco:
Adjusted per conversation with Daz to be skipable.

Arvirago:
Skipping this event will just not decrement the spell count?
I have a class with infinite spells, and right now I have to give back all spells in the spellhook.

WilliamDraco:
> Skipping this event will just not decrement the spell count? I have a class with infinite spells, and right now I have to give back all spells in the spellhook.

That's correct - Skipping this event causes the spell to proceed as normal, but the spell does not become unready (Wiz-like) or consume a spell slot (Sorc-like)

PR #1494, opened 2022-03-28
Creature: add {Get|Set}SkillPointsRemainingByLevel()

summonFox:
As advertised on the tin.

### DIFF
--- a/Plugins/Creature/Creature.cpp
+++ b/Plugins/Creature/Creature.cpp
@@ -1269,6 +1269,46 @@ NWNX_EXPORT ArgumentStack SetSkillPointsRemaining(ArgumentStack&& args)
     return {};
 }
 
+NWNX_EXPORT ArgumentStack GetSkillPointsRemainingByLevel(ArgumentStack&& args)
+{
+    if (auto *pCreature = Utils::PopCreature(args))
+    {
+        const auto level = args.extract<int32_t>();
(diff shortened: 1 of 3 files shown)


Plugins/Creature/Creature.cpp
```diff
@@ -1269,6 +1269,46 @@ NWNX_EXPORT ArgumentStack SetSkillPointsRemaining(ArgumentStack&& args)
...
+        const auto value = args.extract<int32_t>();
+          ASSERT_OR_THROW(value >= 0);
+          ASSERT_OR_THROW(value <= 65535);
+
+        if (level > 0 && level <= pCreature->m_pStats->m_lstLevelStats.num)
+        {
+            auto *pLevelStats = pCreature->m_pStats->m_lstLevelStats.element[level-1];
+            ASSERT_OR_THROW(pLevelStats);
+
+            pLevelStats->m_nSkillPointsRemaining = static_cast<uint8_t>(value);
```
Daztek, line 1306: This should be a cast to `uint16_t` since `m_nSkillPointsRemaining` is an `uint16_t`

Daztek (commented):
Hi! Thank you for your contribution, just one minor thing, the rest looks good!

Plugins/Creature/Creature.cpp
summonFox: Ah yeah, woops. Good catch, thanks!

PR #1495, opened 2022-03-28
Events: Add event data to various NWNX_ON_DM_* events

summonFox:
Adds event data to:
- NWNX_ON_DM_SET_STAT_{BEFORE|AFTER}
- NWNX_ON_DM_GET_VARIABLE_{BEFORE|AFTER}
- NWNX_ON_DM_SET_VARIABLE_{BEFORE|AFTER}
- NWNX_ON_DM_SET_FACTION_{BEFORE|AFTER}

### DIFF
--- a/Plugins/Events/NWScript/nwnx_events.nss
+++ b/Plugins/Events/NWScript/nwnx_events.nss
@@ -479,21 +479,71 @@ _______________________________________
     PASSWORD              | string | The password the DM provided, only valid for NWNX_ON_DM_PLAYERDM_LOGIN_*
 
 _______________________________________
+    ## DM Set Stat Events
+    - NWNX_ON_DM_SET_STAT_BEFORE
+    - NWNX_ON_DM_SET_STAT_AFTER
+
+    `OBJECT_SELF` = The DM
+
+    Event Data Tag        | Type   | Notes
+    ----------------------|--------|-------
+    STAT                  | int    | Returns NWNX_EVENTS_DM_SET_STAT_TYPE_*
+    VALUE                 | int    | 
+    TARGET                | object | Convert to object with StringToObject()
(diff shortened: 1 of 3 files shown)


Plugins/Events/NWScript/nwnx_events.nss
```diff
@@ -1459,6 +1509,22 @@ const int NWNX_EVENTS_TIMING_BAR_LOCK          = 8;
 const int NWNX_EVENTS_TIMING_BAR_CUSTOM        = 10;
 */
 
+/*
+const int NWNX_EVENTS_DM_SET_STAT_TYPE_STRENGTH         = 5;
+const int NWNX_EVENTS_DM_SET_STAT_TYPE_DEXTERITY        = 6;
+const int NWNX_EVENTS_DM_SET_STAT_TYPE_CONSTITUTION     = 7;
+const int NWNX_EVENTS_DM_SET_STAT_TYPE_INTELLIGENCE     = 8;
+const int NWNX_EVENTS_DM_SET_STAT_TYPE_WISDOM           = 9;
+const int NWNX_EVENTS_DM_SET_STAT_TYPE_CHARISMA         = 10;
```
Daztek, line 1518: If you subtract 5 from the stat value and push that, you can just use the basegame `ABILITY_*` constants

Daztek (commented):
Looks good, just one minor thing noted below. Hope you tested it, because I haven't!

Plugins/Events/NWScript/nwnx_events.nss
summonFox: Can do. Wasn't sure if you'd want the values to be left as is given by nwserver for the scripter to convert in nwscript themselves or not, -5 it is!

PR #1497, opened 2022-03-28
Added SpellId, school and spell level parameters to the NWNX_Object_D…

BhaalM:
…oSpellLevelAbsorption function

### DIFF
--- a/Plugins/Object/Object.cpp
+++ b/Plugins/Object/Object.cpp
@@ -822,11 +822,23 @@ NWNX_EXPORT ArgumentStack DoSpellLevelAbsorption(ArgumentStack&& args)
     {
         if(auto *pVersus = Utils::PopObject(args))
         {
-            const auto spellId     = args.extract<int32_t>();
-            const auto spellLevel  = args.extract<int32_t>();
-              ASSERT_OR_THROW(spellLevel <= 10);
-            const auto spellSchool = args.extract<int32_t>();
-              ASSERT_OR_THROW(spellSchool <= 8);
+            int32_t spellId, spellLevel, spellSchool;
+            try
+            {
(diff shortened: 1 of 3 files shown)


Plugins/Object/Object.cpp
```diff
@@ -820,7 +821,67 @@ NWNX_EXPORT ArgumentStack DoSpellLevelAbsorption(ArgumentStack&& args)
     if (auto *pObject = Utils::PopObject(args))
     {
         if(auto *pVersus = Utils::PopObject(args))
-            return pObject->DoSpellLevelAbsorption(pVersus);
+        {
+            const auto spellId     = args.extract<int32_t>();
+            const auto spellLevel  = args.extract<int32_t>();
+              ASSERT_OR_THROW(spellLevel <= 10);
+            const auto spellSchool = args.extract<int32_t>();
+              ASSERT_OR_THROW(spellSchool <= 8);
```
Daztek, line 829: If you do something like this:
```c
int32_t spellId, spellLevel, spellSchool;        
try
{
    spellId = args.extract<int32_t>();
    spellLevel = args.extract<int32_t>();
    spellSchool = args.extract<int32_t>();
}
{
    LOG_WARNING("NWNX_Object_DoSpellLevelAbsorption() called from NWScript without new parameters. Please update nwnx_object.nss");
    spellId = -1;
    spellLevel = -1;
    spellSchool = -1;
}

ASSERT_OR_THROW(spellLevel <= 10);
ASSERT_OR_THROW(spellSchool <= 8);
```

it won't break the ABI and previously compiled scripts.

BhaalM: Nice! didn't know we can do this.

PR #1498, opened 2022-04-02
Race: Fix feat usages being reset upon character relog

### LINKED ISSUE #1412: [NWNX_Races] - FEATUSAGE parameter not resetting usages of specified feats when a character relogs
It seems like there is a problem in `nwnx_races` and its `FEATUSAGE` function. (unless the problem is with the way i set it up)

Let me explain:

I have a custom race receiving a `harm` feat at level 10. Also, the usages of that feat do increase by 1 each 8 character levels by using the `FEATUSAGE` parameter from `nwnx_races`. 
To make it work, i had to set the uses per day of that feat to 99 into `feats.2da` (setting `****` or `1` or a lower value than the maximum reachable (`1 + 1 each 8 levels = 6`), did not work for me).

The function seem to work fine, meaning that at lvl 10 i receive the feat `harm`, it has 2 uses per day `1 + 1 each 8 levels = 2 uses at character level 10`.
The problem comes when you relog. after relog uses per day are set to 99 automatically (the value from `feats.2da`, and the only way to have them set back to the right value is to have the character rest.

Is there anyway to work around this problem? Maybe the `nwnx_race` should check `OnClientEnter` as it does for other race effects and set uses of the feat accordingly?

Here is the `racemod.2da` for that race, in case you are wondering:
```
2DA V2.0

    LABEL                       Type            Param1  Param2  Param3
[....other bonus i took out since out of the scope of this message....]    
14  HarmSelf                    FEAT            1216    10      ****
15  HarmSelfUsages        FEATUSAGE  1216    1       8
```

plenarius:
Fixes #1412 

To bypass character validation with `NWNX_Race` feats, we drop the feat and add it back. This resets their feat usage though so we have to make sure to store the current feat usage then set it back.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -48,6 +48,7 @@ https://github.com/nwnxee/unified/compare/build8193.34...HEAD
 - N/A
 
 ### Fixed
+- Feat: Fixed an issue with feat usages being reset upon character relog
 - Object: GetLocalVariable() now recognizes variables of type json.
 - Tweaks: Language override tweak now works for area names.
 - Events: Fixed a crash when skipping `NWNX_ON_CLIENT_CONNECT_BEFORE`
--- a/Plugins/Race/Race.cpp
+++ b/Plugins/Race/Race.cpp
@@ -644,6 +644,9 @@ int32_t Race::ValidateCharacterHook(CNWSPlayer *pPlayer, int32_t *bFailedServerR
 
     auto nRace = pCreature->m_pStats->m_nRace;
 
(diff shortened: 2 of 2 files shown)


Arvirago:
Why don't you check in elc event if the race and feat is correct and then allow the PC?

plenarius:
Thanks so much for the feedback @Arvirago - it's always been the stance that if we can keep plugins independent of other ones then we should do so. To that end, controlling the character validation from the Feat plugin side instead of Events/ELC is the better route. This method used in the C++ side also saves builders time from writing custom ELC bypasses for every custom racial feat they add.

PR #1504, opened 2022-04-10
Events: expose RunScriptEvent() and make skippable

### LINKED ISSUE #1503: [FEATURE REQUEST] Event that fires on each area enter/exit independent of area script
I have some code that I want to run for each area enter and exit, for all the areas, always. Of course I can make sure that all new and modified areas always have the correct scripts set (and I do) but it would be great to just have two global events.

Maybe NWNX_ON_AREA_ENTER and NWNX_ON_AREA_EXIT.
OBJECT_SELF could be the area and the player be in the event data.

niv:
Fixes #1503

This reuses the Get/SetEventScript constant scheme.

* [x] Tested it locally.
* [x] Have a ID whitelist on the event type.

### DIFF
--- a/Plugins/Events/Events/EventScriptEvents.cpp
+++ b/Plugins/Events/Events/EventScriptEvents.cpp
@@ -1,0 +1,92 @@
+#include "Events.hpp"
+
+#include "API/CNWSArea.hpp"
+#include "API/CNWSModule.hpp"
+#include "API/CNWSObject.hpp"
+
+namespace Events {
+
+using namespace NWNXLib::API;
+using namespace NWNXLib::API::Constants;
+
+// BOOL CNWSArea::RunEventScript(int32_t nScript, CExoString* psOverrideScriptName)
+static NWNXLib::Hooks::Hook s_runAreaEventScript;
+// BOOL CNWSModule::RunEventScript(int32_t nScript, CExoString* psOverrideScriptName)
+static NWNXLib::Hooks::Hook s_runModuleEventScript;
+// BOOL CNWSObject::RunEventScript(int32_t nScript, CExoString* psOverrideScriptName)
+static NWNXLib::Hooks::Hook s_runObjectEventScript;
+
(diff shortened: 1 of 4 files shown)


Plugins/Events/NWScript/nwnx_events.nss
```diff
@@ -1622,6 +1633,7 @@ string NWNX_Events_GetEventData(string tag);
 /// - Input Drop Item
 /// - Decrement Spell Count event
 /// - Play Visual Effect event
+/// - EventScript event (BEFORE only)
```
mtijanic, line 1636: All skippage is valid in BEFORE only, so no need to be explicit.

Plugins/Events/Events/EventScriptEvents.cpp
```diff
@@ -0,0 +1,92 @@
...
+// BOOL CNWSModule::RunEventScript(int32_t nScript, CExoString* psOverrideScriptName)
+static NWNXLib::Hooks::Hook s_runModuleEventScript;
+// BOOL CNWSObject::RunEventScript(int32_t nScript, CExoString* psOverrideScriptName)
+static NWNXLib::Hooks::Hook s_runObjectEventScript;
+
+static void AreaRunEventScript(CNWSArea*, int32_t, CExoString*);
+static void ModuleRunEventScript(CNWSModule*, int32_t, CExoString*);
+static void ObjectRunEventScript(CNWSObject*, int32_t, CExoString*);
+
+#define CNWSAREA_SCRIPTARRAY_MAX_SCRIPTS            4
```
mtijanic, line 23: You can sizeof the relevant field in the class.

mtijanic:
As a stretch goal, check out "ID whitelist" in nwnx_events.nss

Daztek:
Can the hooked functions be a void when the originals return a BOOL?

niv:
> Can the hooked functions be a void when the originals return a BOOL?

I've amended the code to pass the BOOL through to nwserver. nwserver wasn't ever using it though.

mtijanic (approved):

PR #1511, opened 2022-04-24
Creature: Add "GetBodyBag" function

minikanyas:


Plugins/Creature/NWScript/nwnx_creature.nss
```diff
@@ -1020,6 +1020,11 @@ void NWNX_Creature_SetInitiativeModifer(object oCreature, int nMod, int bPersist
 /// @return the current custom initiative modifier for the creature
 int NWNX_Creature_GetInitiativeModifer(object oCreature);
 
+/// @brief Gets the Body Bag of a creature
+/// @param oCreature The target creature
+/// @return The creatures assigned Body Bag
+int NWNX_Creature_GetBodyBag(object oCreature);
```
Daztek, line 1026: Should return an object here ;D

minikanyas: Oooops, that's what happens when you copy-paste function declarations. Fixed.

PR #1519, opened 2022-05-04
Fix nwnx_creature declaration typo.

Jorteck:


mtijanic:
This is gonna break people's builds, but it's only been 3 weeks so hopefully not many.

Jorteck:
I think it was already broken, as the declaration had no implementation (the implementation had the correct spelling).

I only noticed it as the lexer we use was having a fit about it:

```c
void NWNX_Creature_SetInitiativeModifer(object oCreature, int nMod, int bPersist = FALSE);

void NWNX_Creature_SetInitiativeModifier(object oCreature, int nMod, int bPersist = FALSE)
{
    string sFunc = "SetInitiativeModifier";

    NWNX_PushArgumentInt(bPersist);
    NWNX_PushArgumentInt(nMod);
    NWNX_PushArgumentObject(oCreature);
    NWNX_CallFunction(NWNX_Creature, sFunc);
}
```

PR #1537, opened 2022-08-14
Replace AIMaster m_lEventQueue with a priority queue

Liareth:
This is a big optimization if your module has a lot of queued events (many areas, many delaycommands, lots of fighting, lots of effects, etc). Previously events that were not immediate would be added to the linked list in `AddEventAbsoluteTimeViaTail` which would iterate backwards over every calendar entry until it found a place to insert.

Side effect: Events with the same time of day are now executed in an undefined order. E.g:

```
DelayCommand(0.0f, First());
DelayCommand(0.0f, Second());
```

Who knows which one will happen first? I certainly don't. But if you don't care about that, and your server is big and full of events, this could be a valuable optimization for you.

### DIFF
--- a/Plugins/Optimizations/EventMasterPriorityQueue.cpp
+++ b/Plugins/Optimizations/EventMasterPriorityQueue.cpp
@@ -1,0 +1,95 @@
+#include "nwnx.hpp"
+
+#include "API/CAppManager.hpp"
+#include "API/CGameObjectArray.hpp"
+#include "API/CServerAIMaster.hpp"
+#include "API/CServerExoApp.hpp"
+#include "API/CServerExoAppInternal.hpp"
+#include "API/CWorldTimer.hpp"
+#include "API/CNWSPlayer.hpp"
+#include "API/CNWSObject.hpp"
+#include "API/CNWSCreature.hpp"
+#include "API/CExoBase.hpp"
+#include "API/CExoTimers.hpp"
+#include "API/CNetLayer.hpp"
+#include "API/CNWSMessage.hpp"
+
+#include <queue>
+
+namespace Optimizations
+{
+
+using namespace NWNXLib;
+using namespace NWNXLib::API;
+
+struct CServerAIEventNodeComparator
+{
+    bool operator()(const CServerAIEventNode& lhs, const CServerAIEventNode& rhs) const
+    {
+        return (lhs.m_nCalendarDay == rhs.m_nCalendarDay) ? (lhs.m_nTimeOfDay > rhs.m_nTimeOfDay) : lhs.m_nCalendarDay > rhs.m_nCalendarDay;
+    }
+};
+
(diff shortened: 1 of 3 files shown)


Plugins/Optimizations/EventMasterPriorityQueue.cpp
```diff
@@ -0,0 +1,95 @@
...
+    // 
+    // Don't care about save/load event queue.
+
+    if (Config::Get<bool>("EVENT_MASTER_PRIORITY_QUEUE", false))
+    {
+        static Hooks::Hook _0 = Hooks::HookFunction(API::Functions::_ZN15CServerAIMaster15ClearEventQueueEv, (void*)&ClearEventQueue, Hooks::Order::Early);
+        static Hooks::Hook _1 = Hooks::HookFunction(API::Functions::_ZN15CServerAIMaster20AddEventAbsoluteTimeEjjjjjPv, (void*)&AddEventAbsoluteTime, Hooks::Order::Early);
+        static Hooks::Hook _2 = Hooks::HookFunction(API::Functions::_ZN15CServerAIMaster27AddEventAbsoluteTimeViaTailEjjjjjPv, (void*)&AddEventAbsoluteTime, Hooks::Order::Early);
+        static Hooks::Hook _3 = Hooks::HookFunction(API::Functions::_ZN15CServerAIMaster12EventPendingEjj, (void*)&EventPending, Hooks::Order::Early);
+        static Hooks::Hook _4 = Hooks::HookFunction(API::Functions::_ZN15CServerAIMaster15GetPendingEventEPjS0_S0_S0_S0_PPv, (void*)&GetPendingEvent, Hooks::Order::Early);
```
Daztek, line 91: Please use `Hooks::Order::Final` for stuff that totally replaces a function!

Plugins/Optimizations/EventMasterPriorityQueue.cpp
```diff
@@ -0,0 +1,95 @@
...
+
+struct CServerAIEventNodeComparator
+{
+    bool operator()(const CServerAIEventNode& lhs, const CServerAIEventNode& rhs) const
+    {
+        return (lhs.m_nCalendarDay == rhs.m_nCalendarDay) ? (lhs.m_nTimeOfDay > rhs.m_nTimeOfDay) : lhs.m_nCalendarDay > rhs.m_nCalendarDay;
+    }
+};
+
+std::priority_queue<CServerAIEventNode, std::vector<CServerAIEventNode>, CServerAIEventNodeComparator> g_event_queue;
```
Daztek, line 33: We tend to prefix static globals like this with `s_`

Daztek (commented):
Just a bunch of minor things, also don't forget to update CHANGELOG.md!

mtijanic (approved):

pF-arQon:
> Any reason this was never merged?

Yeah: I got RL'd, then forgot about it. Sorry.

PR #1560, opened 2022-10-19
Support for loading feat modifiers from a 2da file

Cjreek:
That's already possible with racial modifiers and the race plugin.
Added equivalent functionality for the feat plugin.


Cjreek:
I'm done now

mtijanic:
Several things about this that I don't like, but it's a straight-up port from the other plugins that already do it poorly, so fixing it just here would be worse. Merging.

PR #1565, opened 2022-11-29
Update ELC.cpp

Shad000w:
Added racial ability adjustment into minimal ability score for spellcasting check to prevent issue when ELC would consider character with custom race granting wis/int/cha bonus or a custom spellcaster class with str/dex/con as casting ability.
Also note, shouldn't be m_nPrimaryAbility rather m_nSpellcastingAbility ?


Daztek:
Thanks, there's some more ELC issues in the plugin, and the basegame, I am hoping to look at at some point. Will keep this one in mind too

PR #1571, opened 2023-01-26
Added NWNX_Damage_DealMeleeDamage()

Cjreek:
Added a function to deal melee damage from an attacker to a target as if the attacker would have hit the target with its weapon. 

Includes a lot of arguments to customize. For example:

- Artificial damage multiplier to apply before resistances/immunities/reductions (default 1.0)
- Critical damage flags (deal no crit damage, roll for crit, always deal crit damage)
- Sneak attack damage flags (deal no sneak attack damage / check if sneak attack conditions apply / always deal sneak attack damage)
- On Hit effects and On Hit Spells (plus maximum number of triggered On Hit spells: default 1)
- Automatically roll maximum damage (default: false)
- Don't kill the target on a devastating crit (default: false)

Returns 1 for a regular hit, 3 for a crit and 10 for a devastating critical hit.
0 if something went wrong (invalid parameters)

### DIFF
--- a/Plugins/Damage/Damage.cpp
+++ b/Plugins/Damage/Damage.cpp
@@ -9,6 +9,18 @@
 #include "API/CNWSCreature.hpp"
 #include "API/CNWSCombatRound.hpp"
 #include "API/CNWSEffectListHandler.hpp"
+#include "API/CNWSInventory.hpp"
+#include "API/CNWSItem.hpp"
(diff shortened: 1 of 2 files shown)


Cjreek:
Please don't merge yet. The death attack saving throw is not correct. Will edit this when it's fixed.

Cjreek:
Fixed.

PR #1574, opened 2023-01-28
Do not reconcile auto map data on first run

Soren217:
The first time an object is loaded, the auto map data is already synced with the server, so we don't need to actually reconcile there.


Daztek:
Closing this since the opt is gone and it's done in basegame now.

PR #1583, opened 2023-02-13
core: rename Events->ScriptAPI, strip prefix for all ArgumentStack locations

niv:
`Events` was in the way. This renames all occurrences to `ScriptAPI`.

A followup change chould update NWNX_Events to pull the memory map and pubsub stuff into Core (leaving only the NWScript API and various hook in the plugin?)

Diff should also apply cleanly on .35, if you want to pull it in there right away.


mtijanic (approved):
:shipit:

PR #1584, opened 2023-02-19
Add OpenAI plugin

Liareth:
https://user-images.githubusercontent.com/9421831/219975283-5dc8702b-db3b-46bb-8cc4-2975fb90033f.mp4

This is a tech demo / prototype for using OpenAI in NWN. I won't be bringing this all the way, I just wanted to make a PR to share and serve as a reference for people working on making this type of feature production ready.

Next steps, if anyone wants to make this production ready:

* We use httplib in two plugins (this and webhook), it needs to be moved to become a shared dependency
* It would be nice if we could use ChatGPT API instead, which could make for higher quality results - there is no public API but check https://github.com/acheong08/ChatGPT for a reverse engineered reference

See NWNX discord thread for module (https://discord.com/channels/382306806866771978/1076960200898121889/1076973592220287127). Note that scripts in `NWScript/Example` are already included in the module, they're just duplicated here as well for easy reference.

### DIFF
--- a/Plugins/OpenAI/CMakeLists.txt
+++ b/Plugins/OpenAI/CMakeLists.txt
@@ -1,0 +1,7 @@
+find_package(OpenSSL)
+
+if (${OPENSSL_FOUND})
+    add_plugin(OpenAI OpenAI.cpp)
+    target_link_libraries(OpenAI ${OPENSSL_LIBRARIES})
+    target_include_directories(OpenAI PUBLIC ${OPENSSL_INCLUDE_DIR})
+endif()
--- a/Plugins/OpenAI/External/httplib.h
+++ b/Plugins/OpenAI/External/httplib.h
@@ -1,0 +1,8655 @@
+//
+//  httplib.h
+//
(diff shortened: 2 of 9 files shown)


mtijanic:
(as discussed on discord)

Plugins tied to a specific service - particularly proprietary ones - do not belong in the main repo, so we will not be merging this. It lives on as a PR and a branch on Liareth's repo so folks can use it however they see fit.

What we _would_ merge is a generic async HTTP request/response plugin and a nwscript file that leverages it to talk to OpenAI (or ChatGPT, etc)

Also listed in [Awesome-NWNX.md](https://github.com/nwnxee/unified/blob/master/Awesome-NWNX.md#other-nwnx-plugins)

invictus2010:
hi there! how might i test this out?

is the best way:
1)  pull Liareth:liareth/openai
2) fire it up as an NWNX instance on a VPS
3) log-in via a local client to test it out

i'm on Discord so can chat there if you prefer.

PR #1588, opened 2023-03-04
Core/Util: Migrate NWNX_Util_PluginExists to core function

jhett12321:


mtijanic:
This needs a changelog entry. Otherwise looks great.

PR #1591, opened 2023-03-05
[8193.35] Fix SWIG plugin compile. Add optional undefined symbol check.

jhett12321:
Fixes compiling SWIG on 8193.35.

This PR also includes an optional build target that will validate all exported symbols against the headers in NWNXLib.
Put the linux binary in the root SWIG folder, and trigger a reload of the cmake project. Building SWIG will now report any missing symbols:

image

image


Daztek:
Thanks! Cherry-picked the first two commits and fixed the undefined symbols

PR #1627, opened 2023-04-13
Optimize debug logging

jhett12321:
Adds an early return for max debug level, and a char* hashing function for the log level map.

### DIFF
--- a/NWNXLib/Log.cpp
+++ b/NWNXLib/Log.cpp
@@ -128,6 +128,7 @@ void WriteToLogFile(const char* message)
 // doesn't provide a hashing function for a const char* and I'm too lazy to write or copy/paste one.
 // Small perf hit when logging.
 static std::unordered_map<std::string, Channel::Enum> s_LogLevelMap;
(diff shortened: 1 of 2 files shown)


NWNXLib/Log.cpp
```diff
@@ -124,21 +124,46 @@ void WriteToLogFile(const char* message)
...
 
+Channel::Enum GetMaxLogLevel()
+{
+    return s_LogLevelMax;
+}
+
 void SetLogLevel(const char* plugin, Channel::Enum logLevel)
 {
     s_LogLevelMap[plugin] = logLevel;
+    s_LogLevelMax = std::max(s_LogLevelMax, logLevel);
```
mtijanic, line 166: Actually, this should to be:
```
for (auto it : s_LogLevelMap)
    s_LogLevelMax = std::max(s_LogLevelMax, it->second);
```
because levels can change dynamically. Although since that only affects performance if the max drops, we can do it in a follow on.

PR #1630, opened 2023-04-14
Creature: add DoCleaveAttack

summonFox:
Adds Creature: DoCleaveAttack()


Daztek:
Does this perform a cleave attack even if oCreature isn't currently attacking anything?

summonFox:
> Does this perform a cleave attack even if oCreature isn't currently attacking anything?

Ran it through some testing again as I'd forgotten if there were any caveats since I originally wrote this many moons ago, and yup; oCreature needs to both be currently attacking something and have attacks available in their combat round to use for cleave. I've expanded the function brief and added notes for this.

PR #1632, opened 2023-04-14
Object: add OverrideSpellProjectileVFX

summonFox:
This is very similar to NWNX_Creature's OverrideRangedProjectileVFX(), but made for spell VFX and added into NWNX_Object as more objects than just creatures can cast spells.


Daztek:
Maybe make a note in the description that this overrides *all* projectiles

summonFox:
> Maybe make a note in the description that this overrides _all_ projectiles
Note added.

PR #1636, opened 2023-04-14
Damage: SetAttackEventData can set iSneakAttack

summonFox:
Allows SetAttackEventData to set iSneakAttack. This only affects the attack roll message and floating text feedback as immunities and damage will have already been resolved by the time the attack event script is ran, but our use case for this that we've been using on our server is for changing attacks into sneak attacks and calculating all our new damage and immunities manually in nwscript during the attack event.


Plugins/Damage/Damage.cpp
```diff
@@ -211,6 +213,7 @@ NWNX_EXPORT ArgumentStack SetAttackEventData(ArgumentStack&& args)
         k = (int16_t)args.extract<int32_t>();
     }
     s_AttackData.nAttackResult = args.extract<int32_t>();
+    s_AttackData.nSneakAttack = args.extract<int32_t>();
```
Daztek, line 216: This is a breaking ABI change, scripts compiled with the old version will not send this last var and it'll throw a runtime error. 

Since we're currently on a development build I think it's ok, but you should mention it in the changelog with a big ol' `BREAKING CHANGE`

summonFox: ABI breaking note added to the changelog.

PR #1637, opened 2023-04-14
Creature: DoItemCastSpell () allow setting casting item and impact script

summonFox:
Allows oItem retrieved by GetSpellCastItem() to be set in DoItemCastSpell(), as well as overriding the impact script (or even removing the impact script if set to "****").

When combined with NWNX_ON_BROADCAST_SAFE_PROJECTILE_BEFORE it allows us to effectively create a ActionCastCustomFakeSpell() function that casts a ActionCastFakeSpell* or ActionCastSpell*, which we then replace with DoItemCastSpell() during the broadcast event to use a custom projectile or impact script instead.

### DIFF
--- a/Plugins/Creature/Creature.cpp
+++ b/Plugins/Creature/Creature.cpp
@@ -2850,9 +2850,12 @@ NWNX_EXPORT ArgumentStack DoItemCastSpell(ArgumentStack&& args)
           ASSERT_OR_THROW(delay >= 0.0f);
         auto projectilePathType = args.extract<int32_t>();
         auto projectileSpellID = args.extract<int32_t>();
+        auto oidItem = args.extract<ObjectID>();
+        auto impactScript = args.extract<std::string>();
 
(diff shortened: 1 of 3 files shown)


Plugins/Creature/Creature.cpp
```diff
@@ -2850,9 +2850,12 @@ NWNX_EXPORT ArgumentStack DoItemCastSpell(ArgumentStack&& args)
           ASSERT_OR_THROW(delay >= 0.0f);
         auto projectilePathType = args.extract<int32_t>();
         auto projectileSpellID = args.extract<int32_t>();
+        auto oidItem = args.extract<ObjectID>();
+        auto impactScript = args.extract<std::string>();
```
Daztek, line 2854: Like your other PR, this is also a breaking ABI change, would be nice to mention it in the changelog.

summonFox: Ah, my bad! I didn't realize that would be ABI breaking. Can I blame inexperience? I'll note it in the changelog, same with the other PR.

PR #1638, opened 2023-04-14
Util: add GetStringLevenshteinDistance

summonFox:
Super niche use, but maybe someone else out there also could find some use for it.

For us, we use it for calculating the crafting cost of editing item descriptions. Every character difference between the strings increases the cost for adding or remove text from the item.


Daztek (approved):
Will merge once merge conflicts are resolved

PR #1639, opened 2023-04-14
Object: add GetInventoryItemCount

summonFox:
Another niche use.

For our use case, we ran into a few players hoarding literal thousands of items (many 1x1 items filled into 1x1 containers, filled across many inventory tabs) which would cause any of our scripts that needed to loop over inventories to TMI. Getting an item count without hitting TMI in nwscript allows us to create a discord alert and warning messages for that player, or alter behavior of scripts that need to loop their inventory.


Plugins/Object/NWScript/nwnx_object.nss
```diff
@@ -385,6 +385,10 @@ int NWNX_Object_GetLastSpellCastDomainLevel(object oObject);
 /// @param sUUID The UUID to force
 void NWNX_Object_ForceAssignUUID(object oObject, string sUUID);
 
+/// @brief Returns how many items are in oObject's inventory.
+/// @param oObject A creature, placeable, or item.
```
Daztek, line 389: Missing the `/// @return ...`

summonFox: Oops, added!

PR #1640, opened 2023-04-14
Tweaks: add `NWNX_TWEAKS_CAST_ALL_ON_HIT_CAST_SPELL_ITEM_PROPERTIES`

summonFox:
Adds `NWNX_TWEAKS_CAST_ALL_ON_HIT_CAST_SPELL_ITEM_PROPERTIES` which casts all On Hit: Cast Spell item properties on hit, instead of only the first On Hit: Cast Spell property.


Plugins/Tweaks/CastAllOnHitCastSpellItemProperties.cpp
```diff
@@ -0,0 +1,76 @@
...
+void CastAllOnHitCastSpellItemProperties() __attribute__((constructor));
+void CastAllOnHitCastSpellItemProperties()
+{
+    if (!Config::Get<bool>("CAST_ALL_ON_HIT_CAST_SPELL_ITEM_PROPERTIES", false))
+        return;
+
+    LOG_INFO("All On Hit: Cast Spell item properties will be cast on hit");
+
+    static Hooks::Hook s_ApplyOnHitCastSpellHook = Hooks::HookFunction(&CNWSCreature::ApplyOnHitCastSpell,
+        (void*)+[](CNWSCreature *pCreature, CNWSObject *pTarget, __attribute__((unused)) CNWItemProperty *pProperty, CNWSItem *pItem) -> void
```
Daztek, line 32: You can just do `CNWItemProperty*` to get rid of the unused var warning

summonFox: Oh thank you! TIL!

PR #1648, opened 2023-04-29
Events: add `NWNX_ON_AREA_PLAY_BATTLE_MUSIC_*`

Daztek:


Arvirago:
Could this be used to detect when PCs enter combat?
Right now I use NWNX_ON_COMBAT_ENTER_BEFORE

Daztek:
Only that *some* player in the area is fighting a hostile creature.

PR #1649, opened 2023-05-14
Added NWNX_EXPERIMENTAL_END_COMBATROUND_AFTER_SPELLCAST 

Cjreek:
Makes it so that combat rounds end immediately after casting or aborting a spell so another spell can be cast right after.

### DIFF
--- a/Plugins/Experimental/EndCombatRoundAfterSpellcast.cpp
+++ b/Plugins/Experimental/EndCombatRoundAfterSpellcast.cpp
@@ -1,0 +1,26 @@
+#include "nwnx.hpp"
+
+#include "API/CNWSCombatRound.hpp"
+
+namespace Experimental {
+
+    using namespace NWNXLib;
+    using namespace NWNXLib::API;
+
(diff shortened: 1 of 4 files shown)


Plugins/Experimental/EndCombatRoundAfterSpellcast.cpp
```diff
@@ -0,0 +1,24 @@
...
+
+    using namespace NWNXLib;
+    using namespace NWNXLib::API;
+
+    void EndCombatRoundAfterSpellcast() __attribute__((constructor));
+    void EndCombatRoundAfterSpellcast()
+    {
+        if (!Config::Get<bool>("END_COMBATROUND_AFTER_SPELLCAST", false))
+            return;
+
```
Daztek, line 15: A `LOG_INFO("EXPERIMENTAL: ...");` here like the others have would be nice!

Cjreek: Done!

PR #1652, opened 2023-05-17
Added NWNX_ON_SPELL_FAILED_{BEFORE|AFTER}

Cjreek:
Notifies when a spell failed for any reason. The reason is passed to the event handler in the `REASON` event data and can be one of the following values:

```nwscript
const int NWNX_EVENTS_SPELLFAIL_REASON_CANCELED               = 0;
const int NWNX_EVENTS_SPELLFAIL_REASON_COUNTERSPELL           = 1;
const int NWNX_EVENTS_SPELLFAIL_REASON_ASF                    = 2;
const int NWNX_EVENTS_SPELLFAIL_REASON_SPELLFAILURE           = 3;
const int NWNX_EVENTS_SPELLFAIL_REASON_LOST_TARGET            = 4;
const int NWNX_EVENTS_SPELLFAIL_REASON_SILENCED               = 5;
const int NWNX_EVENTS_SPELLFAIL_REASON_DEFCAST_CONCENTRATION  = 6;
const int NWNX_EVENTS_SPELLFAIL_REASON_ENTANGLE_CONCENTRATION = 7;
const int NWNX_EVENTS_SPELLFAIL_REASON_POLYMORPHED            = 8;
const int NWNX_EVENTS_SPELLFAIL_REASON_CANT_CAST              = 9;
const int NWNX_EVENTS_SPELLFAIL_REASON_CANT_USE_HANDS         = 10;
```


Plugins/Events/Events/SpellEvents.cpp
```diff
@@ -4,25 +4,55 @@
...
+    float fTargetY;
+    float fTargetZ;
+    BOOL bInstant;
+    uint32_t nProjectilePathType;
+    uint32_t nFeat;
+    uint8_t nCasterLevel;
+    BOOL bFake;
+};
+
+static SpellCastAction LastSpellAction;
```
Daztek, line 33: Please prefix static globals with `s_` !

Plugins/Events/Events/SpellEvents.cpp
```diff
@@ -249,5 +286,155 @@ int32_t DecrementSpellReadyCountHook(CNWSCreature *thisPtr, uint32_t nSpellID, u
...
+            {61 /* NWNX_FEEDBACK_CAST_ARCANE_SPELL_FAILURE */,           2  /* NWNX_EVENTS_SPELLFAIL_REASON_ASF */},
+            {236 /* NWNX_FEEDBACK_CAST_EFFECT_SPELL_FAILURE */,          3  /* NWNX_EVENTS_SPELLFAIL_REASON_SPELLFAILURE */}, 
+            {21 /* NWNX_FEEDBACK_CAST_LOST_TARGET */,                    4  /* NWNX_EVENTS_SPELLFAIL_REASON_LOST_TARGET */},
+            {211 /* NWNX_FEEDBACK_CAST_USE_MOUTH */,                     5  /* NWNX_EVENTS_SPELLFAIL_REASON_SILENCED */}, 
+            {233 /* NWNX_FEEDBACK_CAST_DEFCAST_CONCENTRATION_FAILURE */, 6  /* NWNX_EVENTS_SPELLFAIL_REASON_DEFCAST_CONCENTRATION */}, 
+            {65 /* NWNX_FEEDBACK_CAST_ENTANGLE_CONCENTRATION_FAILURE */, 7  /* NWNX_EVENTS_SPELLFAIL_REASON_ENTANGLE_CONCENTRATION */}, 
+            {107 /* NWNX_FEEDBACK_CAST_CANT_CAST_WHILE_POLYMORPHED */,   8  /* NWNX_EVENTS_SPELLFAIL_REASON_POLYMORPHED */}, 
+            {22 /* NWNX_FEEDBACK_CAST_CANT_CAST */,                      9  /* NWNX_EVENTS_SPELLFAIL_REASON_CANT_CAST */}, 
+            {210 /* NWNX_FEEDBACK_CAST_USE_HANDS */,                     10 /* NWNX_EVENTS_SPELLFAIL_REASON_CANT_USE_HANDS*/}, 
+        };
```
Daztek, line 408: I feel just a switch would be better for this

Plugins/Events/Events/SpellEvents.cpp
```diff
@@ -249,5 +286,155 @@ int32_t DecrementSpellReadyCountHook(CNWSCreature *thisPtr, uint32_t nSpellID, u
...
+    LastSpellAction.nFeat = pNode->m_pParameter[10];
+    LastSpellAction.nCasterLevel = pNode->m_pParameter[11] & 0x000000FF;
+    LastSpellAction.bFake = (pNode->m_pParameter[9] & 0x80000000) != 0;
+
+    return s_AIActionCastSpellHook->CallOriginal<uint32_t>(thisPtr, pNode);
+}
+
+BOOL ClearActionHook(CNWSCreature* thisPtr, CNWSObjectActionNode* pNode, BOOL bIsTopmostAction)
+{
+    if ((IsIDInWhitelist("NWNX_ON_SPELL_FAILED", pNode->m_pParameter[0])) &&
```
Daztek, line 312: You check the whitelist, but I don't see any info in the documentation about it using it.

The whitelist isn't enabled by default, I'm not sure if that's what you intended, you can force enable it by calling `ForceEnableWhitelist("NWNX_ON_SPELL_FAILED");` in the `InitOnFirstSubscribe()` bit if needed

Plugins/Events/Events/SpellEvents.cpp
```diff
@@ -249,5 +286,155 @@ int32_t DecrementSpellReadyCountHook(CNWSCreature *thisPtr, uint32_t nSpellID, u
...
+{
+    if ((IsIDInWhitelist("NWNX_ON_SPELL_FAILED", LastSpellAction.nSpellId)) &&
+        (
+            (nFeedbackID == 61 /* NWNX_FEEDBACK_CAST_ARCANE_SPELL_FAILURE */) || 
+            (nFeedbackID == 236 /* NWNX_FEEDBACK_CAST_EFFECT_SPELL_FAILURE */) || 
+            (nFeedbackID == 21 /* NWNX_FEEDBACK_CAST_LOST_TARGET */) || 
+            (nFeedbackID == 211 /* NWNX_FEEDBACK_CAST_USE_MOUTH */) || 
+            (nFeedbackID == 233 /* NWNX_FEEDBACK_CAST_DEFCAST_CONCENTRATION_FAILURE */) ||
+            (nFeedbackID == 65 /* NWNX_FEEDBACK_CAST_ENTANGLE_CONCENTRATION_FAILURE */) ||
+            // Untested:
```
Daztek, line 392: how could you

Daztek (commented):
Seems ok for the most part, haven't ran it

Also please fix the changelog entry!

Plugins/Events/Events/SpellEvents.cpp
Cjreek: I was going for the same behaviour as the other spell events which are using a whitelist that's not enabled by default. Do you think it should be whitelisted by default?

PR #1656, opened 2023-05-18
Added feat modifier for spell DCs for spell schools and individual spells

Cjreek:
`NWNX_FEAT_MODIFIER_SPELLSAVEDCFORSCHOOL` + `NWNX_FEAT_MODIFIER_SPELLSAVEDCFORSPELL`.

Also added missing readme entry for `NWNX_FEAT_MODIFIER_SPELLSAVEDC`.

### DIFF
--- a/Plugins/Feat/Feat.cpp
+++ b/Plugins/Feat/Feat.cpp
@@ -320,8 +320,8 @@ void Feat::ApplyFeatEffects(CNWSCreature *pCreature, uint16_t nFeat)
         g_plugin->DoEffect(pCreature, nFeat, SeeInvisible);
     }
 
-    // SPELLSAVEDC
-    if (g_plugin->m_FeatSpellSaveDC[nFeat] != 0)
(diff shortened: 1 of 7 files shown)


Plugins/Feat/Feat.cpp
```diff
@@ -330,6 +330,8 @@ void Feat::ApplyFeatEffects(CNWSCreature *pCreature, uint16_t nFeat)
                     &CNWSCreature::CalculateSpellSaveDC,
                     +[](CNWSCreature *pThis, int32_t nSpellID) -> int32_t
                     {
+                        auto pSpell = Globals::Rules()->m_pSpellArray->GetSpell(nSpellID);
```
Daztek, line 333: When using `auto` please explicitly prefix pointers with a `*`, `auto *pSpell = ...` in this case.

You should also check it isn't returning a nullptr!

Cjreek: Fixed

PR #1665, opened 2023-06-05
Creature: Add `{Get|Set}MaxSellToStorePriceOverride()`

summonFox:
Add `NWNX_Creature_{Get|Set}MaxSellToStorePriceOverride()` for overriding the maximum gold price a creature can sell items to a store for.


Plugins/Creature/Creature.cpp
```diff
@@ -3393,3 +3394,92 @@ NWNX_EXPORT ArgumentStack BroadcastAttackOfOpportunity(ArgumentStack&& args)
...
+    +[](CNWSStore *pThis, CNWSItem *pItem, OBJECT_ID oidSeller) -> int32_t
+    {
+        if (auto *pSeller = Utils::AsNWSCreature(Utils::GetGameObject(oidSeller)))
+        {
+            if (auto maxBuyPriceOverride = pSeller->nwnxGet<int32_t>("STORE_MAX_SELL_TO_PRICE_OVERRIDE!" + NWNXLib::Utils::ObjectIDToString(pThis->m_idSelf)))
+            {
+                if (pItem->m_bPlotObject)
+                    return 0;
+
+                if (auto nCost = pItem->GetCost(true, true, false, false))
```
Daztek, line 3426: The basegame calls GetCost() using the default args (true, false, false, false), is your second true here intended?

Plugins/Creature/Creature.cpp
```diff
@@ -3393,3 +3394,92 @@ NWNX_EXPORT ArgumentStack BroadcastAttackOfOpportunity(ArgumentStack&& args)
...
+                {
+                    auto nMaxBuyPriceOverride = maxBuyPriceOverride.value();
+                    auto nBuyRate = pThis->GetCustomerBuyRate(oidSeller, pThis->m_bBlackMarket);
```
Daztek, line 3429: `pThis->m_bBlackMarket` should be `pItem->m_bStolen`, `m_bBlackMarket` is checked in GetCustomerBuyRate() and returns the relevant markdown based on that.

Plugins/Creature/Creature.cpp
summonFox: Oop, sorry! That'd be a typo on my end.

Plugins/Creature/Creature.cpp
summonFox: That makes infinitely more sense! Both issues fixed.

PR #1667, opened 2023-06-07
Add API Headers for NWSync

### LINKED ISSUE #1581: Expose headers required for custom nwsync usage
Regarding this feature that was added to the base game a while back:
```
- Dedicated servers can now load their module data from a specially prepared NWSync repository, instead of having to have a copy of the module and hak files. (This mainly benefits big servers that have a CI/CD pipeline set up.)
- Added argument -moduleurl, from which the module data will be sourced
- Added a new optional argument -modulehash; which is considered to be `latest` if omitted
- Generate the server/module repository with --with-module
- You can reuse the public repository to save on duplication/disk space; or use a separate/private one. If you reuse the same repository and do not want to leak the module data, turn off http content listing and write the data with --no-latest.
- Data is stored in SERVERHOME/nwsync (this works the same as on the client). It will auto-prune old data if the group_id matches. The rest is left up to you.
```

I want to make an NWNX tweak that allows to run this in a mixed mode. So you would load a .mod file, but additional content would be loaded from a standard nwsync repo (without --with-module) instead of hak files. Or a mix of both hak files, and the nwsync repo.

This is not possible right now as the following structures and methods do not exist in NWNXLib, or are incomplete. Could these please be exposed?
- [ ] NWSync::CNWCSync
- [ ] Hash::SHA1
- [ ] NWSync::CNWCSync::Progress
- [ ] InstanceLookup::List<T>
- [ ] CNWSync
- [ ] HttpBufferRef
- [ ] HttpUrl
- [ ] SimplyCopyingHttpBufferReceiver
- [ ] HttpConnection
- [ ] HttpAuthorizer
- [ ] CExoResMan::GetNWSync

jhett12321:
Adds support for implementing NwSync downloads through NWNX code.

Resolves #1581.

### DIFF
--- a/NWNXLib/API/API/CNWSync.hpp
+++ b/NWNXLib/API/API/CNWSync.hpp
@@ -8,14 +8,104 @@ NWN_API_PROLOGUE(CNWSync)
 #endif
 
 
+namespace NWSync {
 
+struct CNWSync {
+    void *m_internal;
+    char *m_tmp1;
+    uint32_t m_tmp2;
 
-
-struct CNWSync
-{
-    void * m_internal;
-
-
+    struct ManifestMetaData {
+        char field0_0x0;
+        long long field1_0x1;
+        long long field2_0x9;
+        int field3_0x11;
+        char field4_0x15;
+        char field5_0x16;
+        char field6_0x17;
+        char field7_0x18;
+        char field8_0x19;
+        char field9_0x1a;
+        char field10_0x1b;
+        char field11_0x1c;
+        char field12_0x1d;
+        char field13_0x1e;
+        char field14_0x1f;
+        char field15_0x20;
+        char field16_0x21;
+        char field17_0x22;
+        char field18_0x23;
+        char field19_0x24;
+        char field20_0x25;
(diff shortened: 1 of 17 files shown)


NWNXLib/API/API/CNWSync.hpp
```diff
@@ -8,14 +8,104 @@ NWN_API_PROLOGUE(CNWSync)
...
+    char *m_tmp1;
+    uint32_t m_tmp2;
 
-
-struct CNWSync
-{
-    void * m_internal;
-
-
+    struct ManifestMetaData {
```
mtijanic, line 18: Actual data:
```C
        Hash::SHA1 m_sha1;
        uint32_t m_hashTreeDepth;
        CExoString m_moduleName;
        CExoString m_description;
        CUUID m_uuid; // for SP
        uint32_t m_groupId; // for PWs/autopurge
        bool m_includesModuleContents;
        bool m_includesClientContents;
        uint32_t m_totalBytes;
        uint32_t m_totalFiles;
        uint32_t m_createdAt;
```

PR #1668, opened 2023-06-07
Implement Resources Plugin (NWSync/Custom Hak List)

jhett12321:
Merge first: https://github.com/nwnxee/unified/pull/1667

Adds support for synchronising a server to a nwsync repository, and running without hak files.


Plugins/Resources/Resources.cpp
```diff
@@ -0,0 +1,282 @@
...
+{
+    s_loadModuleStartHook = Hooks::HookFunction(&CNWSModule::LoadModuleStart, &OnLoadModuleStart, Hooks::Order::Late);
+}
+
+uint32_t Resources::OnLoadModuleStart(CNWSModule *pModule, CExoString param_1, int param_2, int param_3, const NWSync::Advertisement &param_4)
+{
+    bool reloadRequired = false;
+    uint32_t retVal = 0;
+
+    if (Config::Get<bool>("USE_CUSTOM_HAK_LIST"))
```
mtijanic, line 58: This will return a `std::optional<bool>`, which in an `if()` will evaluate to true if a value is set, regardless of which value. Since the docs say this should default to false, you should add a `, false` argument.

Plugins/Resources/Resources.cpp
```diff
@@ -0,0 +1,282 @@
...
+
+CExoString Resources::OnReadFieldCExoString(CResGFF *pGff, CResStruct *pStructure, char *szFieldID, BOOL &bSuccess, const CExoString &sDefault)
+{
+    if (strncmp(pGff->m_pFileType, "IFO", 3) == 0 && strcmp(szFieldID, "Mod_CustomTlk") == 0)
+    {
+        if (auto tlk = g_plugin->FindNwSyncTlk())
+        {
+            LOG_INFO("Using TLK file '%s' from NWSync", tlk->CStr());
+            bSuccess = true;
+            return {*tlk};
```
mtijanic, line 98: This should only be read during load, right? Would help to clear the hook afterwards, just to avoid the needless overhead in every GFF read ever afterwards.

Plugins/Resources/Resources.cpp
```diff
@@ -0,0 +1,282 @@
...
+        LOG_INFO("Using custom hak list: %s", hakListStr);
+    }
+    else
+    {
+        LOG_INFO("Skipping load of all hak files");
+    }
+
+    for (const std::string &hakName: hakList)
+    {
+        if (Globals::ExoResMan()->AddEncapsulatedResourceFile(CExoString("HAK:") + hakName, RESMAN_PRIORITY_USER_HAK))
```
mtijanic, line 187: Did you verify this loads them in correct priorities and not reversed?

Plugins/Resources/Resources.cpp
```diff
@@ -0,0 +1,282 @@
...
+}
+
+CExoString* Resources::FindNwSyncTlk()
+{
+    if (auto *pList = Globals::ExoResMan()->GetResOfType(Constants::ResRefType::TLK, false))
+    {
+        if (pList->m_nCount > 0)
+        {
+            // We can only load 1 custom tlk, so just use the first one we find.
+            return pList->m_pStrings[0];
```
mtijanic, line 208: This isn't quite right. Ignoring the fact that someone could arbitrarily put many tlks in the manifest for no good reason, there is also a perfectly good reason to put both `custom.tlk` and `dialog.tlk` in there. So at very least, this should skip `dialog{,f}.tlk`

Plugins/ThreadWatchdog/ThreadWatchdog.cpp
```diff
@@ -24,12 +25,18 @@ static Hooks::Hook s_MainLoopHook;
...
     : Plugin(services)
 {
-    s_MainLoopHook = Hooks::HookFunction(&CServerExoAppInternal::MainLoop,
-                                                  &MainLoopUpdate, Hooks::Order::Earliest);
-
     s_watchdogPeriod = Config::Get<uint32_t>("PERIOD", 15);
     // Default to effectively infinite
     s_watchdogKillThreshold = Config::Get<uint32_t>("KILL_THRESHOLD", ~0);
+
+    static Hooks::Hook loadModuleStartHook = Hooks::HookFunction(&CNWSModule::LoadModuleStart,
```
mtijanic, line 32: I have no clue why this is needed. The commit message is very light on details. Such a hack, if really needed, needs a detailed comment in the code and commit message.

mtijanic (changes requested):

PR #1670, opened 2023-06-13
Added transform arguments to VFX functions

Cjreek:
Added arguments for passing transform data (scale, translation, rotation) to `NWNX_Player_ShowVisualEffect()` and `NWNX_Player_ApplyInstantVisualEffectToObject()`


Cjreek:
I'm pretty sure the documentation error is not correct. Not sure what its problem is here.

PR #1680, opened 2023-07-09
Added NWNX_Creature_GetMaxAttackRange()

Cjreek:


Plugins/Creature/Creature.cpp
```diff
@@ -3573,3 +3573,12 @@ NWNX_EXPORT ArgumentStack SetAbilityIncreaseByLevel(ArgumentStack&& args)
...
     return {};
 }
+
+NWNX_EXPORT ArgumentStack GetMaxAttackRange(ArgumentStack&& args)
+{
+    auto* pCreature = Utils::PopCreature(args);
+    ASSERT_OR_THROW(pCreature);
+    auto oidTarget = args.extract<OBJECT_ID>();
+
+    return pCreature->MaxAttackRange(oidTarget);
```
Daztek, line 3583: Why an assert and not just
```c
if (auto *pCreature = Utils::PopCreature(args))
{
    auto oidTarget = args.extract<OBJECT_ID>();
    return pCreature->MaxAttackRange(oidTarget);
}
return 0.0f;  
``` 
?

Cjreek: I don't know 😄 
I can change it. Probably because I don't have to document the 2nd possible return value.

Daztek: If you would yea, everything else uses that format too. If the assert throws a runtime error it'd return 0.0f too. So just mention it returns 0.0f on error or something.

Cjreek: It's done

PR #1683, opened 2023-07-15
Added NWNX_ON_ATTACK_TARGET_CHANGE_BEFORE/AFTER events

Cjreek:
Also added `Utils::StringToObjectID()` as I did need that.

### DIFF
--- a/Plugins/Events/Events/CombatEvents.cpp
+++ b/Plugins/Events/Events/CombatEvents.cpp
@@ -28,6 +28,8 @@ static Hooks::Hook s_AddAttackActionsHook;
 static Hooks::Hook s_AddAttackOfOpportunityHook;
 static Hooks::Hook s_SetBroadcastedAOOToHook;
 static Hooks::Hook s_PlayBattleMusicHook;
+static Hooks::Hook s_AddActionHook;
+static Hooks::Hook s_ChangeAttackTargetHook;
 
 static void StartCombatRoundHook(CNWSCombatRound*, ObjectID);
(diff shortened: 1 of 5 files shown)


Plugins/Events/Events/CombatEvents.cpp
```diff
@@ -310,4 +321,96 @@ void PlayBattleMusicHook(CNWSAmbientSound *pThis, BOOL bPlay)
     }
 }
 
+static void AddActionHook(CNWSObject* pObject, uint32_t nActionId, uint16_t nGroupId, uint32_t nParamType1, void* pParameter1, 
+    uint32_t nParamType2, void* pParameter2, uint32_t nParamType3, void* pParameter3, uint32_t nParamType4, void* pParameter4, 
+    uint32_t nParamType5, void* pParameter5, uint32_t nParamType6, void* pParameter6, uint32_t nParamType7, void* pParameter7, 
+    uint32_t nParamType8, void* pParameter8, uint32_t nParamType9, void* pParameter9, uint32_t nParamType10, void* pParameter10, 
+    uint32_t nParamType11, void* pParameter11, uint32_t nParamType12, void* pParameter12)
+{
+    auto bOriginalCalled = false;
```
Daztek, line 330: Would like this `auto` to be explicitly `bool`. Same for the `auto oidXXX` ones further below and in the other hook, ObjectID/OBJECT_ID in this case.  

`auto` should only be used when the type is really obvious, due to a cast or a `Utils::AsNWSCreature(pObject);` like `auto *pCreature` has below.

Cjreek:
Fixed

PR #1689, opened 2023-07-27
Fixed: NWNX_Damage - damage event did not fire for doors

Cjreek:
Now it does.


Daztek:
I'd say it's more of a change rather than a fix. So it's probably better to list it under `Changed` in the changelog, also some driveby formatting seems to have snuck in!

Cjreek:
> also some driveby formatting seems to have snuck in!

Yes, I removed unnecessary, ugly empty lines. Do you want me to put them back in? 🤐

Cjreek:
Changed the changelog

PR #1690, opened 2023-08-08
Added NWNX_Player_GetTurd()

Cjreek:


Plugins/Player/Player.cpp
```diff
@@ -1952,3 +1952,18 @@ NWNX_EXPORT ArgumentStack SendPartyInvite(ArgumentStack&& args)
...
+
+NWNX_EXPORT ArgumentStack GetTURD(ArgumentStack&& args)
+{
+    const auto oidPlayer = args.extract<OBJECT_ID>();
+
+    CExoLinkedListInternal* pTURDS = Utils::GetModule()->m_lstTURDList.m_pcExoLinkedListInternal;
+    for (CExoLinkedListPosition pNode = pTURDS->pHead; pNode; pNode = pNode->pNext)
+    {
+        auto* pTURD = static_cast<CNWSPlayerTURD*>(pNode->pObject);
+        if (pTURD->m_oidPlayer == oidPlayer)
```
mtijanic, line 1964: Since this is nwn and all sorts of inconsistent states are possible, can you do a nullcheck for pTURD before dereferencing it? Maybe also for pTURDS?

Cjreek: Done

PR #1695, opened 2023-08-26
Reveal: Fix Reveal plugin always revealing PCs when the plugin is enabled, regardless of usage.

summonFox:
As the label on the tin says, this fixes PCs always being revealed to other PCs when the plugin is enabled regardless of plugin usage, as mentioned by someone on discord a while ago which I ran into today when enabling and attempting to use the plugin: https://discord.com/channels/382306806866771978/387223972808359936/1121979712818917436

### DIFF
--- a/Plugins/Reveal/Reveal.cpp
+++ b/Plugins/Reveal/Reveal.cpp
@@ -57,10 +57,9 @@ BOOL Reveal::HookStealthDetection(CNWSCreature* pObserverCreature, CNWSCreature*
             {
                 if (pObserverCreature->GetFaction()->GetLeader() == pHidingCreature->GetFaction()->GetLeader())
                 {
-                    if(pHidingCreature->nwnxGet<int>(detectionKey + "PARTY"))
-                    {
-                        *bSeen = true;
-                    }
+                    auto detectionVector = pHidingCreature->nwnxGet<int>(detectionKey + "PARTY");
+                    if (detectionVector && *detectionVector)
+                        *bSeen = *detectionVector;
                     *bHeard = true;
                     return true;
                 }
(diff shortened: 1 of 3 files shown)


Daztek:
I'll assume this fixes whatever is broken since I don't use it.

Plugins/Reveal/Reveal.cpp
```diff
@@ -46,28 +46,31 @@ Reveal::Reveal(Services::ProxyServiceList* services)
...
+            auto partyReveal = pHidingCreature->nwnxGet<int>(revealKey + "PARTY");
+            if (partyReveal && *partyReveal)
             {
                 if (pObserverCreature->GetFaction()->GetLeader() == pHidingCreature->GetFaction()->GetLeader())
                 {
-                    if(*pHidingCreature->nwnxGet<int>(detectionKey + "PARTY"))
+                    if(pHidingCreature->nwnxGet<int>(detectionKey + "PARTY"))
                     {
                         *bSeen = true;
                     }
```
Daztek, line 63: Shouldn't this be something like...

```c
if(auto detectionVector = pHidingCreature->nwnxGet<int>(detectionKey + "PARTY"))
    *bSeen = *detectionVector;
else
    *bSeen = false;
```

nwnxGet returns a std:optional() which is true when it's set, to get the actual value you need to deref it.

summonFox: I fixed the revealkeys and forgot to fix the detectionkeys. 🤦

summonFox: Fixed. Underlying logic in how seen/heard is set by the plugin is unchanged to not break anything unforeseen.

Plugins/Reveal/Reveal.cpp
```diff
@@ -46,37 +46,40 @@ Reveal::Reveal(Services::ProxyServiceList* services)
...
             {
                 if (pObserverCreature->GetFaction()->GetLeader() == pHidingCreature->GetFaction()->GetLeader())
                 {
-                    if(*pHidingCreature->nwnxGet<int>(detectionKey + "PARTY"))
-                    {
-                        *bSeen = true;
-                    }
+                    auto detectionVector = pHidingCreature->nwnxGet<int>(detectionKey + "PARTY");
+                    if (detectionVector && *detectionVector)
+                        *bSeen = *detectionVector;
```
Daztek, line 62: needs a

```
else
    *bSeen = false;
```

the variables that bSeen/bHeard are pointing to aren't initialized, so they should always be set to something

summonFox: Ah; done

PR #1696, opened 2023-08-29
Add DamageIncrease/Decrease to nwnx_feat

WilliamDraco:
Hacked together from existing examples. Tested functionality and it seems to work as expected

I was torn between inserting the constant/enum alphabetically where it should be vs just adding to the bottom. ABI-breaking implications? Only if they don't update nwnx_feat.nss. Let me know if you'd rather I maintain numbers and add it to the bottom.


Daztek:
Add it at the end, please!

PR #1699, opened 2023-09-17
Object: GetInventoryItemCount() will also work on stores.

ReachPW:
Stores have 5 CItemRepository's (one for each tab).

This adds a a function to count items in CItemRepository and loops over the store's to get full count.

Otherwise, existing Creature/Placeables/Items will call it and just return since they only have 1 CItemRepository.


Daztek:
Please fix the formatting, brackets on a new line, indentation, etc 🙃

ReachPW:
ok, hopefully better I tried to fix by hand.

Does this project use a formatter ?  I tried clang-format with a few style, but it raised a bunch of diffs.

PR #1706, opened 2023-11-05
ELC: Fix ability score calculation from class stat bonuses

jhett12321:


Plugins/ELC/ELC.cpp
```diff
@@ -730,19 +730,9 @@ static auto s_ValidateCharacter = Hooks::HookFunction(&CNWSPlayer::ValidateChara
                     nAbilityAtLevel[nAbilityIndex] += nStatMods[nAbilityIndex];
                 }
 
-                for (int nMultiClass = 0; nMultiClass < NUM_MULTICLASS; nMultiClass++)
+                for (int nAbilityIndex = 0; nAbilityIndex <= Ability::MAX; nAbilityIndex++)
```
Daztek, line 733: The constant had `None` added, so it should be `< Ability::MAX`

Daztek:
YOLO

PR #1708, opened 2023-11-18
DotNET: Add crash handler override support.

jhett12321:
Adds support for overriding the NWNX crash handler.

This allows a custom C# handler to attempt to resolve a managed backtrace before crashing the server, giving better visibility when debugging. And hopefully less Anvil bugs ending up in the NWNX chat ;)

Example from a double-free issue:

image

I tried to implement this through the existing `MessageBus::Broadcast("NWNX_CORE_SIGNAL")` system, but wasn't able to reach managed code again before encountering more dragons.

### DIFF
--- a/Plugins/DotNET/DotNETExports.cpp
+++ b/Plugins/DotNET/DotNETExports.cpp
@@ -9,6 +9,7 @@
 #include "API/CNWVirtualMachineCommands.hpp"
 #include "API/CWorldTimer.hpp"
 
+#include <csignal>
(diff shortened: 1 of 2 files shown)


mtijanic:
This is fine. By the way, you should be able to call the original crash handler from either managed code or the dotnet plugin. In the cpp, just declare as `extern "C" void nwnx_signal_handler(int sig);` and then call. In the plugin, use P/Invoke or whatever.

PR #1711, opened 2023-12-04
Clean NWNX binaries from LD_PRELOAD.

jhett12321:


Core/NWNXCore.cpp
```diff
@@ -145,6 +146,20 @@ NWNXCore::~NWNXCore()
...
+
+    std::string newPreload = preload;
+    while (std::regex_search(newPreload, regex))
+    {
+        newPreload = std::regex_replace(newPreload, regex, "");
+    }
+
+    setenv("LD_PRELOAD", newPreload.c_str(), true);
+}
+
```
Mingun, line 162: I think, that such global changes in user-supplied environment requires some comment about reasons, because this is a very non-standard actions

mtijanic: Hi, you're right, sorry, it's easy to forget about context when you've been freshly discussing it elsewhere. For the record, the issue is that sub-processes launched by any nwnx plugin will inherit LD_PRELOAD and attempt to preload NWNX_Core.so. This only works when preloaded onto nwn (nwmain or nwserver), and other process startups will fail. 
@jhett12321 can you please add a comment here as well? Maybe something in the readme too if you find a good place for it.

PR #1739, opened 2024-02-27
Add NWNX_Player_ReloadTlk()

julien-lecomte:
In case you want the less capable, but also less dangerous thing ;-)

Tested with 2 tlks: faerun_en.tlk and faerun_fr.tlk. Both served with nwsync.

Changed from "en" to "fr" via these two lines:

```
NWNX_Player_SetResManOverride(oSpeaker, RESTYPE_TLK, "faerun_en", "faerun_fr");
NWNX_Player_ReloadTlk(oSpeaker);
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@ https://github.com/nwnxee/unified/compare/build8193.36.12...HEAD
 - Store: {Get|Set}BlackMarketMarkDown()
 - Store: {Get|Set}MarkDown()
 - Store: {Get|Set}MarkUp()
+- Player: ReloadTlk()
 
 ### Changed
 - N/A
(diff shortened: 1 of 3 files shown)


julien-lecomte:
Hadn't noticed that this PR embarked other modifications that were not meant to be here.

julien-lecomte:
To continue on the conversation we had on discord: is this "good enough"? Or do you plan on doing a more complicated version with better features?

For my usage, it's all that I need. The only issue I could regret is not being able to set the tlk before chargen (for new feats or classes for example)

Daztek:
You could probably write a tweak that sends the tlk resource override + tlk reload before chargen

julien-lecomte:
I can look into this.

mtijanic:
Eh, let's just merge this anyway. Worst case, it'll get moved to deprecated as a wrapper of something else later. Or, removed and replaced with the tweak. But I think it's fine as is.

Please resolve the changelog conflict and I'll merge.

PR #1742, opened 2024-03-01
Add Object: {Get|Set}NameByLanguage

julien-lecomte:
In my effort to have dual-language module, using `SetName` will overwrite translations. For this reason, if we set the name for specific languages, and then usually `SetName(oObject, "")`, the translations will be correct for all players.

Their might be 3 gender values, neutral, male, and female, but experimenting with French made out to have male=0, female=1. Unsure, but I believe German might be a language with three language genders.

### DIFF
--- a/Plugins/Object/NWScript/nwnx_object.nss
+++ b/Plugins/Object/NWScript/nwnx_object.nss
@@ -418,6 +418,21 @@ int NWNX_Object_GetLastSpellInstant();
 /// @param oCreator The new creator of the trap. Any non-creature creator will assign OBJECT_INVALID (similar to toolset-laid traps)
 void NWNX_Object_SetTrapCreator(object oObject, object oCreator);
 
+/// @brief Return the name of the object for nLanguage.
+/// @param oObject an object
+/// @param nLanguage A PLAYER_LANGUAGE constant.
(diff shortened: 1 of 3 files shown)


Plugins/Object/NWScript/nwnx_object.nss
```diff
@@ -418,6 +418,20 @@ int NWNX_Object_GetLastSpellInstant();
...
+/// @param nGender   Gender to use, 0 or 1.
+/// @return The localized string.
+string NWNX_Object_GetLocalizedName(object oObject, int nLanguage, int nGender = 0);
+
+/// @brief Set the name of the object as set in the toolset for nLanguage.
+/// @note You may have to SetName(oObject, "") for the translated string to show.
+/// @param oObject an object
+/// @param nLanguage A PLAYER_LANGUAGE constant.
+/// @param nGender   Gender to use, 0 or 1.
+void NWNX_Object_SetLocalizedName(object oObject, string sName, int nLanguage, int nGender = 0);
```
Daztek, line 433: Documentation for sName is missing

julien-lecomte: Fixed.

PR #1743, opened 2024-03-01
Create a new store plugin

julien-lecomte:
##### New NWScript Functions
- Store: New `GetIsRestrictedBuyItem`
- Store: New `{Get|Set}BlackMarketMarkDown`
- Store: New `{Get|Set}MarkDown`
- Store: New `{Get|Set}MarkUp`

Since there's the possibilty of having a whitelist or blacklist for selling back to the store, I'm unsure if exposing just ``GetIsRestrictedBuyItem`` is sufficient.

### DIFF
--- a/Plugins/Store/Store.cpp
+++ b/Plugins/Store/Store.cpp
@@ -1,0 +1,90 @@
+#include "nwnx.hpp"
+
+#include "API/CNWSStore.hpp"
+
+using namespace NWNXLib;
+using namespace NWNXLib::API;
+
+NWNX_EXPORT ArgumentStack GetIsRestrictedBuyItem(ArgumentStack&& args)
+{
+    if (auto *pStore = Utils::PopStore(args))
+    {
+        const auto nBaseItemID = args.extract<int32_t>();
+        return pStore->GetIsRestrictedBuyItem(nBaseItemID);
+    }
+    return -1;
+}
+
+NWNX_EXPORT ArgumentStack GetBlackMarketMarkDown(ArgumentStack&& args)
+{
+    if (auto *pStore = Utils::PopStore(args))
+    {
+        return pStore->m_nBlackMarketMarkDown;
+    }
+    return -1;
+}
+
+NWNX_EXPORT ArgumentStack SetBlackMarketMarkDown(ArgumentStack&& args)
+{
+    if (auto *pStore = Utils::PopStore(args))
+    {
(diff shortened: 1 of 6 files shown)


Plugins/Store/Store.cpp
```diff
@@ -0,0 +1,92 @@
...
+    return -1;
+}
+
+NWNX_EXPORT ArgumentStack SetMarkUp(ArgumentStack&& args)
+{
+    if (auto *pStore = Utils::PopStore(args))
+    {
+        const auto nValue = args.extract<int32_t>();
+          ASSERT_OR_THROW(nValue >= 1);
+          ASSERT_OR_THROW(nValue <= 1000);
```
Daztek, line 77: Why are the asserts `>=1` and `<= 1000` here while for the others it's `>=0` and `<=100`, that said, are they percentages or flat goldpiece values?

Plugins/Store/Store.cpp
```diff
@@ -0,0 +1,92 @@
...
+        pStore->m_nMarkUp = nValue;
+    }
+    return {};
+}
+
+NWNX_EXPORT ArgumentStack GetCustomerCount(ArgumentStack&& args)
+{
+
```
Daztek, line 85: Style nit: please remove the extra newline

Plugins/Store/Store.cpp
```diff
@@ -0,0 +1,92 @@
...
+          ASSERT_OR_THROW(nValue >= 0);
+          ASSERT_OR_THROW(nValue <= 100);
+        pStore->m_nMarkDown = nValue;
+    }
+    return {};
+}
+
+NWNX_EXPORT ArgumentStack GetMarkUp(ArgumentStack&& args)
+{
+
```
Daztek, line 63: Style nit: please remove the extra newline

Plugins/Store/Store.cpp
julien-lecomte: The values are the min and max that the toolset editor allows us to use in the UI. The're percentage points: 110 is 10% increase in price.

Plugins/Store/Store.cpp
julien-lecomte: Fixed.

Plugins/Store/Store.cpp
julien-lecomte: Fixed.

PR #1747, opened 2024-03-16
Fixing GetPlayOption and SetPlayOption in Administration.cpp

LaputianBird:
Removed the asserts on the "option" variable, as unknown cases are caught by the default case of the following switch statement.


LaputianBird:
Starting again with this github hoodoo

PR #1756, opened 2024-05-05
DotNET: Implement new bootstrap

jhett12321:
This PR introduces a new method of bootstrapping C# assemblies that reduces the boilerplate code in C# & C++, and improves portability with simple C-style exports that matches the convention used by other NWNX plugins.

The updated bootstrapping code is backwards compatible, and should not be ABI breaking.


CMakeLists.txt
```diff
@@ -30,6 +30,11 @@ function(add_plugin target)
     configure_plugin(${target})
 endfunction()
 
+function(add_shared_plugin target)
+    add_library(${target} SHARED ${ARGN})
+    configure_plugin(${target})
+endfunction()
+
```
jhett12321, line 37: PInvoke cannot accessed exported symbols of "Module" visible libraries, so I added a new function here with different visibility.

I can inline this directly in `DotNET/CMakeLists` if you do not want this added here.

Plugins/DotNET/DotNET.cpp
```diff
@@ -122,50 +154,85 @@ static bool InitThunks()
...
-    auto dll = *assembly + ".dll";
-    auto full_ep = entrypoint + ", " + assembly->substr(assembly->find_last_of("/\\") + 1);
-    rc = load_assembly_and_get_function_pointer(dll.c_str(), full_ep.c_str(),
-                                                "Bootstrap", nullptr, nullptr, (void**)&bootstrap);
-    if (rc != 0 || bootstrap == nullptr)
-        LOG_FATAL("Unable to get %s.Bootstrap() function: dll='%s'; rc=0x%x", full_ep, dll, rc);
+    if (messageType == "ON_NWNX_LOADED")
+    {
+        Bootstrap();
+    }
```
jhett12321, line 182: Bootstrapping was moved to the end of NWNX init to allow for all enabled NWNX plugins to be loaded, in-case C# code attempts to use exports from other plugins.

Plugins/SWIG/SWIG_DotNET/SWIG_DotNET.cpp
jhett12321, line 1: These files are completely unused by the SWIG build, and can be removed.

Daztek:
Is this good to go in? I assume it doesn't beak any existing DotNet stuff?

PR #1771, opened 2024-09-08
Events: Client disconnect: Add PLAYER_NAME as event data

hendrikgit:


Daztek:
Would you like to be fancy and add the public cdkey too?

`auto cdkey = pPlayerInfo->m_lstKeys[0].sPublic;`

hendrikgit:
I'm happy to do that if you think it would help. It's not necessary for my usecase. I do this because I'm starting to log this information to sqlite and if I have the player name I can find the corresponding client connect event which already has all the info like cdkey.

hendrikgit:
Done. Now there is player name and cdkey.

Daztek:
:tada:

Issue #1450, opened 2021-12-02
[FEATURE REQUEST] Add NWNX_ON_EXPERIENCE_GAINED events


silent-rock:
Can we have NWNX_ON_EXPERIENCE_GAINED before and after added to nwnx_events plugin?

It would be great to have it skippable aswell.

gm-xanther:
Seconded, this would be really helpful for things like anti-cheat measures, exp boosting items/events, or making creatures that can't give exp like commoners or minion summons.

PR #1775, opened 2024-09-10
Events: Add events for player XP changes

hendrikgit:
#1450


plenarius:
@hendrikgit Please commit a CHANGELOG entry with this addition 🙏

plenarius:
Ah never mind, I completely missed in in my first look.

PR #1778, opened 2024-09-22
NoStack: Ignore Supernatural Innate Effects (Feat, Race, SkillRanks)

plenarius:
This offers the builder the option to set `NWNX_NOSTACK_IGNORE_SUPERNATURAL_INNATE` to `true` so your NoStack rules are ignored for effects granted through the Feat, Race and SkillRanks plugins.

These plugins set the subtype to `Constants::EffectSubType::Supernatural | Constants::EffectDurationType::Innate` on the effects granted.

This also corrects an issue when NoStack `GetTotalEffectBonus` hook was not increasing max caps for ability bonuses granted via the other plugins.


plenarius:
I don't know if there are any other Supernatural Innate effects in game that you'd want to honour your NoStack rules with but not these. I doubt it?

PR #1780, opened 2024-09-26
Tweaks: add `NWNX_TWEAKS_CAN_USE_ITEMS_WHILE_POLYMORPHED`

hendrikgit:
I'm not sure about the Hooks::Order


Daztek:
`Hooks::Order::Early` should be ok.

Here's what I usually follow, but it can be fuzzy
```
The rule for ordering should be roughly:
Earliest for pure event notification hooks
Early for skippable events, or events that modify state in before/after
Late for things that provide alternative implementation in some cases, like the POS getters in creature
Latest for things that almost never call the original
Final for things that fully reimplement base game functions
```

hendrikgit:
Changed to Hooks::Order::Early.

PR #1784, opened 2024-10-05
Race: Suppress and Reactive for temp disabling nwnx_race effects

WilliamDraco:
By request of Noler on discord.
Splits out the removal of the effects into its own function, and then adds nwscript calls to the Apply/Removal of them.

Compiled locally but untested. Only substantive c++ change was effect countdown remove rather than count-up creating a list to then count-through and remove from.

### DIFF
--- a/Plugins/Race/Race.cpp
+++ b/Plugins/Race/Race.cpp
@@ -184,14 +184,17 @@ void Race::RemoveRaceEffects(CNWSCreature *pCreature)
     if (pCreature->m_pStats == nullptr || effectsLevelAdded == 0)
         return;
 
+    std::vector<uint64_t> remove(128);
     for (int i = pCreature->m_appliedEffects.num; i >= 0; --i)
     {
         auto eff = (CGameEffect*)pCreature->m_appliedEffects.element[i];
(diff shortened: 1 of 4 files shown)


Plugins/Race/Race.cpp
```diff
@@ -178,6 +178,24 @@ void Race::DoEffect(CNWSCreature *pCreature,
...
+        return;
+
+    for (int i = pCreature->m_appliedEffects.num; i >= 0; --i)
+    {
+        auto eff = (CGameEffect*)pCreature->m_appliedEffects.element[i];
+        if (eff->m_sCustomTag == "NWNX_Race_RacialMod")
+        {
+            pCreature->RemoveEffectById(eff->m_nID);
+        }
+    }
```
Daztek, line 194: Deleting an effect can remove multiple linked ones so this would have a big chance for out of bound array access, you can't trust the effect list after a removal. The previous method of adding IDs to be removed to a list is better

WilliamDraco: Ah of course. I had reasoned that, under the old method, RemoveEffectById would try and remove invalid effects anyway, so it /must/ be fine to attempt to do so. Didn't think through the array access. New commit reverts this section to the previous method.

PR #1788, opened 2024-10-09
Creature: Add SetMulticlassLimit

hendrikgit:
Adds `NWNX_Creature_SetMulticlassLimit(object oCreature, int nLimit)` to limit the number of classes a PC can take dynamically.

### DIFF
--- a/Plugins/Creature/Creature.cpp
+++ b/Plugins/Creature/Creature.cpp
@@ -3584,3 +3584,67 @@ NWNX_EXPORT ArgumentStack GetMaxAttackRange(ArgumentStack&& args)
 
     return 0.0f;
 }
+
+NWNX_EXPORT ArgumentStack GetMulticlassLimit(ArgumentStack&& args)
+{
+    if (auto *pCreature = Utils::PopCreature(args))
+    {
+        return pCreature->nwnxGet<int32_t>("MULTICLASS_LIMIT").value_or(0);
+    }
+    return 0;
+}
+
+NWNX_EXPORT ArgumentStack SetMulticlassLimit(ArgumentStack&& args)
+{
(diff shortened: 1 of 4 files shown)


Plugins/Creature/Creature.cpp
```diff
@@ -3584,3 +3586,56 @@ NWNX_EXPORT ArgumentStack GetMaxAttackRange(ArgumentStack&& args)
 
     return 0.0f;
 }
+
+NWNX_EXPORT ArgumentStack SetMulticlassLimit(ArgumentStack&& args)
+{
+    if (!s_bGetIsClassAvailableHookInitialized)
```
Daztek, line 3592: No need to check/set a static bool, the static hook below will only be initialized once

Plugins/Creature/Creature.cpp
```diff
@@ -51,10 +51,12 @@ static bool s_bCriticalMultiplierHooksInitialized = false;
 static bool s_bCriticalRangeHooksInitialized = false;
 static bool s_bResolveAttackRollHookInitialized = false;
 static bool s_bResolveInitiativeRollHookInitialized = false;
+static bool s_bGetIsClassAvailableHookInitialized = false;
 static Hooks::Hook s_GetClassLevelHook = nullptr;
 
 static std::unordered_map<uint8_t, std::unordered_map<ObjectID, int16_t>> s_RollModifier;
 static std::unordered_map<ObjectID, bool> s_ParryAllAttacks;
+static std::unordered_map<OBJECT_ID, uint> s_MulticlassLimits;
```
Daztek, line 59: Why an unordered_map instead of using NWNX's POS? With POS you can have it persist to the bic too

hendrikgit:
Thanks Daz. I removed the static bool and switched to using POS.

Daztek:
Maybe add a `GetMulticlassLimit` too?

hendrikgit:
Added a get function.

PR #1795, opened 2024-10-31
Tweaks: add `NWNX_TWEAKS_RULESET_BARBARIAN_DR_CHANGE_FOR_PC_ONLY`

hendrikgit:
The goal is to allow PC barbarians to gain more DR but not make NPCs stronger at the same time.


hendrikgit:
Closing because it is too specific for the main repo.

PR #1819, opened 2025-02-12
MaxLevel: GetSpellsKnownPerLevelHook rewrite

hendrikgit:
This fixes the issue with Bards not having their cantrips when getting them on a class level other than 1 due to low charisma attribute. Other problems with the previous implementation might have existed but this is the one that I noticed.

@Daztek I found the [[maybe_unused]] attribute after a short Google search and don't know if it is appropriate here.

### DIFF
--- a/Plugins/MaxLevel/MaxLevel.cpp
+++ b/Plugins/MaxLevel/MaxLevel.cpp
@@ -60,7 +60,7 @@ static void SummonAssociateHook(CNWSCreature *pCreature, CResRef cResRef, CExoSt
 static void LoadSpellGainTableHook(CNWClass* pClass, CExoString sTable);
 static void LoadSpellKnownTableHook(CNWClass* pClass, CExoString sTable);
 static uint8_t GetSpellGainHook(CNWClass *pClass, uint8_t nLevel, uint8_t nSpellLevel);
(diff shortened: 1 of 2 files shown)


Plugins/MaxLevel/MaxLevel.cpp
```diff
@@ -329,43 +329,41 @@ static void LoadSpellKnownTableHook(CNWClass* pClass, CExoString sTable)
...
+    if (nLevel < 1 || nLevel > MAX_LEVEL_MAX) return 0;
 
-    if (pClass->m_lstSpellKnownTable)
+    if (!pClass->m_lstSpellKnownTable) return 0;
+
+    int nSpellLevelForLevel = nLevel <= CORE_MAX_LEVEL
+        ? pClass->m_lstSpellLevelsPerLevel[nLevel - 1]
+        : s_nSpellLevelsPerLevelAdded[pClass->m_nName][nLevel - 1];
+
+    if (nSpellLevel >= nSpellLevelForLevel) return 0;
```
Daztek, line 345: I think the original has a `nSpellLevel < 0 || ` check here. Without it, it might access the array at an invalid index

Daztek (commented):
I don't use this plugin so I'll trust you it works! Also a changelog entry would be nice

PR #1822, opened 2025-02-13
Tweaks: add `NWNX_TWEAKS_CON_DOES_NOT_AFFECT_HP`

hendrikgit:
@zunath asked about this and I gave it a try with the suggestions from Discord.

I wasn't sure if you wanted this for all creatures or only PCs and if negative con should also be disregarded so it may not be exactly what you wanted @zunath, but perhaps can be used as a start.


zunath:
Thanks for this @hendrikgit . I actually went ahead and put this into my own code base but it's neat that others can benefit from it. For our use case, we simply remove it for all creatures but others may want it to only apply to players. Maybe this is fine for now and could be expanded later if someone thinks it's needed?

PR #1835, opened 2025-07-11
Events: Add NWNX_ON_ABILITY_CHANGE

hendrikgit:


Daztek:
All the nwnxGet/Set stuff is horrible but I can't think of a better way myself :D 
Since I don't think it'll run that much it'll likely be fine and since it's an event it's opt-in so I don't really care either way

PR #1839, opened 2025-09-05
Sorting the characterlist by last played date

Cjreek:
Added NWNX_TWEAKS_CHARLIST_SORT_BY_LAST_PLAYED_DATE to enable sorting the character list by last played date, instead of character name

### DIFF
--- a/Plugins/Tweaks/SortCharListByLastPlayedDate.cpp
+++ b/Plugins/Tweaks/SortCharListByLastPlayedDate.cpp
@@ -60,7 +60,7 @@ static void SetCharacterListIndex(CharacterInfoWithFiletime* pCharInfo, int nInd
     }
 }
 
-static std::time_t GetBICFileTime(CExoString sVaultPath, const CResRef& sBicResRef)
+static std::time_t GetBICFileTime(const CExoString& sVaultPath, const CResRef& sBicResRef)
 {
     CExoString sFilename = sVaultPath + CExoString(sBicResRef) + ".bic";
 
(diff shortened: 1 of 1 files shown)


Cjreek:
Ehh.. does NWNX use C++11? It seems like <filesystem> is not available here?

Plugins/Tweaks/SortCharListByLastPlayedDate.cpp
```diff
@@ -0,0 +1,271 @@
...
+}
+
+static void SetCharacterListIndex(CharacterInfoWithFiletime* pCharInfo, int nIndex)
+{
+    auto& pStrList = pCharInfo->sLocFirstName.m_pExoLocStringInternal->m_lstString;
+    for (auto *pNode = pStrList.GetHeadPos(); pNode; pNode = pNode->pNext)
+    {
+        if (EXOLOCSTRING* sCharName = pStrList.GetAtPos(pNode))
+        {
+            sCharName->sString = CExoString::F("<c%c\x01\x01></c>", (char)(nIndex + 1)) + sCharName->sString;
```
Daztek, line 53: What if I have a thousand characters tho

Daztek (commented):
Don't forget to add a changelog entry

Plugins/Tweaks/SortCharListByLastPlayedDate.cpp
Daztek: I don't care all that much, a note somewhere should be fine

PR #1841, opened 2025-09-06
Delete NWNX_Lua and NWNX_SpellChecker

Daztek:
Can't wait for half a dozen people who actually use these to magically appear


jd28:
Why delete these? Another poor technical decision. Why not delete nwnx ruby? Is it because virusman's and niv's ZERO player servers, and I mean literally twenty years of ZERO players, depends on it?

PR #1854, opened 2025-10-31
NWNX_TWEAKS_CHARLIST_SORT_BY_LAST_PLAYED_DATE: env var takes number of characters to sort

hendrikgit:
I added no special env var value to mean "all", one can just set it to the maximum instead I figured. 2 million characters should be enough for anyone.

### DIFF
--- a/Plugins/Tweaks/SortCharListByLastPlayedDate.cpp
+++ b/Plugins/Tweaks/SortCharListByLastPlayedDate.cpp
@@ -27,6 +27,7 @@ using namespace NWNXLib::API::Constants;
 
 static BOOL SendServerToPlayerCharListHook(CNWSMessage* pMessage, CNWSPlayer* pPlayer);
 static Hooks::Hook s_SendServerToPlayerCharListHook;
+static size_t s_NumCharactersToSort;
 
 struct CharacterInfoWithFiletime : NWPlayerCharacterList_st
 {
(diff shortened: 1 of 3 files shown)


Plugins/Tweaks/SortCharListByLastPlayedDate.cpp
```diff
@@ -39,19 +40,33 @@ void SortCharListByLastPlayedDate()
     if (!Config::Get<bool>("CHARLIST_SORT_BY_LAST_PLAYED_DATE", false))
         return;
 
-    LOG_INFO("Character list will be sorted by last played date");
+    const int MAX_SORT_VALUE = 2097152; // 2^7 * 2^7 * 2^7 = 2^21 = 2097152, see below in SetCharacterListIndex
+
+    s_NumCharactersToSort = Config::Get<int>("CHARLIST_SORT_BY_LAST_PLAYED_DATE_NUM_CHARS", MAX_SORT_VALUE);
```
hendrikgit: How can I make `Config::Get<int>("CHARLIST_SORT_BY_LAST_PLAYED_DATE_NUM_CHARS", MAX_SORT_VALUE);` work with MAX_SORT_VALUE? I'm stuck on the `cannot bind rvalue reference of type ‘int&&’ to lvalue of type ‘int`.

hendrikgit:
Ok, made the Config::Get work via an int cast. Should be ready to merge @Daztek

PR #1872, opened 2026-04-17
fix: remove DIRT from surface speed map and tileset exclusions

plenarius:
## Summary
- Remove `DIRT (1)` from `m_SurfaceMaterialSpeeds` — only `DIRTPATH (28)` and `STONEPATH (29)` get +33% road speed
- Remove `ttu01`/`tcr10`/`tbm01` tileset exclusions from `SetPositionHook` — these were bandaids for incorrect walkmesh paint

Companion: layonara/nwn-haks#10 repaints all walkmesh road surfaces to use the correct `DIRTPATH`/`STONEPATH` materials.


plenarius:
Wrong target — this is a Layonara-specific plugin change.

PR #1873, opened 2026-04-19
Hide the player password from logs 

julien-lecomte:
In case there is a need for PCI DSS certification... /jk

It just kinda of bothered me seeing the password in the log files.

The default is to show the password, thus keeping the previous behavior. Hiding the password is opt-in.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,7 @@ https://github.com/nwnxee/unified/compare/build8193.37.13...HEAD
 - Damage: Added bRangedAttack to the NWNX_Damage_AttackEventData struct.
 - Events: Added ID to the NWNX_ON_ITEMPROPERTY_EFFECT_* events data.
 - Utils: Change LOG_INFO to LOG_DEBUG for console commands.
+- Admin: Player/DM password functions no longer print the passwords to the log.
 
 ### Deprecated
 - N/A
--- a/Plugins/Administration/Administration.cpp
+++ b/Plugins/Administration/Administration.cpp
@@ -57,15 +57,14 @@ static CExoLinkedListNode* FindTURD(std::string playerName, std::string characte
 NWNX_EXPORT ArgumentStack GetPlayerPassword(ArgumentStack&&)
 {
     const CExoString password = Globals::AppManager()->m_pServerExoApp->GetNetLayer()->GetPlayerPassword();
-    LOG_DEBUG("Returned player password '%s'.", password.m_sString);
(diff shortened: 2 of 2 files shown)


Plugins/Administration/Administration.cpp
```diff
@@ -54,17 +54,19 @@ static CExoLinkedListNode* FindTURD(std::string playerName, std::string characte
     return foundNode;
 }
 
-NWNX_EXPORT ArgumentStack GetPlayerPassword(ArgumentStack&&)
+NWNX_EXPORT ArgumentStack GetPlayerPassword(ArgumentStack&& args)
 {
     const CExoString password = Globals::AppManager()->m_pServerExoApp->GetNetLayer()->GetPlayerPassword();
-    LOG_DEBUG("Returned player password '%s'.", password.m_sString);
```
Daztek, line 60: Just remove the password printing from this one

Plugins/Administration/Administration.cpp
```diff
@@ -54,17 +54,19 @@ static CExoLinkedListNode* FindTURD(std::string playerName, std::string characte
...
+    const auto hidePass = !!args.extract<std::int32_t>();
+    LOG_DEBUG("Returned player password '%s'.", hidePass ? "REDACTED" : password.m_sString);
     return std::string(password.m_sString ? password.m_sString : "");
 }
 
 NWNX_EXPORT ArgumentStack SetPlayerPassword(ArgumentStack&& args)
 {
     const auto newPass = args.extract<std::string>();
-    LOG_NOTICE("Set player password to '%s'.", newPass);
```
Daztek, line 67: IMO, change it to LOG_DEBUG and remove the password printing from it

julien-lecomte: Also moved it below the action.

CHANGELOG.md
```diff
@@ -28,6 +28,7 @@ https://github.com/nwnxee/unified/compare/build8193.37.13...HEAD
 - Damage: Added bRangedAttack to the NWNX_Damage_AttackEventData struct.
 - Events: Added ID to the NWNX_ON_ITEMPROPERTY_EFFECT_* events data.
 - Utils: Change LOG_INFO to LOG_DEBUG for console commands.
+- Admin: NWNX_Admin password functions no longer print the passwords to the log.
```
julien-lecomte: Should I rather add the names of each of the 4 functions?

Daztek: `Admin: Player/DM password functions no longer print the passwords to the log.` or something like that I guess

PR #1878, opened 2026-05-05
Watchcat ^.^~

niv:
This change imports the old Watchcat.cpp.

On top of that, it adds some very light refactoring:

- Configuration is now driven by ENV, see README.md.
- More smarter bucketing of stacks.
- No longer hangs the watcher thread if the main thread signal does not get returned, for whatever reason.
- Added explicit exclusion of reentrant stalls like save and load. This should also address Resources.cpp taking a while to sync a repository in LoadModule.
- Some minor cross-thread var robustness.

The cat is entirely incompatible with the dog. Never run them both at the same time: They will interfere with each other and probably make your life miserable.

Brief testing on a local server seems to have preserved functionality.

### DIFF
--- a/NWNXLib/API/API/CAppManager.hpp
+++ b/NWNXLib/API/API/CAppManager.hpp
@@ -10,13 +10,34 @@ NWN_API_PROLOGUE(CAppManager)
 #endif
 
 struct CClientExoApp;
-struct CNWReentrantServerStats;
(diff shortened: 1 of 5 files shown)


Daztek:
Looks ok to me, just needs a Changelog.md entry under New Plugin

PR #1881, opened 2026-06-02
Add events ADD_TO_AREA_* REMOVE_FROM_AREA, and RemoveObjectFromAllDispatchLists

julien-lecomte:
Adds the event to get the objects added to the area.

Why is this needed? When a mob dies, there's no event to get which objects were added to the floor of the area. Also, the objects are added with a slight delay to the floor. The default "item lost" script does not fire. Please note that the creature treasure mode is set to "leave lootable corpse"

For my area garbage collector, when a mob dies, I currently add a delay event to scan the whole area and detect new items not already in my gc.

An alternative is  adding the mob death location to the delay, and only scanning a radius around the area.

With this event, when the mob dies, we get an event for the "remains" object.

Note: this will also fire for other events such as a PC entering the area.


Daztek:
Do you get a flood of events triggering if you do a CreateArea() that has a lot of objects? I don't remember if this function is hit when initializing areas

julien-lecomte:
Forgot to answer here; but yes there's a flood of events on CreateArea.

PR #1887, opened 2026-09-10
HTTPClient: Enable three custom fields to pass to response.

julien-lecomte:


Plugins/HTTPClient/NWScript/nwnx_httpclient.nss
```diff
@@ -80,6 +83,9 @@ int NWNX_HTTPClient_SendRequest(struct NWNX_HTTPClient_Request s)
     NWNXPushString(s.sHost);
     NWNXPushInt(s.nRequestMethod);
     NWNXPushString(s.sTag);
+    NWNXPushString(s.sCustom1);
+    NWNXPushString(s.sCustom2);
+    NWNXPushString(s.sCustom3);
```
Daztek, line 88: These need to be pushed first so it doesn't mess up pre-existing compiled scripts of people who only update nwnx and not recompile/update their scripts

Plugins/HTTPClient/NWScript/nwnx_httpclient.nss
```diff
@@ -91,6 +97,9 @@ struct NWNX_HTTPClient_Request NWNX_HTTPClient_GetRequest(int nRequestId)
 
     struct NWNX_HTTPClient_Request s;
 
+    s.sCustom3         = NWNXPopString();
+    s.sCustom2         = NWNXPopString();
+    s.sCustom1         = NWNXPopString();
```
Daztek, line 102: And here they need to be popped last

Plugins/HTTPClient/HTTPClient.cpp
```diff
@@ -311,6 +314,9 @@ NWNX_EXPORT ArgumentStack SendRequest(ArgumentStack &&args)
 {
     auto clientReq = Request();
     clientReq.id = s_clientRequestId++;
+    clientReq.custom3 = ScriptAPI::ExtractArgument<std::string>(args);
+    clientReq.custom2 = ScriptAPI::ExtractArgument<std::string>(args);
+    clientReq.custom1 = ScriptAPI::ExtractArgument<std::string>(args);
```
Daztek, line 319: These should be extracted last in a try/catch block

Plugins/HTTPClient/HTTPClient.cpp
```diff
@@ -349,5 +355,8 @@ NWNX_EXPORT ArgumentStack GetRequest(ArgumentStack &&args)
     ScriptAPI::InsertArgument(stack, clientReq.host);
     ScriptAPI::InsertArgument(stack, (int32_t) clientReq.requestMethod);
     ScriptAPI::InsertArgument(stack, clientReq.tag);
+    ScriptAPI::InsertArgument(stack, clientReq.custom1);
+    ScriptAPI::InsertArgument(stack, clientReq.custom2);
+    ScriptAPI::InsertArgument(stack, clientReq.custom3);
```
Daztek, line 360: These should be inserted first

ELadner:
HttpClient/README.md should be updated

